PR #144, opened 2015-10-08
[PPW-3351] Update Up-Next behavior and refactor upNextScreen

ksunika:
-- moved upNextPanel from upNext screen to Playing screen and Pause screen
-- now upNext is shown when video is paused, with stopped countdown


chibzu:
Looks good.

PR #156, opened 2015-10-15
Pbw 3253 -- glow effect

ksunika:
Adding glow effect to icons in control bar, more options screen, and close button in other screens. NOTE: doesn't work in IE and Edge


chibzu:
It looks like the highlight styles are the same across the board. Could we create a function in Utils.js that applies these styles and just pass it the element to highlight instead? This is a lot of repeated code I think we could streamline.

chibzu:
Wunderbar

PR #161, opened 2015-10-16
[PBW-3460] makes postroll ad play before upNext video if upNext is started by the timer.

ksunika:


js/components/countDownClock.js
```diff
@@ -156,7 +156,8 @@ var CountDownClock = React.createClass({
       "clickedVideo" : this.props.upNextInfo.upNextData,
       "custom" : {"source": CONSTANTS.SCREEN.UP_NEXT_SCREEN}
     };
-    this.props.controller.sendDiscoveryClickEvent(eventData);
+    var isAutoUpNext = true;
+    this.props.controller.sendDiscoveryClickEvent(eventData, isAutoUpNext);
   },
```
chibzu: Does startDiscoveryVideo need this flag also?

js/controller.js
```diff
@@ -620,10 +629,16 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       }
     },
 
-    sendDiscoveryClickEvent: function(selectedContentData) {
+    sendDiscoveryClickEvent: function(selectedContentData, isAutoUpNext) {
```
chibzu: All of the calls to this method should be updated to include "false" so we aren't relying on using undefined as false.

ksunika: done!

js/components/countDownClock.js
aldigun: @ksunika on Zhibzu comment.

js/components/countDownClock.js
```diff
@@ -156,7 +156,8 @@ var CountDownClock = React.createClass({
       "clickedVideo" : this.props.upNextInfo.upNextData,
       "custom" : {"source": CONSTANTS.SCREEN.UP_NEXT_SCREEN}
     };
-    this.props.controller.sendDiscoveryClickEvent(eventData);
+    var isAutoUpNext = true;
```
aldigun: We can remove this line and set sendDiscoveryClickEvent(eventData, true)

js/controller.js
```diff
@@ -58,7 +58,9 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         "countDownFinished": false,
         "countDownCancelled": false,
         "timeToShow": 0,
-        "showing": false
+        "showing": false,
+        "delayedSetEmbedCodeEvent": false,
+        "selectedContentData": null
```
aldigun: - Weird naming, "selectedContentData" can be confusing to developer.
- selectedContentData is only being used if delayedSetEmbedCodeEvent is set to TRUE. Do you agree to refactor these 2 variables?

aldigun: After re-thinking it, I don't like the other possibility even more.
Let's stays as is for 2 variables. But at least reset selectedContentData after it is being consumed.

ksunika: Ok, I will rename selectedContentData to delayedContentData and reset it after consumed.

js/components/countDownClock.js
ksunika: @chibzu, the problem was that postroll did not play if upNext timer was up. 
For Discover, we first see the postroll and only then the Discover screen with the timer appear, so the issue does not occur.

js/controller.js
ksunika: @aldigun, done

aldigun:
LGTM

PR #162, opened 2015-10-16
[PBW-3444] adds loading screen after Discovery click is initialized; …

ksunika:
…also remove propagation for 2 elements in upNext panel for accessibility control handling

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -622,6 +622,8 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
 
     sendDiscoveryClickEvent: function(selectedContentData) {
       this.state.upNextInfo.showing = false;
+      this.state.screenToShow = CONSTANTS.SCREEN.LOADING_SCREEN;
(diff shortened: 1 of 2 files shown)


js/controller.js
```diff
@@ -642,6 +642,8 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         this.state.upNextInfo.delayedSetEmbedCodeEvent = true;
       }
       else {
+        this.state.screenToShow = CONSTANTS.SCREEN.LOADING_SCREEN;
+        this.renderSkin();
```
aldigun: What would happen if we move renderSkin() after the 2 events publish below? I.e does renderSkin() need to be called here?

ksunika: Nothing would happen, I just did not want to separate two related lines - setting the screen and rendering. I can move it below if you think it will be better.

aldigun:
LGTM

PR #163, opened 2015-10-16
[pbw-ff-fix] making sure event is present where its called

ksunika:
Firefox wasn't able to play anything because on multiple screens 'event' was referenced, but not passed to the method.


aldigun:
Do we cover everything?

ksunika:
I looked through the changes I made last time, and searched through all the files. I think that is all.

aldigun:
looks good.

PR #164, opened 2015-10-20
[PBW-3258] Adding CSS blur to the video element on pause

chibzu:


js/controller.js
```diff
@@ -231,6 +233,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
           //Safari only can set cc when the video is playing, not before
           this.setClosedCaptionsLanguage();
         }
+        this.state.mainVideoElement.css({"-webkit-filter": "", "-moz-filter": "", "filter": ""});
```
aldigun: Should we out all the css either in inlineStyle.js or assets/styles.css?

PR #166, opened 2015-10-21
[PBW-3166] NaN time will be replaced by formatted 0.

ksunika:
-NaN time will be replaced by formatted 0
-end time on end screen is updated to reflect the latest passed video duration (was taken from content tree before)


chibzu:
Solid

PR #167, opened 2015-10-21
[pbw-glow-fix] getting moreOptions button opacity from the correct sk…

ksunika:
…in config setting


ksunika:
@chibzu, @mistermendez could you please review a very small fix?

PR #170, opened 2015-10-22
[PBW-3370] Subscribe to the earlier STREAM_PAUSED event for mobile lo…

chibzu:
…ck screen scenarios.


mistermendez:
LGTM

js/controller.js
```diff
@@ -84,7 +84,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       this.mb.subscribe(OO.EVENTS.CONTENT_TREE_FETCHED, 'customerUi', _.bind(this.onContentTreeFetched, this));
       this.mb.subscribe(OO.EVENTS.AUTHORIZATION_FETCHED, 'customerUi', _.bind(this.onAuthorizationFetched, this));
       this.mb.subscribe(OO.EVENTS.PLAYING, 'customerUi', _.bind(this.onPlaying, this));
-      this.mb.subscribe(OO.EVENTS.PAUSED, 'customerUi', _.bind(this.onPaused, this));
+      this.mb.subscribe(OO.EVENTS.STREAM_PAUSED, 'customerUi', _.bind(this.onPaused, this));
```
aldigun: Should not used STREAM_PAUSED. This is not a public event and is already removed with the next VTC feature.

chibzu: Well, that makes this fix more interesting then.

aldigun: We can stay as is using STREAM_PAUSED, but will need to change later on.

PR #171, opened 2015-10-22
[PBI-993] changing to hasClickUrl to match AMC change

chibzu:


chibzu:
http://github.services.ooyala.net/Platform/mjolnir/pull/350

PR #178, opened 2015-10-27
[PBW-3507] 

ksunika:
- fixes the problem with end screen shown when scrub back from the end screen using iOS controls.
- also modifies 2 places that were affected by PAUSED event previously replaced by STREAM_PAUSED

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -193,7 +193,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       if (this.state.screenToShow !== CONSTANTS.SCREEN.AD_SCREEN ) {
         this.state.duration = duration;
         if (this.skin.props.skinConfig.upNextScreen.showUpNext) {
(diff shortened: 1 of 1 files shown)


js/controller.js
```diff
@@ -193,7 +193,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       if (this.state.screenToShow !== CONSTANTS.SCREEN.AD_SCREEN ) {
         this.state.duration = duration;
         if (this.skin.props.skinConfig.upNext.showUpNext) {
-          if (!Utils.isIPhone()){//no UpNext for iPhone
+          if (!(Utils.isIPhone() || (Utils.isIos && this.state.fullscreen))){//no UpNext for iPhone or fullscreen iPad
```
chibzu: Missing parens at Utils.isIos, this will check if the function exists and not if it's iOS.

chibzu:
LGTM

PR #182, opened 2015-10-28
Pbw 3554

chibzu:


js/controller.js
```diff
@@ -207,7 +207,6 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       } else {
         this.state.queuedPlayheadUpdate = [currentPlayhead, duration, buffered];
       }
-      this.renderSkin();
```
aldigun: Do you not want to re-render the scrubber bar when this happen? Or this already managed inside scrubberBar.js?

chibzu: this.skin.updatePlayhead forces a render via forceUpdate

PR #183, opened 2015-10-29
[PBW-3186] makes text entered to one of three text fields in email ta…

ksunika:
…b not disappear after focusing


aldigun:
looks good.

PR #185, opened 2015-10-29
[PBW-3571] Fixing float behavior for short titles and descriptions.

chibzu:


js/views/startScreen.js
```diff
@@ -123,7 +123,7 @@ var StartScreen = React.createClass({
             <span className={playClass} style={playStyle} aria-hidden="true"></span>
           </div>
           <div className="startScreenInfo" style={screenStyle.infoPanel.style}>
-            {titleMetadata}
+            {titleMetadata}<div style={{clear: "both"}}></div>
```
ksunika: @chibzu, probably the same should be done to the case when promoImageSize == "small"

mistermendez: Any reason you didn't do it like this?

<div style={{clear: "both"}}>{titleMetadata}</div>

chibzu:
Good points, fixed.

PR #192, opened 2015-11-02
[PBW-3582] fix alignment on share screen

mistermendez:


js/styles/inlineStyle.js
```diff
@@ -458,7 +458,7 @@ module.exports = {
       background: "rgba(0, 0, 0, 0.75)",
       left: 0,
       right: 0,
-      display: "flex",
+      display: "-webkit-flex",
```
ksunika: @mistermendez, are you sure it won't break alignment in Firefox?

mistermendez: let me verify in FF..

PR #194, opened 2015-11-02
[PBW-3590] makes sure properties from skin config are applied in skin

ksunika:


chibzu:
Looks good, any dependency on skin-config changes we need to coordinate?

ksunika:
@chibzu, no dependencies on skin-config changes (there were changes in values for skin config, but these were merged yesterday -- https://github.com/ooyala/skin-config/pull/87)

chibzu:
Alright then, merging.

PR #201, opened 2015-11-05
[PBW-upNextIssue] sets isPlayingAd to true when pauseForAdPlayback; c…

ksunika:
…hecks for isPlayingAd when determining whether to show upNext


chibzu:
Very nice.

Issue #207, opened 2015-11-09
`gulp build` fails right after cleaning `/build/` directory


kuu:
Though `insertVersion` task depends on both `browserify` and `pretty`, they are all executed in parallel.

```
/ Build All
gulp.task('build', ['browserify', 'pretty', 'buildCss', 'insertVersion']);
```

mistermendez:
fixed in https://github.com/ooyala/html5-skin/pull/197

PR #197, opened 2015-11-04
Insert version fix

chibzu:

PR #208, opened 2015-11-09
Change the build sequence.

kuu:
This is a fix for the issue #207
`npm install` is required to install `gulp-sequence`.
(though gulp 4.0 will natively support for defining task dependencies in series or in parallel, which will make `gulp-sequence` unnecessary.)


mistermendez:
this got fixed in https://github.com/ooyala/html5-skin/pull/197 w/o using `gulp-sequence`

mistermendez:
@kuu also be sure to use branch `master` instead of `stable` for PR

PR #209, opened 2015-11-10
upgrade jest to work with node v4

mistermendez:
@ksunika @chibzu @aldigun @rchalooyala 

We need to upgrade our local node to v4 to work w/ this update.

Latest node version is v4.2.2


chibzu:
Good stuff.

PR #211, opened 2015-11-10
[PBW-3127] control bar hiding logic moved to controller

ksunika:
Updates control bar hiding logic for Playing and Ad screens:
- used in full screen desktop, full/normal screen Android, normal screen iOS
- timer is set/cancelled in controller
- if control bar items touched, timer restarts


ksunika:
forgot about scrubber bar touched

PR #219, opened 2015-11-17
[PBW-3706] autoHide == false keeps CB shown; fixes issue for upNext appearing before midroll for iPad

ksunika:


mistermendez:
LGTM

PR #221, opened 2015-11-18
[PBW-moreOptionsAnimation] animation for moreOptions is back

ksunika:
fix for broken moreOptions screen animation


mistermendez:
yum! :chocolate_bar:

PR #225, opened 2015-11-19
[PBW-3607] end screen to display correct duration of main video after ads

ksunika:


js/controller.js
```diff
@@ -314,6 +320,12 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       this.renderSkin();
     },
 
+    onVcPlayed: function(event, props) {
```
chibzu: Lets rename "props" to something clearer, like "source" just to avoid confusion with React props.

ksunika:
@chibzu, I renamed 'props' to 'source'

chibzu:
Good stuff.

PR #228, opened 2015-11-19
Unit test for discovery panel

mistermendez:


js/components/discoveryPanel.js
```diff
@@ -28,9 +26,39 @@ var DiscoveryPanel = React.createClass({
         })
       }),
       icons: React.PropTypes.objectOf(React.PropTypes.object)
+    }),
```
chibzu: Why use the shape type here?

js/components/discoveryPanel.js
```diff
@@ -28,9 +26,39 @@ var DiscoveryPanel = React.createClass({
...
+    controller: React.PropTypes.shape({
+      sendDiscoveryClickEvent: React.PropTypes.func
     })
   },
 
+  getDefaultProps: function () {
```
chibzu: Any benefit here outside of testing? Could we mock this up in the tests instead?

mistermendez: make the component less fragile by ensuring props have values that may not have been specified by parent component.

better summary [here](https://facebook.github.io/react/docs/reusable-components.html#default-prop-values)

js/components/discoveryPanel.js
mistermendez: I need to drill in and explicitly define proptype for sendDiscoveryClickEvent.

The controller proptype is less important.

chibzu: Let's change the controller type to be object to avoid confusion with external developers.

mistermendez: donezo

PR #230, opened 2015-11-19
[PBW-3740] initial volume set to 1 in controller instead of null, so …

ksunika:
…if no events are received, volume UI is at 1


chibzu:
Nice

PR #244, opened 2015-12-04
[PBW-3807] centering video in iPad

ksunika:


js/styles/inlineStyle.js
```diff
@@ -11,6 +11,24 @@ module.exports = {
     }
   },
 
+  iPadVerticalVideoStyle: {
```
mistermendez: is there a way to do this with CSS instead of inline JS style?

js/controller.js
```diff
@@ -146,6 +151,9 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       $("#" + elementId + " .player_skin").css("z-index", OO.CSS.ALICE_SKIN_Z_INDEX);
       this.state.mainVideoElement = $("#" + elementId + " .video");
 
+      this.state.playerHeight = $("#" + elementId + " .innerWrapper").height();
```
mistermendez: Can we do this without direct DOM manipulation (jQuery) and stick to React's virtual DOM?

It's considered best practice to not mix direct DOM manipulation with React.

See this for more details - https://facebook.github.io/react/docs/working-with-the-browser.html

ksunika: We should probably then change the above statements from jQuery to React as well. I am not sure about how to do it right now.
@chibzu, your thoughts?

js/styles/inlineStyle.js
ksunika: I am not sure that using the changing class name as a way to apply a different style for the main video element is completely safe. 
@chibzu, what do you think?

aldigun: I would think that the 4 bottom can be put on CSS. While width and height can be assigned on the fly in controller.js

js/controller.js
aldigun: AFAIK, innerWrapper is not made of react, so it is fine.
player_skin is the one we need to be careful of not mixing.

js/controller.js
```diff
@@ -45,6 +45,9 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       "elementId": null,
       "buffering": false,
       "mainVideoPlayhead": null,
+      "playerHeight": null,
+      "playerWidth": null,
+      "iPadVideoPositionSet": false,
```
aldigun: iPadVideoPositionSet is a weird naming

js/controller.js
```diff
@@ -92,6 +95,8 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       this.mb.subscribe(OO.EVENTS.CONTENT_TREE_FETCHED, 'customerUi', _.bind(this.onContentTreeFetched, this));
       this.mb.subscribe(OO.EVENTS.AUTHORIZATION_FETCHED, 'customerUi', _.bind(this.onAuthorizationFetched, this));
       this.mb.subscribe(OO.EVENTS.PLAYING, 'customerUi', _.bind(this.onPlaying, this));
+      this.mb.subscribe(OO.EVENTS.VC_VIDEO_ELEMENT_IN_FOCUS, 'customerUi', _.bind(this.onVideoControllerVideoElementInFocus, this));
```
aldigun: Do we really need to rely on in_focus and lost_focus event? Seems an overkill

js/controller.js
ksunika: how about iPadVideoCentered?

js/controller.js
```diff
@@ -262,8 +270,40 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+    onVideoControllerVideoElementInFocus: function() {
+      if (Utils.isIPad() && this.state.iPadVideoPositionSet == false) {
+        //hiding the video before it is positioned correctly by setIPadVideoPosition
+        this.state.mainVideoElement.css({"visibility":"hidden"});
+      }
+    },
+
     onPlaying: function() {
       // pause/resume of Ad playback is handled by different events => WILL_PAUSE_ADS/WILL_RESUME_ADS
+      if (Utils.isIPad() && this.state.iPadVideoPositionSet == false) {
```
aldigun: Couple comments:
1. this.state.iPadVideoPositionSet == false inside IF condition can be refactored to !this.state.iPadVideoPositionSet
2. I would recommend to move isIpad and iPadVideoPositionSet check inside the setIPadVideoPosition()

js/controller.js
ksunika: in in_focus we make the main video hidden, and then show it later. Without it, we have the video jumping from the side to the center and it is very visible.

js/controller.js
```diff
@@ -262,8 +270,40 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       }
     },
 
+    setIPadVideoPosition: function() {
```
aldigun: - weird naming as well. I suggest setVideoCentering?
- Put a comment explaining this method.

aldigun:
After discussion, this fix should be done by HTML5 video plugin rather than skin.
Will port the fix in the correct side and closing this merge.

PR #245, opened 2015-12-04
[PBW-3656] adding preventdefault to start, end, ad screens

ksunika:
- fixing what was missing from last time


js/views/startScreen.js
```diff
@@ -124,7 +129,7 @@ var StartScreen = React.createClass({
       // Small Promo Image configuration
       posterStyle.backgroundSize = "auto";
       return (
-        <div className="startScreen" onClick={this.handleClick} onTouchEnd={this.handleClick} style={screenStyle.style}>
+        <div className="startScreen" onClick={this.handleClick} onTouchEnd={this.handleClick} onMouseDown={this.handleMouseDown} style={screenStyle.style}>
```
mistermendez: why is onMouseDown here?

js/views/startScreen.js
```diff
@@ -138,7 +143,7 @@ var StartScreen = React.createClass({
       // Default configuration
       posterStyle.backgroundImage = "linear-gradient(to bottom, rgba(0,0,0,0.8) 0%, rgba(0,0,0,0.3) 100%), url('" + posterImageUrl + "')";
       return (
-        <div className="startScreen" onMouseUp={this.handleClick} onTouchEnd={this.handleClick} style={screenStyle.style}>
+        <div className="startScreen" onMouseUp={this.handleClick} onTouchEnd={this.handleClick} onMouseDown={this.handleMouseDown} style={screenStyle.style}>
```
mistermendez: why is onMouseDown here?

js/views/startScreen.js
ksunika: @mistermendez, the problem in PBW-3656 is that when you hold down the mouse button for a while, the cursor changes to 'text selection' cursor. To prevent it, onMouseDown is used here.

chibzu:
Can this be accomplished with user-select and/or pointer CSS rules instead of Javascript?

ksunika:
@chibzu, yes, looks like the combination of user-select:none and cursor:default will do it. I will change it for the rest of the screens as well then.

ksunika:
I will wait until reworked screens are in since it makes it easier to add css to the right place.

PR #248, opened 2015-12-08
Convert Screens to Sass

mistermendez:


mistermendez:
merge https://github.com/ooyala/html5-skin/pull/250 into this PR before merging this into master

mistermendez:
conflicts merged

js/mixins/truncateTextMixin.js
```diff
@@ -0,0 +1,12 @@
...
+
+var TruncateTextMixin = {
+  truncateText: function(node, text) {
+    // CSS doesn't support "truncate N lines" so we need to do DOM width
+    // calculations to figure out where to truncate the description
+    var descriptionNode = React.findDOMNode(node);
+    descriptionNode.innerHTML = Utils.truncateTextToWidth(descriptionNode, text);
+  }
+};
+module.exports = TruncateTextMixin;
```
chibzu: What's the benefit here? Not sure I see the benefit in saving one line of code for another layer of abstraction/misdirection.

js/mixins/resizeMixin.js
```diff
@@ -0,0 +1,19 @@
...
+
+  componentWillUnmount: function() {
+    window.removeEventListener('resize', this.handleResize);
+    window.removeEventListener('webkitfullscreenchange', this.handleResize);
+    window.removeEventListener('mozfullscreenchange', this.handleResize);
+    window.removeEventListener('fullscreenchange', this.handleResize);
+    window.removeEventListener('msfullscreenchange', this.handleResize);
+  }
+};
+module.exports = ResizeMixin;
```
chibzu: We should document the implicit contract this mixin creates, which is that there's a handleResize function on the component it's being mixed into.

js/mixins/truncateTextMixin.js
mistermendez: Hi @chibzu, mainly to make it consistent across components.

Some components were using deprecated getDOMNode() or truncating text after the component mounted with state variables, it was easier to consolidate into one function.

And we may need to use this in additional components like discovery

js/mixins/resizeMixin.js
mistermendez: @chibzu added js docs to mixins

js/mixins/truncateTextMixin.js
chibzu: Ah ha, hadn't seen that getDOMNode was deprecated or that those APIs were volatile. Early adopter woes, I suppose.

chibzu:
Looks good.

PR #249, opened 2015-12-08
[PBW-3766] removing WILL_PAUSE_ADS and WILL_RESUME_ADS, using VC_PAUS…

ksunika:
…ED and VC_PLAYED instead


js/controller.js
```diff
@@ -274,6 +271,14 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         this.state.mainVideoElement.css(InlineStyle.pauseScreenStyle.videoUnblur);
         this.renderSkin();
       }
+      if (source == OO.VIDEO.ADS) {
```
aldigun: Is VC_PLAYING always published for main -> ads -> main transition?

ksunika: Yes, every time ad or main video starts playing, we get the VC_PLAYING event. I tested skipping the ads also.

aldigun:
Looks good.

PR #261, opened 2015-12-15
UX font tweaks

mistermendez:
this was merged into `pbw-3634-screens` after it was merged into master and did not make it into master.

this PR is to merge `pbw-ux-font-tweaks` into master


mistermendez:
already CR'd and merged previously

PR #271, opened 2015-12-30
[PBW-3963] Prevent the secondary click event on mobile so we don't tr…

chibzu:
…igger the handler on the video.


aldigun:
looks good

PR #272, opened 2015-12-30
[PBW-3976] Fix for setting embed code after error

brianfwong:
Unsubscribes specific events when error occurs and then re-subscribes those events
when we set the embed code through API call.


js/controller.js
```diff
@@ -192,6 +192,18 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+      this.mb.subscribe(OO.EVENTS.VC_PLAYING, 'customerUi', _.bind(this.onPlaying, this));
+      this.mb.subscribe(OO.EVENTS.VC_PAUSED, 'customerUi', _.bind(this.onPaused, this));
+      this.mb.subscribe(OO.EVENTS.PAUSE, 'customerUi', _.bind(this.onPause, this));
+      this.mb.subscribe(OO.EVENTS.PLAYED, 'customerUi', _.bind(this.onPlayed, this));
+      this.mb.subscribe(OO.EVENTS.PLAYHEAD_TIME_CHANGED, 'customerUi', _.bind(this.onPlayheadTimeChanged, this));
+      this.mb.subscribe(OO.EVENTS.SEEKED, 'customerUi', _.bind(this.onSeeked, this));
+      this.mb.subscribe(OO.EVENTS.PLAYBACK_READY, 'customerUi', _.bind(this.onPlaybackReady, this));
+      this.mb.subscribe(OO.EVENTS.BUFFERING, 'customerUi', _.bind(this.onBuffering, this));
+      this.mb.subscribe(OO.EVENTS.BUFFERED, 'customerUi', _.bind(this.onBuffered, this));
+
```
aldigun: unnecessary new lines

js/controller.js
```diff
@@ -192,6 +192,18 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
 
     onAuthorizationFetched: function(event, authorization) {
       this.state.authorization = authorization;
+
+      this.mb.subscribe(OO.EVENTS.VC_PLAYED, 'customerUi', _.bind(this.onVcPlayed, this));
```
aldigun: These events are subscribed during init method. We should avoid double subscription.

brianfwong: @aldigun Have added a state check before re-subbing the events. However the "OO.EVENTS.PLAYBACK_READY" has to be present in order to perform the state check, so it was left outside the if condition.

js/controller.js
```diff
@@ -515,7 +530,16 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
     onErrorEvent: function(event, errorCode){
-      this.unsubscribeFromMessageBus();
+
+      this.mb.unsubscribe(OO.EVENTS.VC_PLAYED, 'customerUi');
+      this.mb.unsubscribe(OO.EVENTS.VC_PLAYING, 'customerUi');
+      this.mb.unsubscribe(OO.EVENTS.VC_PAUSED, 'customerUi');
+      this.mb.unsubscribe(OO.EVENTS.PLAYED, 'customerUi');
+      this.mb.unsubscribe(OO.EVENTS.PLAYHEAD_TIME_CHANGED, 'customerUi');
+      this.mb.unsubscribe(OO.EVENTS.SEEKED, 'customerUi');
+      this.mb.unsubscribe(OO.EVENTS.PLAYBACK_READY, 'customerUi');
```
aldigun: PLAYBACK_READY should not be unsubscribed. It is not part of basic playback events

js/controller.js
```diff
@@ -192,6 +192,21 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
 
     onAuthorizationFetched: function(event, authorization) {
       this.state.authorization = authorization;
+
+      // Outside of state check as this is needed in order to be able to perform the state check.
+      this.mb.subscribe(OO.EVENTS.PLAYBACK_READY, 'customerUi', _.bind(this.onPlaybackReady, this));
```
aldigun: I still don't think this is the right approach. I have couple suggestions:
1. Seems like there are couple places that need to subscribe / unsubscribe to the same events. let's just refactor them into 2 methods: subscribeBasicPlaybackEvents() and unsubscribeBasicPlaybackEvents()
2. Rather than listening to authorization fetched, perhaps you should listen to EMBED_CODE_CHANGED event. Then calls subscribeBasicPlaybackEvents().
3. We can move the same events inside init() to be listened in point number 2
4. Find a way to make sure that there is no duplicate subscription. You can use a flag by utilizing this.state

mistermendez:
@brianfwong please review and test, the logic is the same just rearranged to isolate core events

PR #275, opened 2015-12-31
[PBW-3627] Removing the resize logic for now. We don't want icons dis…

chibzu:
…appearing from the screen when it's open.


aldigun:
okay, we need to address this on responsive design.

PR #276, opened 2016-01-05
[PBW-3857] Discovery tray in full screen is not scaling properly

brianfwong:
Makes the size of discovery screen elements dynamic rather than fixed,
so they will adjust to match the view. Using a min-height/min-width to
prevent too small of an image from occurring on tiny screens.

### DIFF
--- a/scss/components/_discovery-screen.scss
+++ b/scss/components/_discovery-screen.scss
@@ -28,6 +28,7 @@
     height: 100%;
 
     .discovery-panel-title {
+      position: absolute;
       line-height: 1;
       font-weight: bold;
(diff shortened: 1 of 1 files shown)


aldigun:
@mistermendez can you review? Thanks.

mistermendez:
this looks off from page 6 & 7 of desktop specs - https://drive.google.com/a/ooyala.com/file/d/0B_g2pYPHx0pAUzlOTXpuTThzWHM/view

mistermendez:
I recommend reverting these changes and set `discoveryToasterContainerStyle` height to fixed 320px and remove min-height

mistermendez:
```
.discoveryToasterContainerStyle {
      //min-height: 54%;
      height: 320px;
    }
```

mistermendez:
:+1:

PR #277, opened 2016-01-06
[PBW-3906] unit tests for cc screen

ksunika:


package.json
```diff
@@ -31,7 +31,7 @@
     "gulp-sourcemaps": "1.5.2",
     "gulp-uglify": "1.1.0",
     "gulp-util": "3.0.6",
-    "jest-cli": "0.4.18",
+    "jest-cli": "*",
```
mistermendez: don't commit this update, the rest of the pbw team is using a previous version of node that is compatible with 0.4.18

mistermendez:
:+1:

PR #278, opened 2016-01-06
Pbw 3907

chibzu:
Ad screen moved to CSS approach and adding unit tests.


mistermendez:
@dustin does moving the height constants to Sass variables work here? see last commit.

mistermendez:
:+1:  LGTM

PR #279, opened 2016-01-07
[PBW-3992] adding 1px padding so that iPad renders CC correctly

ksunika:
For some reason iPad Safari refused to position closed caption element correctly until I added a 1px padding.


config
```diff
@@ -1 +1 @@
-Subproject commit c6b508798dfcf4821b315f13a1184a97fa4b2e38
+Subproject commit 7add71c9278b403119a960ab321afd8d53922e68
```
aldigun: Where is this pointing to?

aldigun:
Looks good.

PR #282, opened 2016-01-07
Pbw 4017

brianfwong:
Skip button logic was checking ad property, when it should be
checking the state for the skip button from the controller - as the
state is set by the "showAdSkipButton" EVENT.

Likewise the null check for the skipAdButtonEnabled should also be based on the
event controller rather than the ad property.

### DIFF
--- a/js/components/adPanel.js
+++ b/js/components/adPanel.js
@@ -120,21 +120,23 @@ var AdPanel = React.createClass({
     }
 
     // Skip
-    var handleSkipAdButtonClick;
+    var handleSkipAdButtonClick = null;
+    InlineStyle.adScreenStyle.skipButtonStyle.cursor = "default";
+    InlineStyle.adScreenStyle.skipButtonStyle.opacity = "0.3";
     if (this.props.currentAdsInfo.skipAdButtonEnabled === false) {
       InlineStyle.adScreenStyle.skipButtonStyle.visibility = "hidden";
       InlineStyle.adScreenStyle.skipButtonStyle.width = "0";
       InlineStyle.adScreenStyle.skipButtonStyle.marginLeft = "0";
     }
-    else {
+    else if (this.props.currentAdsInfo.skipAdButtonEnabled) {
+      handleSkipAdButtonClick = this.handleSkipAdButtonClick;
       InlineStyle.adScreenStyle.skipButtonStyle.visibility = "visible";
       InlineStyle.adScreenStyle.skipButtonStyle.width = "initial";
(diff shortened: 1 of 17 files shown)


js/components/adPanel.js
```diff
@@ -121,7 +121,7 @@ var AdPanel = React.createClass({
 
     // Skip
     var handleSkipAdButtonClick;
-    if (this.props.currentAdsInfo.currentAdItem.skippable === false) {
+    if (this.props.controller.state.currentAdsInfo.skipAdButtonEnabled === false) {
```
aldigun: This should use this.props.currentAdsInfo.skipAdButtonEnabled

js/components/adPanel.js
```diff
@@ -132,7 +132,7 @@ var AdPanel = React.createClass({
       InlineStyle.adScreenStyle.skipButtonStyle.marginLeft = "30";
     }
 
-    if (!this.props.currentAdsInfo.skipAdButtonEnabled) {
+    if (!this.props.controller.state.currentAdsInfo.skipAdButtonEnabled) {
       InlineStyle.adScreenStyle.skipButtonStyle.opacity = "0.3";
```
aldigun: The greying out can be removed as it is no longer needed. Either make it visible or hide it.

js/components/adPanel.js
```diff
@@ -132,17 +132,9 @@ var AdPanel = React.createClass({
...
-      handleSkipAdButtonClick = null;
-      InlineStyle.adScreenStyle.skipButtonStyle.cursor = "default";
-    }
-    else {
-      InlineStyle.adScreenStyle.skipButtonStyle.opacity = "1";
-      handleSkipAdButtonClick = this.handleSkipAdButtonClick;
-      InlineStyle.adScreenStyle.skipButtonStyle.cursor = "pointer";
-    }
-
+    InlineStyle.adScreenStyle.skipButtonStyle.opacity = "1";
```
aldigun: Line 135-137 can be refactored inside ELSE in line 129

js/components/adPanel.js
```diff
@@ -132,17 +132,9 @@ var AdPanel = React.createClass({
       InlineStyle.adScreenStyle.skipButtonStyle.marginLeft = "30";
     }
 
-    if (!this.props.currentAdsInfo.skipAdButtonEnabled) {
-      InlineStyle.adScreenStyle.skipButtonStyle.opacity = "0.3";
-      handleSkipAdButtonClick = null;
```
aldigun: Are these 2 lines 137-138 must be included inside IF statement in line 124?

PR #290, opened 2016-01-12
[PBI-1302] More special ad mode tweaks. Hide plugin element if empty …

chibzu:
…(not used) and put the ad marquee above the plugin element.


mistermendez:
:+1:

PR #291, opened 2016-01-12
Pbw 4017

brianfwong:
@aldigun New Pull request for refactored code.
Change of logic to use this.props.currentAdsInfo.skipAdButtonEnabled.
Now using up to date code that was refactored by both PBI-1302 (@chibzu) and
PBW-3907. Have also changed the z-index of the adScreen as it was
resting below a hidden layer added by PBI-1302, preventing both the
learn more and skip buttons from being clicked, as well as blocking all
other user-triggered ad interactions.

Left both the "visible" and "enabled" variables in the new button class even though they are identically set in this case, as they appear to be 'modularized' variables feeding this button to a common codebase for advert UI elements.

### DIFF
--- a/js/components/adOverlay.js
+++ b/js/components/adOverlay.js
@@ -2,20 +2,14 @@
   AD OVERLAY
 *********************************************************************/
 var React = require('react'),
-  InlineStyle = require('../styles/inlineStyle');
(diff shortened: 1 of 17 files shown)


aldigun:
Looks good.

PR #300, opened 2016-01-19
Pbw 4052

chibzu:


js/controller.js
```diff
@@ -177,8 +178,9 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         this.state.configLoaded = true;
         this.renderSkin();
 
-        $("#" + elementId + " .player_skin").append("<div class='player_skin_plugins'></div>");
+        $("#" + this.state.elementId + " .player_skin").append("<div class='player_skin_plugins'></div><div class='player_skin_plugins_click_layer'></div>");
         this.state.pluginsElement = $("#" + elementId + " .player_skin_plugins");
+        this.state.pluginsClickElement = $("#" + elementId + " .player_skin_plugins_click_layer");
```
mistermendez: should we also declare `pluginsClickElement` in the `state` object?

`"pluginsClickElement": null,`

chibzu: Good catch, thanks.

mistermendez:
:+1:

PR #302, opened 2016-01-19
[PBW-4095] added inline embed source to sample.html

rchalooyala:
This new inline embed source will be updated automatically with our stable deployments.


aldigun:
LGTM

PR #305, opened 2016-01-19
[PBW-4117] removing resize for MoreOptionsScreen to keep elements con…

ksunika:
…sistent after it was first open


mistermendez:
sweet :candy:

PR #306, opened 2016-01-19
[PBW-4052] Re-apply showing when we get the playing message for ads.

chibzu:


scss/components/_ad-screen.scss
```diff
@@ -1,6 +1,5 @@
 .adScreen {
   cursor: default;
-  z-index: 13000;
```
aldigun: Was this intentional? Can we do git blame and make sure we are not deleting other issue's fix?

chibzu: I did a git blame and found the relevant ticket commit, https://jira.corp.ooyala.com/browse/PBW-4017 . The behavior being fixed shouldn't be happening with all the latest code from the various repos.

aldigun:
Looks good!

PR #307, opened 2016-01-21
[PBW-4096] volumeSlider for Android appears on touch

ksunika:
-- Volume slider appears for Android when volume button is touched
-- If Volume button is touched while slider is shown, mute functionality works
-- Volume slider is hidden after Control Bar hides or is unmounted


chibzu:
Solid.

PR #314, opened 2016-01-23
[PBW-4025] Close caption is showing on End screen

brianfwong:
Publishing event to end CC at end of play. The check for ‘Safari’ was
preventing CC on replay as a result of the ending event change.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -326,10 +326,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         this.state.pluginsClickElement.removeClass("showing");
         this.state.screenToShow = CONSTANTS.SCREEN.PLAYING_SCREEN;
         this.state.playerState = CONSTANTS.STATE.PLAYING;
-        if (Utils.isSafari()){
-          //Safari only can set cc when the video is playing, not before
-          this.setClosedCaptionsLanguage();
-        }
+        this.setClosedCaptionsLanguage();
         this.state.mainVideoElement.removeClass('blur');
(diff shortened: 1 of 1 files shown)


js/controller.js
```diff
@@ -326,10 +326,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         this.state.pluginsClickElement.removeClass("showing");
         this.state.screenToShow = CONSTANTS.SCREEN.PLAYING_SCREEN;
         this.state.playerState = CONSTANTS.STATE.PLAYING;
-        if (Utils.isSafari()){
-          //Safari only can set cc when the video is playing, not before
```
mistermendez: do we not need this anymore?

js/controller.js
```diff
@@ -411,6 +408,9 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
     },
 
     onVcPlayed: function(event, source) {
+      var language = "";
```
mistermendez: var language unused?

js/controller.js
brianfwong: The issue was that this check was causing "this.setClosedCaptionsLanguage();" to only be done for Safari when it should be done for all browsers.

mistermendez:
@brianfwong conflicts

js/controller.js
brianfwong: As part of our intention here to wipe out the data, we set the language to a blank entry so it doesn't display on the endscreen.

PR #323, opened 2016-01-27
Responsive Start Screen

mistermendez:


mistermendez:
Works with VTC updates here - http://github.services.ooyala.net/Platform/mjolnir/pull/485

PR #324, opened 2016-01-28
[PBW-4214] rendering bitrate selection icon in CB

ksunika:
should work with https://github.com/ooyala/skin-config/pull/108


mistermendez:
:+1:

PR #330, opened 2016-01-29
[PBW-4108] Fixes Playhead icon not moving after seek.

brianfwong:
Flips of the order of the check to see if we are scrubbing (seeking) and if so move icon to scrubbed position. If not, we move the icon according the the currentPlayhead. Also sets the 'scrubbingPlayheadX' (the delta on scrubbing) to 0 at the end of handlePlayheadMouseUp so that it is not constantly trying to apply the value.


mistermendez:
dragging the playhead backwards doesn't work

mistermendez:
:+1: works now -

PR #331, opened 2016-01-29
[PBW-2965] adPanel responsive

ksunika:


mistermendez:
tested with freewheel ad :+1:

PR #332, opened 2016-01-30
[PBW-4018] Pointer/timeline reseting after overlay

brianfwong:
Playhead was intentionally being set to 0,0,0 at the end of ad play.
Have instead set the values (playhead, duration, buffered) to the
current states for each


mistermendez:
:+1:

PR #333, opened 2016-02-01
Bitrate popover

mistermendez:


ksunika:
@mistermendez, looks awesome! One thing I think should be added is that popover should close when we select the new bitrate (by Joe's design).

mistermendez:
@ksunika updated

ksunika:
looks good :+1:

PR #334, opened 2016-02-01
[PBW-4204] End screen doesn't appear if video is not configured with …

brianfwong:
…Discovery

Ensure there is discovery data before we try to set it.


mistermendez:
:+1:

PR #337, opened 2016-02-02
[PBW-4293] Clickthrough (resume from pause to play during ad) doesn't…

brianfwong:
Also fixes PBW-4182

When we tap on the ad screen the "click" event is registered always
while "touchEnd" event is registered inconsistently. So instead of
listening to "touchEnd”, if we listen to "click" event the issue will
be fixed.

### DIFF
--- a/js/views/adScreen.js
+++ b/js/views/adScreen.js
@@ -67,10 +67,13 @@ var AdScreen = React.createClass({
     event.cancelBubble = true; // IE
 
     this.props.controller.state.accessibilityControlsEnabled = true;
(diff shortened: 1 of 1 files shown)


ksunika:
It looks to me that the problem is in not that touchEnd is fired inconsistently, it's that when it's fired we don't go beyond the initial 'if' statement in handleTouchEnd that checks if Control Bar is visible. 
Now, even though it is visible, this.state.controlBarVisible == false when the ad is paused. I think the fix should correct the state of Control Bar.

brianfwong:
@ksunika Have switched the logic to instead prevent the control bar from being hidden during pause of an ad - based on the same logic used for playingScreen.js .

ksunika:
:+1:

PR #343, opened 2016-02-04
[PBW-4172] Clicking scrubber pointer reseting video to start

brianfwong:
Formular for scrubber pointer used a variable that was set to 0. Using
this.props.currentPlayhead instead.


ksunika:
looks good

PR #344, opened 2016-02-04
[PBW-3857] Responsive Values.

brianfwong:
Responsive values for discovery panel.

### DIFF
--- a/scss/components/_discovery-screen.scss
+++ b/scss/components/_discovery-screen.scss
@@ -64,9 +64,9 @@
 
     .discoveryImageWrapperStyle {
       position: relative;
-      height: 84px;
-      width: 150px;
-      margin: 24px;
+      width:214px;
+      height:170px;
+      margin: 30px;
       flex-shrink: 0;
     }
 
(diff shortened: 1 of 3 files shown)


mistermendez:
Discover items now appear on one line and pagination is off -

screen shot 2016-02-08 at 12 27 57 pm

mistermendez:
According to current mocks assets should not scale up in size.

screen shot 2016-02-08 at 12 32 58 pm

mistermendez:
Leave the small (_small.scss) and default medium (_discovery-screen.scss) style as-is and only adjust large style.

brianfwong:
The style document states that large is normal scale, small is 70% scale, medium is normal scale. That would be hard to achieve if I don't adjust small scale.

mistermendez:
1. The medium style in the app is already scaled down to 70%
2. Scale up 30% to achieve large size

This update changes the default medium style to large and the small style to medium. This update should only update the large style.

* small - the item dimensions are the same as medium, the items per page changes to 2 (done)
* medium - item dimensions already scaled down to 70% (done)
* large - scale item dimensions up by 30%

brianfwong:
@mistermendez Adjustments made to limit change to the large size file.

mistermendez:
Sweet :candy:

PR #347, opened 2016-02-05
Pbw 4106

chibzu:
Support IMA workflow

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -538,14 +538,16 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       var newCSS = {};
       var overlayPadding = (adInfo.ad.paddingRequired ? 50 : 0);
       if (adInfo.ad.height && adInfo.ad.height !== -1) {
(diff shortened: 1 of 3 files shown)


mistermendez:
There is issue with hidden video appearing on player resize

freewheel-ad-black-rect

mistermendez:
* Consider using relative positioning to center. i.e.:
  * `margin: 0 auto;`
  * https://github.com/zurb/foundation-sites/blob/develop/scss/util/_mixins.scss#L195
* Consider using this to visually hide:
  * https://github.com/h5bp/html5-boilerplate/blob/master/src/css/main.css#L124

aeng7:
I made a comment in the JIRA ticket about this. The player_skin_plugins div should keep its size when the player resizes. This will ensure the overlay is centered and the fake video is hidden behind the overlay.

chibzu:
Updated so that the overlay div is always the same size and just stays centered as the player resizes.

mistermendez:
:+1:

PR #348, opened 2016-02-05
[PBW-4303] Ensure mainVideoElement object only contains the main vide…

angieisrani:
…o element

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -159,7 +159,6 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
      ---------------------------------------------------------------------*/
     onPlayerCreated: function (event, elementId, params) {
       $("#" + elementId + " .innerWrapper").append("<div class='player_skin'></div>");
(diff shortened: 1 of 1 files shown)


js/controller.js
```diff
@@ -230,10 +229,15 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+    onVcVideoElementCreated: function(eventname, params) {
+      var element = $("#" + params["domId"]);
       if (Utils.isIE10()) {
-        this.state.mainVideoElement.attr("controls", "controls");
+        element.attr("controls", "controls");
+      }
+
+      if (params["videoId"] === OO.VIDEO.MAIN)
+      {
+        this.state.mainVideoElement = element;
```
mityaha: So, it is guaranteed that arguments for onVcVideoElementCreated are *correct*, right?

angieisrani: What do you mean when you ask if the arguments are correct?  Jquery will return a jquery object whether or not it finds an element.  Therefor the attr api will not faile even if there is no element with the given domId.

PR #349, opened 2016-02-08
[PBW-4142] User have to tap twice to replay the postroll video

brianfwong:
Adds an END_SCREEN_SHOWN event to onPlayed.


mistermendez:
:+1:

PR #354, opened 2016-02-09
[internal] Make fonts paths quickly configurable

chibzu:
Adding a variable to define the font paths in the CSS.


mistermendez:
:fire:

PR #355, opened 2016-02-09
[PBW-4018] FIx playhead duration on overlayAd end

brianfwong:
OverlayAds were setting main video duration to 0 on end - visible in
pause as the scrubber moving to the end, and not easily seen during
playback because it was very quickly set back to normal, but it was
happening in both playback and pause. Change instead sets duration to
mainvideo duration at end of ad.


aldigun:
looks good.

PR #356, opened 2016-02-09
[PBW-4205] Discovery button not shown if no discovery data

ksunika:
also, discovery on pause and on end not called if relatedVideos.videos is not defined.


mistermendez:
:+1:

PR #357, opened 2016-02-10
[PBW-4213][VTC] Publish AVAILABLE_BITRATE events

mityaha:
Passing real bitrate info from VTC, removed test code

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -255,10 +255,6 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
     onEmbedCodeChanged: function(event, embedCode, options) {
       this.state.videoQualityOptions.availableBitrates = null;
 
-      //test data, remove when real data available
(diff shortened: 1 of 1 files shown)


js/controller.js
```diff
@@ -578,8 +574,10 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
     },
 
     onBitrateInfoAvailable: function(event, bitrates) {
-      this.state.videoQualityOptions.availableBitrates = bitrates;
-      this.renderSkin({"videoQualityOptions": {"availableBitrates": bitrates}});
+      if (bitrates && bitrates.bitrates) {
+        this.state.videoQualityOptions.availableBitrates = bitrates.bitrates;
+        this.renderSkin({"videoQualityOptions": {"availableBitrates": bitrates.bitrates }});
+      }
```
ksunika: if PBF is going to pass the bitrates in the same format, then it looks fine to me.
@nagavenivn, can you look at this?

angieisrani: The format for passing the bitrates is documented in the event JSDoc.

mityaha: Yes, and bitrates.bitrates is an array of objects that satisfy this format.  We just had to wrap them into another object so that message bus won't split this array into many arguments.  So, Xeniq correctly mentioned that flash plugin should also pass to message bus such array wrapped in the object of the same format

angieisrani: Well the wrapping is done by the VTC, right?

mityaha: Yes

mistermendez:
real data, nice :smiley:

PR #358, opened 2016-02-10
[PBW-4205-ut] fixing unit tests for conditional discovery button

ksunika:


ksunika:
commited the same as part of 2961

PR #360, opened 2016-02-11
[pbw-4318] fix fullscreen ESC bug, added fallback if fullscreen API…

mistermendez:
…not supported natively

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -45,6 +45,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       "mainVideoDuration": 0,
       "adVideoDuration": 0,
       "mainVideoElement": null,
+      "mainVideoAspectRatio": 0,
       "mainVideoWrapper": null,
       "elementId": null,
       "pluginsElement": null,
@@ -93,6 +94,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       "errorCode": null,
       "isSubscribed": false,
       "isSkipAdClicked": false,
+      "isInitialPlay": false,
       "isFullScreenSupported": false,
       "isFullWindow": false
     };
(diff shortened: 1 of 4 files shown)


js/controller.js
```diff
@@ -604,13 +612,17 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       }
     },
 
-    onFullscreenChanged: function(event, fullscreen, paused) {
+    onFullscreenChanged: function() {
+      if (this.state.isFullScreenSupported) {
```
aldigun: This IF block is checked again inside toggeFullScreen method: if (this.state.isFullScreenSupported)
We can refactor this.state.fullscreen = Fullscreen.isFullscreen; into toggleFullScreen()

js/controller.js
```diff
@@ -687,15 +698,55 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+
+      // unhide scroll bars
+      document.documentElement.style.overflow = 'visible';
+
+      //remove full window style
+      this.state.mainVideoWrapper.removeClass('fullscreen');
+    },
+
+    // exit full window on ESC key
+    exitFullWindowOnEscKey: function(event) {
```
aldigun: This exit with ESC button is only applied to fullWindow. Does the other true fullScreen already has ESC button functionality?

mistermendez: yup, native fullscreen already works with ESC

js/controller.js
mistermendez: that should work, let me update and test

mistermendez: I'm getting this error - 

Failed to execute 'requestFullScreen' on 'Element': API can only be initiated by a user gesture.

mistermendez:
@aldigun conflicts merged

aldigun:
sweet!

PR #365, opened 2016-02-12
Add id attr to responsive div and fix unit tests

mistermendez:
id attr is to get ready for automated testing


ksunika:
looks good!

PR #367, opened 2016-02-17
Fix live stream style

mistermendez:


js/components/utils.js
```diff
@@ -84,23 +84,25 @@ var Utils = {
   * @return {String} The time as a string in the HH:MM format
```
rchalooyala: Should it be documented that nothing is returned if the timeInSeconds is not finite?

mistermendez:
@rchalooyala removed isFinite check from Util.formatSeconds().

isFinite check is in controlBar time duration.

rchalooyala:
yep, perfect

PR #372, opened 2016-02-18
[PBW-4184] pause animation for IMA

ksunika:
fixes the problem that pause animation is not visible for IMA ads because adScreen is behind the ad.


mistermendez:
:+1:

PR #374, opened 2016-02-18
[PBW-4293] Switch back to 'click' event-but with fix for showing controls

brianfwong:
Switching back to using ‘click’ as it will fire consistently while
‘touchend’ will only fire when/if the user releases the touch, and
might not still send an event to the controller. But keeping the fix to
keep the control bar visible in pause.


brianfwong:
Need to open this with candidate branch instead for CP.

PR #375, opened 2016-02-18
Pbw 4293

brianfwong:
Switching back to using ‘click’ as it will fire consistently while
‘touchend’ will only fire when/if the user releases the touch, and
might not still send an event to the controller. But keeping the fix to
keep the control bar visible in pause.


ksunika:
If you remove touchend events from handlePlayerClicked altogether, then you should think about what should be done in handleTouchEnd, since it calls handlePlayerClicked passing it touchend as an event.

handleTouchEnd: function(event) {
    if (!this.state.controlBarVisible && this.props.skinConfig.adScreen.showControlBar) {
      this.showControlBar();
      this.props.controller.startHideControlBarTimer();
    }
    else {
      this.handlePlayerClicked(event);
    }
  }

brianfwong:
Closing for now, to investigate further and in more detail after the release as this is a p3 not p2 issue.

PR #378, opened 2016-02-19
Pbw 4380

brianfwong:


js/views/endScreen.js
```diff
@@ -28,10 +28,15 @@ var EndScreen = React.createClass({
   },
 
   handleClick: function(event) {
-    // pause or play the video if the skin is clicked
-    event.preventDefault();
-    this.props.controller.state.accessibilityControlsEnabled = true;
-    this.props.controller.togglePlayPause();
+    if (event.type == 'touchend' || !this.isMobile){
```
ksunika: we never set this.isMobile on this screen, so it is undefined. This check does not do anything here. 

Also, since the handleClick is called from onClick={this.handleClick}, the event is always click, never touchend.

brianfwong:
This PR is in error, issue was actually fixed by earlier code fix (PR 366).

PR #380, opened 2016-02-19
Pbw 4463

chibzu:


chibzu:
Branched off of 4.1.1.fix and seemed to have picked up a couple CP commits, 5db6031 is the relevant commit for the bug.

PR #381, opened 2016-02-19
[PBW-4463] Preventing pause logic if an unfocused element is paused.

chibzu:


aldigun:
Looks good. Does this need a unit test?

chibzu:
In the grand scheme, yes. We don't currently have controller.js mocked out for testing though, just the components. It's complicated since it's not a standalone react component and has a bunch of core dependencies :\

js/controller.js
```diff
@@ -590,6 +592,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
     },
 
     onVideoElementFocus: function(event, source) {
+      this.focusedElement = source;
```
angieisrani: Note that it's also possible to lose focus and have no elements in focus

PR #382, opened 2016-02-20
[PBW-4259] only show buffering spinner when buffering or right after …

ksunika:
…the play button clicked on start screen.

Adds the condition that prevents buffering spinner to be shown when pausing the ads.

Also, since the new fullscreen logic prevents the onFullscreenChanged to be called for iPhone, moving the logic from there to onPaused and onPlayed.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -406,6 +406,10 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
           // default
           this.state.screenToShow = CONSTANTS.SCREEN.PAUSE_SCREEN;
         }
+        if (Utils.isIPhone()){
+          //iPhone pause screen is the same as start screen
+          this.state.screenToShow = CONSTANTS.SCREEN.START_SCREEN;
+        }
         this.state.playerState = CONSTANTS.STATE.PAUSE;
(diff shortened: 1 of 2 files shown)


ksunika:
@mistermendez or @chibzu please review

js/controller.js
```diff
@@ -441,6 +445,10 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         this.state.screenToShow = CONSTANTS.SCREEN.END_SCREEN;
         this.mb.publish(OO.EVENTS.END_SCREEN_SHOWN);
       }
+      if (Utils.isIPhone()){
+        //iPhone end screen is the same as start screen, except for the replay button
+        this.state.screenToShow = CONSTANTS.SCREEN.START_SCREEN;
```
aldigun: If we have this here, will it went into end_screen and shows replay button?

ksunika: @aldigun, for iPhone we always show start screen for everything. We have a special check that looks if the start screen is shown when the playerState == CONSTANTS.STATE.END, and then instead of play button we show replay button.

aldigun: :+1:

PR #385, opened 2016-02-23
Pbw 4293

brianfwong:
Only start the process to hide the controlbar when leaving a paused
state, but also handle the click event even when touchend is also used
to set the controlbar to visible.

### DIFF
--- a/js/views/adScreen.js
+++ b/js/views/adScreen.js
@@ -74,11 +74,9 @@ var AdScreen = React.createClass({
   },
 
   handlePlayerClicked: function(event) {
-    if (event.type == 'click' || !this.isMobile) {
+    if (event.type == 'touchend' || !this.isMobile) {
       //since mobile would fire both click and touched events,
       //we need to make sure only one actually does the work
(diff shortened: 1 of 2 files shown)


js/views/adScreen.js
```diff
@@ -68,6 +68,9 @@ var AdScreen = React.createClass({
     event.cancelBubble = true; // IE
 
     this.props.controller.state.accessibilityControlsEnabled = true;
+    if ((event.type == 'click' || !this.isMobile) && !this.props.skinConfig.adScreen.showAdMarquee) {
+      this.props.controller.onAdsClicked(CONSTANTS.AD_CLICK_SOURCE.VIDEO_WINDOW);
```
chibzu: Why is the showAdMarquee setting relevant here?

brianfwong:
showAdMarquee relates to PBW-4182, which needs the fixes from PBW-4293 as well as its own fix. So this is telling the video to allow clicking when showAdMarquee is not present.

chibzu:
LGTM

PR #387, opened 2016-02-23
[PBW-4484] IMA adds shifting up

brianfwong:
Style sheet is expressely setting the plugin container to the top. Can
either remove absolute position (and create unpredictable future
issues), or set the height of the plugin to fill its container (height:
100%).


brianfwong:
Closing - further comments in ticket imply the bug might be as intended.

PR #389, opened 2016-02-24
[PBW-4071] publishing OO.EVENTS.NON_LINEAR_AD_DISPLAYED when overlay …

ksunika:
…is loaded (VAST)

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -605,7 +605,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
 
     onAdOverlayLoaded: function() {
       this.state.adOverlayLoaded = true;
-      this.mb.publish(OO.EVENTS.NON_LINEAR_AD_DISPLAYED);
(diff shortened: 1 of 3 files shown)


ksunika:
should be looked at together with http://github.services.ooyala.net/Platform/mjolnir/pull/516

js/controller.js
```diff
@@ -36,6 +36,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       "adOverlayUrl": null,
       "showAdOverlay": false,
       "showAdOverlayCloseButton": false,
+      "adOverlayLoaded": false,
```
chibzu: Trying to think through the use case, do we actually need to store this state or just pass the message along the message bus? I'm not coming up with any reason the skin needs to care about the overlay being loaded or not aside from the check in adOverlay.js .

ksunika:
@chibzu, you are right. I removed the unnecessary state from controller.

chibzu:
Awesome, looks good.

PR #391, opened 2016-02-24
[PBW-4384] No sample text if language chosen has no locality settings

brianfwong:
Added ‘en’ as a default for the sample text if the results are blank.


mistermendez:
:+1:

PR #395, opened 2016-02-26
[PBW-internal] removing warnings by checkint NaN not passed to style …

ksunika:
…and adding keys to array elements


mistermendez:
:+1:

PR #398, opened 2016-02-26
[PBW-4437] fix for selecting closed captions on small screen

rchalooyala:


chibzu:
We lose the preview, is that intended for the small screen?

rchalooyala:
Yes, exactly. Here is why. 

cc in normal screen

chibzu:
Ah, hadn't scrunched the player down that far. Slick stuff.

PR #402, opened 2016-02-29
[PBW-4390] CC button staying on replay

brianfwong:
Ensure that the available languages is null when setting a new video.
We cannot do this in ‘Played’ because we cannot be sure a prior video
that has CC was fully played before a new video is set, so nulling the
list on setting an embed code, much as is done with availableBitrates.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -276,6 +276,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
     onEmbedCodeChanged: function(event, embedCode, options) {
       this.state.videoQualityOptions.availableBitrates = null;
       this.state.closedCaptionOptions.availableLanguages = null;
+      this.state.discoveryData = null;
 
       this.state.assetId = embedCode;
       $.extend(true, this.state.playerParam, options);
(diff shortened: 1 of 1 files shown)


js/controller.js
```diff
@@ -275,6 +275,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
 
     onEmbedCodeChanged: function(event, embedCode, options) {
       this.state.videoQualityOptions.availableBitrates = null;
+      this.state.closedCaptionOptions.availableLanguages = null;
```
ksunika: Looks like we should null everything we store from the previous video here. For example, discoveryData should not carryover as well.
@brianfwong, could you set the discoveryData to null as well here?

brianfwong: @ksunika Is it safe to wipe the discoveryData here? If we're watching a series of discovery videos, would that risk upsetting our place in them as each video loaded?

ksunika: From what I see, every time we switch to a new video, we are getting OO.EVENTS.DISCOVERY_API.RELATED_VIDEOS_FETCHED which replaces discoveryData. I was thinking in case no related videos available for some reason, we shouldn't be reusing the discovery data from the previous video.

ksunika:
looks good

PR #403, opened 2016-02-29
[PBW-4508] changing paths for social icons to absolute

ksunika:
Related PRs: 
https://github.com/ooyala/skin-config/pull/115
http://github.services.ooyala.net/Platform/player-v4/pull/19

### DIFF
--- a/scss/components/_share-screen.scss
+++ b/scss/components/_share-screen.scss
@@ -85,22 +85,22 @@
 
     .facebook {
       @extend .social-icon;
-      background-image: url("assets/images/social-facebook.svg");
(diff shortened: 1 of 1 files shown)


mistermendez:
why 4.1.4 and not 'latest' ?

ksunika:
@mistermendez, as I understand we are not planning to support stable/latest, it is going to be the exact versions. Am I wrong?
When we deploy to stable, the version is updated by build.sh in player-v4 (see http://github.services.ooyala.net/Platform/player-v4/pull/19) so it will be always the correct one for production deploys.

For master, I guess whenever we are changing images or paths, we will update the version as well.

rchalooyala:
^^correct

PR #405, opened 2016-03-01
[PBW-4490] Accept EVENT from VTC and set aspect ratio accordingly

rchalooyala:


js/controller.js
```diff
@@ -510,6 +510,13 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       this.resetUpNextInfo(false);
     },
 
+    onAssetDimensionsReceived: function(event, params) {
+      if (params.videoId == OO.VIDEO.MAIN) {
+        this.state.mainVideoAspectRatio = this.calculateAspectRatio(params.width, params.height);
```
ksunika: @rchalooyala, what happens if the aspect ratio is set explicitly in skin config, will we still receive the  ASSET_DIMENSION event and adjust automatically?

rchalooyala: nice catch, thanks

ksunika:
looks good :+1:

PR #406, opened 2016-03-01
[PBW-3542] autoPlay not working after UpNext countdown

brianfwong:
Added passing the playerParams to the SET_EMBED_CODE event.
(matching change to mjolnir mentioned in original check-in was NOT made as it was unnecessary.)


mistermendez:
:+1:

PR #417, opened 2016-03-08
PBW cleanup

mistermendez:
Main updates:

1.) reorder propTypes and defaultProps to JSX style guide
2.) Use Responsive Manager componentWidth instead of controlBarWidth
3.) Use Responsive Manager to detect resize
4.) remove unused code


js/components/scrubberBar.js
```diff
@@ -10,32 +10,13 @@ var React = require('react'),
...
-          }
-        }
-      }
-    };
-  },
-
   getInitialState: function() {
     this.isMobile = this.props.controller.state.isMobile;
-    this.responsiveUIMultiple = this.getResponsiveUIMultiple(this.props.responsiveView);
-    this.lastScrubX = null;
```
chibzu: lastScrubX is just bookkeeping, does it need to be in the React state? I don't think there's a reason to trigger a redraw when it changes.

js/components/scrubberBar.js
```diff
@@ -126,13 +98,13 @@ var ScrubberBar = React.createClass({
...
-      var scrubbingPlayheadX = this.props.currentPlayhead * this.scrubberBarWidth / this.props.duration + deltaX;
-      this.props.controller.updateSeekingPlayhead((scrubbingPlayheadX / this.scrubberBarWidth) * this.props.duration);
+      var deltaX = evt.clientX - this.state.lastScrubX;
+      var scrubbingPlayheadX = this.props.currentPlayhead * this.state.scrubberBarWidth / this.props.duration + deltaX;
+      this.props.controller.updateSeekingPlayhead((scrubbingPlayheadX / this.state.scrubberBarWidth) * this.props.duration);
       this.setState({
         scrubbingPlayheadX: scrubbingPlayheadX
       });
-      this.lastScrubX = evt.clientX;
+      this.state.lastScrubX = evt.clientX;
```
chibzu: From React's docs:

> NEVER mutate this.state directly, as calling setState() afterwards may replace the mutation you made. Treat this.state as if it were immutable.

js/components/scrubberBar.js
```diff
@@ -146,7 +118,7 @@ var ScrubberBar = React.createClass({
     evt.stopPropagation(); // W3C
     evt.cancelBubble = true; // IE
 
-    this.lastScrubX = null;
+    this.state.lastScrubX = null;
```
chibzu: Ditto here.

mistermendez:
@chibzu moved lastScrubX back to instance var

chibzu:
Wunderbar

PR #421, opened 2016-03-09
[PBW-4287] new slider

ksunika:


ksunika:
related PR: https://github.com/ooyala/skin-config/pull/122

js/components/controlBar.js
```diff
@@ -231,20 +186,12 @@ var ControlBar = React.createClass({
...
-    volumeSlider.push(
-      <div className="volumeSlider" ref="volumeSlider" onTouchEnd={this.handleVolumeBarTouchEnd} key={i}>
-        <div className="volumeIndicator" style={volumeIndicatorStyle}></div>
-        <div className="playheadPadding" style={volumeHeadPaddingStyle}
-          onTouchStart={this.handleVolumeHeadTouchStart}>
-          <div className="volumeHead"></div>
-        </div>
-      </div>);
+    var volumeSlider = <div className="volumeSlider"><Slider volumeSliderValue={parseFloat(this.props.controller.state.volumeState.volume)}
+                        changeVolumeSlider={this.changeVolumeSlider}
```
mistermendez: Let's use generic prop names, like:
  * `volumeSliderValue ` to `value`
  * `changeVolumeSlider ` to `onChange `

js/components/slider.js
```diff
@@ -0,0 +1,61 @@
...
+Utils = require('./utils');
+
+var Slider = React.createClass({
+
+  componentDidMount: function() {
+    this.handleSliderColoring();
+  },
+
+  componentDidUpdate: function() {
+    this.handleSliderColoring();
```
mistermendez: consider using `componentWillReceiveProps ` and check if the value changes, there are cases where `componentDidUpdate ` runs when there is no slider value change.

```
    if (nextProps.value != this.props.value) {
      this.handleSliderColoring();
    }
  },
```

js/components/slider.js
```diff
@@ -0,0 +1,61 @@
...
+            '-webkit-gradient(',
+              'linear, ',
+              'left top, ',
+              'right top, ',
+              'color-stop(' + value + ', '+ colorBeforeThumb + '), ',
+              'color-stop(' + value + ', '+ colorAfterThumb + ')',
+            ')'
+          ].join('');
+
+          // input.style.backgroundClip = 'content-box';
```
mistermendez: do we need this?

scss/base/_variables.scss
```diff
@@ -359,4 +359,23 @@ $scrubber-bar-padding-height:     24px !default;
 $scrubber-bar-left-right-padding: 15px !default;
 $min-player-width:                320px !default;
 
+//slider
+$thumb-color: yellow;
+$progress-color: green;
+$track-color: red;
+$slider-width: 10px;
```
mistermendez: are yellow, green, red used ?

ksunika:
@mistermendez, comments are addressed

mistermendez:
@ksunika conflicts

ksunika:
@mistermendez resolved

js/components/slider.js
```diff
@@ -0,0 +1,61 @@
...
+
+  componentWillReceiveProps: function(nextProps) {
+    if (nextProps.value != this.props.value) {
+      this.handleSliderColoring();
+    }
+  },
+
+  handleSliderColoring: function(){
+    if (!Utils.isEdge()){
+      var inputs = document.querySelectorAll('input[type=range]');
```
chibzu: This targets the entire page, and not just our components doesn't it?

ksunika: @chibzu, you are right!  I replaced it by  
var inputs = document.getElementsByClassName('slider');

chibzu: You fixed it before I could finish reviewing the whole thing, is there a reason we can't use refs instead of a DOM query? What's the case where we expect multiple sliders?

ksunika: Well we were thinking about using sliders for other screens as well, for example CC page. I can imagine several sliders there. Or one slider there and one in control bar.

chibzu: I think in that case changing the slider in the CC menu would then trigger the slider in the control bar to redraw as well, even though it hadn't changed.  Reassigning the oninput handler every time is a little strange as well, why not just perform the logic in oninput instead of assigning it and then triggering it?

ksunika: yes, good point, will rework this one.

ksunika:
@chibzu, I changed slider.js:

https://github.com/ooyala/html5-skin/commit/c9417c12788b4829ac020ec2146ae762a6f51522

chibzu:
Awesome :+1:

PR #422, opened 2016-03-10
[PBW-4549] Buffering events raised on Safari before clicking play whe…

brianfwong:
…n preloading

in onBuffering if the initialPlay is false and the screen to show is
the start screen then we are setting the this.state.buffering to false
else it will set it to true. Testing with up next showed no need for an
added check in setEmbedCode.


chibzu:
Looks good.

PR #424, opened 2016-03-10
[PBW-4542] We are unable to perform pause/play after closing overlay.

brianfwong:
Even though overlay is being closed, the plugin object is still present
in front of the UI layer of the skin in order to fix a separate PBI
issue that was allowing users to skip past ads. This fix keeps the
object, but makes it take up no screen space, so that the UI can be
interacted with. This new fix is to avoid breaking PBI-1503.


chibzu:
Solid.

PR #425, opened 2016-03-11
[PBW-4638] Fix Bitmovin iPad Fullscreen bug

mistermendez:
Bitmovin plugin adds wrapper elements around the main `video` element.

This was causing the main video element to be incorrectly set to a `div` instead of `video` and subsequently caused iOS iPad to fullscreen the `div` instead of `video`.


angieisrani:
What if the video is flash on desktop?  Video element won't be found

angieisrani:
Nevermind, I see the length check

PR #429, opened 2016-03-14
Debounce Resize

mistermendez:
* limit rate onResize func fires with resize listener
* the purpose of this update is to avoid costly calculations and component prop updates while window is resizing, see - 
https://lodash.com/docs#debounce


rchalooyala:
Cool!

PR #431, opened 2016-03-16
[PBW-4484] Fixing conflict with SHOW_AD_CONTROLS

chibzu:


js/controller.js
```diff
@@ -202,7 +202,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         this.state.configLoaded = true;
         this.renderSkin();
 
-        var fullClass = (this.state.config.adScreen.showControlBar ? "" : " full");
+        var fullClass = (this.state.config.adScreen.showControlBar && this.state.showAdControls ? "" : " full");
```
ksunika: @chibzu, do we need to check this here? It looks like the below add/remove class takes care of everything.

chibzu: Without knowing the timing of when the SHOW_AD_CONTROLS message comes through, I assumed this was the safer way to do it. Will take another look.

aldigun:
LGTM

PR #435, opened 2016-03-17
[PBW-4243] Change cursor to pointer over scrubber bar

brianfwong:
Changes cursor to pointer over scrubber bar.


ksunika:
looks good!

chibzu:
Nice

PR #436, opened 2016-03-17
[PBW-4321] cms-less flow for the skin

ksunika:
Works with http://github.services.ooyala.net/Platform/mjolnir/pull/538

-subscribed to ASSET_CHANGED
-some parameters stored in content_tree
-is_live_asset parameter stored in isLiveAsset (don't need anything else from authorization)


chibzu:
:+1:

PR #451, opened 2016-03-31
[PBW-4006] fixes description set to false breaking the pause screen

ksunika:
- check if description is rendered before calling Utils.truncateTextToWidth
- using different skin config parameters for start and pause screen


mistermendez:
:+1:

PR #452, opened 2016-04-01
[PBW-4937] Use chrome's fullscreen change for resizing as well.

chibzu:


mistermendez:
should we also removeEventListener for webkitfullscreenchange ?

js/mixins/componentWidthMixin.js
```diff
@@ -10,6 +10,7 @@ var ComponentWidthMixin = {
 
   componentDidMount: function() {
     window.addEventListener('resize', debounce(this.onResize, 150));
+    window.addEventListener('webkitfullscreenchange', debounce(this.onResize, 150));
     this.setState({
```
mistermendez: consider using screenfull for cross-browser js fullscreen API -

```
Fullscreen = require('screenfull');

window.addEventListener(Fullscreen.raw.fullscreenchange, debounce(this.onResize, 150));
```

mistermendez:
:+1:

js/mixins/componentWidthMixin.js
```diff
@@ -10,6 +11,7 @@ var ComponentWidthMixin = {
 
   componentDidMount: function() {
     window.addEventListener('resize', debounce(this.onResize, 150));
+    window.addEventListener(screenfull.raw.fullscreenchange, debounce(this.onResize, 150));
```
rchalooyala: The player won't load on iPhone with the latest html5-skin. I believe that fullscreenchange not being valid is the issue. Maybe a simple check for null is all that is needed. Could you take a look?

PR #453, opened 2016-04-04
[REVERT][PBW-4145] Revert the previous commit b63d085cfc415eb4fbfe3d3…

brianfwong:
[REVERT][PBW-4145] Revert the previous commit b63d085cfc415eb4fbfe3d326f87ad92927ced85

Prior solution was not ideal. Fix should be made to the test page only. Prior solution has already been merged.


mistermendez:
:+1:

PR #460, opened 2016-04-06
[PBW-4347] discovery for unbundled player

ksunika:
works with https://git.corp.ooyala.com/projects/PBW/repos/mjolnir/pull-requests/11


chibzu:
Looks good

PR #461, opened 2016-04-07
[PBW-4937] Screenfull API is inconsistent on Safari and we only care …

chibzu:
…about Chrome for this event listener, switching back to native event.


chibzu:
Fix for
https://jira.corp.ooyala.com/browse/PBW-4763

aldigun:
looks good

PR #462, opened 2016-04-07
[PBW-4575] CC relative position

mistermendez:
to avoid overlap with other UI components


rchalooyala:
Tested on sandbox and looks really good!

PR #465, opened 2016-04-08
[PBW-4810] Add new container class name

mistermendez:
* main purpose is to enforce min player width of 320px


mistermendez:
FYI @chibzu

chibzu:
Looks good.

PR #472, opened 2016-04-14
[PBW-4989] Scrubber bar style updates. Include: Updating JSON - scrub…

mistermendez:
… bar. Play-head, buffer, duration


mistermendez:
:+1:

PR #476, opened 2016-04-15
[PBW-5035] replace social icons

ksunika:
the changes in icons will only be noticeable after the new stable is out, since _share-screen.scss points to stable.


mistermendez:
screen shot 2016-04-18 at 5 49 46 pm

mistermendez:
👍

PR #477, opened 2016-04-18
[PBW 4988] Updated background style to control bar.

Jgkelle26:


mistermendez:
* we need to move the image path to skin.json
  * also consider using css or inline .svg to reduce additional http request
  * and let's keep the background image and color as separate values, i.e -

```
background-image: url("paper.svg");
background-color: $control-bar-bg;
```

Jgkelle26:
Ok cool, the idea for having the svg would be so later I could create a more complex gradient. One that I do not think CSS can manage. 

So I guess reject the branch then or something? And I'll work on it. 

I do not know why the spinner is on this branch I created two different branches and two different PRs

PR #478, opened 2016-04-18
[PBW 5055] Spinner speed updated from 1.6 to 0.6

Jgkelle26:


mistermendez:
This should only include the update to _spinner.scss.

There are updates to the control bar that shouldn't be on this branch.

mistermendez:
Let's fix, then reopen

PR #484, opened 2016-04-19
[PBW-4988] Control bar gradient CSS componet

Jgkelle26:
Part 2 of 2

### DIFF
--- a/scss/components/_control-bar.scss
+++ b/scss/components/_control-bar.scss
@@ -5,7 +5,7 @@
   bottom: 0;
 }
 .control-bar {
-  background: rgba($black, 0.3);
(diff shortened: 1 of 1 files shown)


scss/components/_control-bar.scss
```diff
@@ -5,7 +5,7 @@
   bottom: 0;
 }
 .control-bar {
-  background: rgba($black, 0.3);
+  @include vendor-prefixes (background, $control-bar-gradient);
   width: 100%;
```
mistermendez: let's keep original background to support older browsers -
```
background: rgba($black, 0.3);
@include vendor-prefixes (background, $control-bar-gradient);
```

mistermendez:
also PR w/ 'master' branch

PR #485, opened 2016-04-20
[PBW-4997] Removing setTimeout calls for a proper callback.

chibzu:


js/controller.js
```diff
@@ -1055,10 +1064,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         if (this.state.playerState == CONSTANTS.STATE.PLAYING){
           this.mb.publish(OO.EVENTS.PAUSE);
         }
-        setTimeout(function() {
-          this.state.screenToShow = screen;
-          this.renderSkin();
-        }.bind(this), 1);
+
```
ksunika: @chibzu, it looks like we can't switch to quality screen if we remove this.

chibzu:
Good catch, forgot quality has a different interaction on a small player, thanks.

ksunika:
looks good!

PR #492, opened 2016-04-21
Change to show pause screen on pause in iPhone rather than start screen

brianfwong:


mistermendez:
👍

brianfwong:
This was from PBW-4826

PR #501, opened 2016-04-25
[PBW-4456] end seeking if switching screen from end to pause when see…

ksunika:
…king on the end screen

### DIFF
--- a/js/components/scrubberBar.js
+++ b/js/components/scrubberBar.js
@@ -132,11 +132,13 @@ var ScrubberBar = React.createClass({
       document.removeEventListener("touchend", this.handlePlayheadMouseUp, true);
     }
     this.props.controller.seek(this.props.currentPlayhead);
-    this.setState({
-      currentPlayhead: this.props.currentPlayhead,
-      scrubbingPlayheadX: 0
-    });
-    this.props.controller.endSeeking();
+    if (this.isMounted()) {
+      this.setState({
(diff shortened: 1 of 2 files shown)


js/components/scrubberBar.js
```diff
@@ -132,11 +132,13 @@ var ScrubberBar = React.createClass({
       document.removeEventListener("touchend", this.handlePlayheadMouseUp, true);
     }
     this.props.controller.seek(this.props.currentPlayhead);
-    this.setState({
-      currentPlayhead: this.props.currentPlayhead,
-      scrubbingPlayheadX: 0
-    });
-    this.props.controller.endSeeking();
+    if (this.isMounted()) {
```
ksunika: After seeking from the end screen we get to the pause screen and end screen scrubber bar becomes unmounted causing console warning to appear.

js/controller.js
```diff
@@ -1000,6 +1000,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       this.mb.publish(OO.EVENTS.SEEK, seconds);
       if (this.state.screenToShow == CONSTANTS.SCREEN.END_SCREEN) {
         this.state.pauseAnimationDisabled = true;
+        this.endSeeking();
```
ksunika: to ensure playhead is rendered correctly when transitioning from end screen to pause screen

PR #507, opened 2016-04-28
[PBW-5158] resolve - Cannot read 'addEventListener' of undefined

mistermendez:
Resolve - Uncaught exception TypeError: Cannot read property 'addEventListener' of undefined

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -269,7 +269,10 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         element = elementVideo;
       }
 
-      element.get(0).addEventListener("loadedmetadata", this.metaDataLoaded.bind(this));
+      //add loadedmetadata event listener to main video element
+      if (element.get(0)) {
+        element.get(0).addEventListener("loadedmetadata", this.metaDataLoaded.bind(this));
+      }
 
       if (Utils.isIE10()) {
         element.attr("controls", "controls");


js/controller.js
```diff
@@ -269,7 +269,10 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         element = elementVideo;
```
pilievOoyala: Do you need to do any check if "element" exists? A lot of the function depends on it. Also, do you need to do something else if element or element.get(0) doesn't exist or just let it continue on?

mistermendez: * yup just let it continue, we can't add event listener w/o main video element

chibzu: Isn't the point of this event that the element should exist? Do we understand why we're getting an event without the element being instantiated?

pilievOoyala: I've already talked to Angie about it. I don't recall what the final outcome was, but there is already a ticket for why it's not instantiated.  So this checkin should just be concerned with what to do if we get bad data and being able to fail gracefully.

chibzu:
If that's the case, this will fail gracefully. Since the element is grabbed with JQuery it's wrapped so it won't throw exceptions even though there isn't an element there.

chibzu:
I believe the concerns here have been addressed, lgtm.

PR #508, opened 2016-04-29
[PBW-4969] Replay when seeking on end screen

ksunika:
Should be merged with corresponding changes in mjolnir and video-plugins-internal:
https://git.corp.ooyala.com/projects/PBW/repos/video-plugins-internal/pull-requests/14
https://git.corp.ooyala.com/projects/PBW/repos/mjolnir/pull-requests/41


js/controller.js
```diff
@@ -997,13 +997,12 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
     seek: function(seconds) {
-      this.mb.publish(OO.EVENTS.SEEK, seconds);
-      if (this.state.screenToShow == CONSTANTS.SCREEN.END_SCREEN) {
-        this.state.pauseAnimationDisabled = true;
+      if (this.state.playerState == CONSTANTS.STATE.END) {
         this.endSeeking();
-        this.state.screenToShow = CONSTANTS.SCREEN.PAUSE_SCREEN;
-        this.state.playerState = CONSTANTS.STATE.PAUSE;
-        this.renderSkin();
+        this.mb.publish(OO.EVENTS.REPLAY, seconds);
```
aldigun: Hmm I thought you also need to make sure that after replay, you need to seek to the correct time?

ksunika: the modified REPLAY in core takes the optional parameter in seconds, and then seeks to that. This way we do not seek twice, first to 0 within the replay, and then to the right offset. If we first seek to 0, the preroll ads start kicking in.

aldigun: got it.

PR #511, opened 2016-05-03
startScreen.js - removed default linear-gradient

boxyardboy:
Updated the StartScreen posterStyle; removed the default linear-gradient. It would be better to declare this style property via the style sheet (not inline) as it is not easily removed without also removing the poster image.


mistermendez:
great suggestion. we still need the background gradient, I moved the image and gradient to different elements for better customization. 

see PR here - https://github.com/ooyala/html5-skin/pull/533

PR #522, opened 2016-05-12
[PBW-5214] Update time counter on control bar Live stream

mityaha:
* Don't display current time for live stream unless we shifted to the past in which case show negative value related to current time


ksunika:
looks good

PR #524, opened 2016-05-12
Pbw 3946 CMSless work

chibzu:
Integrating current CMSless work


mistermendez:
LGTM, live stream works, assets change, unit tests pass 👍

PR #525, opened 2016-05-12
[PBW-4954] [Alice] Trigger seek-to-end (duration) when live button is clicked

mityaha:
* Converted "live" div to button, added click handler for this button

### DIFF
--- a/js/components/controlBar.js
+++ b/js/components/controlBar.js
@@ -63,6 +63,14 @@ var ControlBar = React.createClass({
     this.props.controller.toggleFullscreen();
   },
 
+  handleLiveClick: function(evt) {
+    evt.stopPropagation();
+    evt.cancelBubble = true;
+    evt.preventDefault();
+    // add 2 more seconds to compensate for time to process this request
+    this.props.controller.seek(-this.props.duration + 2);
+  },
+
   handleVolumeIconClick: function(evt) {
     if (this.isMobile){
       this.props.controller.startHideControlBarTimer();
@@ -210,8 +218,8 @@ var ControlBar = React.createClass({
     var isLiveStream = this.props.authorization.streams[0].is_live_stream;
     var videoQualityPopover = this.state.showVideoQualityPopover ? <VideoQualityPopover {...this.props} togglePopoverAction={this.toggleQualityPopover}/> : null;
     var durationSetting = {color: this.props.skinConfig.controlBar.iconStyle.inactive.color};
-    var playheadTimeContent = isLiveStream ? (Math.abs(this.props.duration - this.props.currentPlayhead < 0.7) ? null :
-                                              Utils.formatSeconds(this.props.currentPlayhead - this.props.duration)) : playheadTime;
+    var timeShift = this.props.currentPlayhead - this.props.duration;
+    var playheadTimeContent = isLiveStream ? (Math.abs(timeShift) < 2 ? null : Utils.formatSeconds(timeShift)) : playheadTime;
     var totalTimeContent = isLiveStream ? null : <span className="oo-total-time">{totalTime}</span>;
 
     // TODO: Update when implementing localization
@@ -230,13 +238,11 @@ var ControlBar = React.createClass({
           onMouseOver={this.highlight} onMouseOut={this.removeHighlight}/>
       </button>,
 
-      "live": <div className="oo-live oo-control-bar-item" key="live">
-        <div className="oo-live-indicator">
-          <div className="oo-live-circle"></div>
-          <span className="oo-live-text"> {liveText}</span>
-        </div>
-      </div>,
-
+      "live": <button className="oo-control-bar-item" onClick={this.handleLiveClick} key="live">
+        <div className="oo-live-circle"></div>
+        <span className="oo-live-text"> {liveText}</span>
+      </button>,
+      
       "volume": <div className="oo-volume oo-control-bar-item" key="volume">
         <Icon {...this.props} icon={volumeIcon} ref="volumeIcon"
           style={this.props.skinConfig.controlBar.iconStyle.inactive}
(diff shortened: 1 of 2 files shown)


js/components/controlBar.js
```diff
@@ -230,13 +238,11 @@ var ControlBar = React.createClass({
...
       </button>,
 
-      "live": <div className="oo-live oo-control-bar-item" key="live">
-        <div className="oo-live-indicator">
-          <div className="oo-live-circle"></div>
-          <span className="oo-live-text"> {liveText}</span>
-        </div>
-      </div>,
-
+      "live": <button className="oo-control-bar-item" onClick={this.handleLiveClick} key="live">
```
ksunika: removing 'oo-live' breaks unit tests, also I think QA is using these unique class names in automation to check for elements present

ksunika: <div className="oo-live-indicator"> missing causes misalignment issues.

js/components/controlBar.js
```diff
@@ -63,6 +63,14 @@ var ControlBar = React.createClass({
     this.props.controller.toggleFullscreen();
   },
 
+  handleLiveClick: function(evt) {
+    evt.stopPropagation();
+    evt.cancelBubble = true;
+    evt.preventDefault();
+    // add 2 more seconds to compensate for time to process this request
+    this.props.controller.seek(-this.props.duration + 2);
```
ksunika: why are we going to negative duration?

js/components/controlBar.js
mityaha: Fixed

js/components/controlBar.js
mityaha: Fixed

js/components/controlBar.js
```diff
@@ -63,6 +63,22 @@ var ControlBar = React.createClass({
...
+    evt.stopPropagation();
+    evt.cancelBubble = true;
+    evt.preventDefault();
+    this.props.controller.seek(this.props.duration);
+  },
+
+  checkLiveCursor: function() {
+    var timeShift = this.props.currentPlayhead - this.props.duration;
+    if (Math.abs(timeShift) < 2) {
+      ReactDOM.findDOMNode(this.refs.LiveButton).style.cursor="default";
```
ksunika: We are trying not to change styling outside of css when possible. Here, for example, instead of changing the style you could change the class name for the element (using ClassNames, see volume as an example) depending on its status. You can do it in populateControlBar, and remove onMouseOver altogether.

    var liveClass = ClassNames({
        "oo-control-bar-item oo-live oo-live-indicator": true,
        "oo-live-nonclickable": Math.abs(this.props.currentPlayhead - this.props.duration) < 2
      });

and then set 'cursor: default' in css for oo-live-nonclickable

mityaha: Can you please do it?  I won't be available for a while.  Thanks!

On Friday, May 13, 2016, Xenia Serbin <notifications@github.com> wrote:

> In js/components/controlBar.js
> <https://github.com/ooyala/html5-skin/pull/525#discussion_r63223605>:
>
> > @@ -63,6 +63,22 @@ var ControlBar = React.createClass({
> >      this.props.controller.toggleFullscreen();
> >    },
> >
> > +  handleLiveClick: function(evt) {
> > +    evt.stopPropagation();
> > +    evt.cancelBubble = true;
> > +    evt.preventDefault();
> > +    this.props.controller.seek(this.props.duration);
> > +  },
> > +
> > +  checkLiveCursor: function() {
> > +    var timeShift = this.props.currentPlayhead - this.props.duration;
> > +    if (Math.abs(timeShift) < 2) {
> > +      ReactDOM.findDOMNode(this.refs.LiveButton).style.cursor="default";
>
> We are trying not to change styling outside of css when possible. Here,
> for example, instead of changing the style you could change the class name
> for the element (using ClassNames, see volume as an example) depending on
> its status. You can do it in populateControlBar, and remove onMouseOver
> altogether.
>
> var liveClass = ClassNames({
>     "oo-control-bar-item oo-live oo-live-indicator": true,
>     "oo-live-nonclickable": Math.abs(this.props.currentPlayhead - this.props.duration) < 2
>   });
>
> and then set 'cursor: default' in css for oo-live-nonclickable
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/ooyala/html5-skin/pull/525/files/36666e08f032caab3a52688a50887e1ebf81eabb#r63223605>
>


-- 
-- 
Dimitry Nasledov
dima@ooyala.com

js/components/controlBar.js
```diff
@@ -224,16 +215,24 @@ var ControlBar = React.createClass({
 
     var playheadTime = isFinite(parseInt(this.props.currentPlayhead)) ? Utils.formatSeconds(parseInt(this.props.currentPlayhead)) : null;
     var isLiveStream = this.props.authorization.streams[0].is_live_stream;
-    var videoQualityPopover = this.state.showVideoQualityPopover ? <VideoQualityPopover {...this.props} togglePopoverAction={this.toggleQualityPopover}/> : null;
     var durationSetting = {color: this.props.skinConfig.controlBar.iconStyle.inactive.color};
     var timeShift = this.props.currentPlayhead - this.props.duration;
-    // checking timeShift < 2 seconds (not == 0) as processing of the click after we rewinded and then went live may take some time
-    var playheadTimeContent = isLiveStream ? (Math.abs(timeShift) < 2 ? null : Utils.formatSeconds(timeShift)) : playheadTime;
+    // checking timeShift < 1 seconds (not == 0) as processing of the click after we rewinded and then went live may take some time
+    var isLiveNow = Math.abs(timeShift) < 1;
```
ksunika: @mityaha, i thought if we get farther than one second then we can as well just show the real time difference? I do not insist though.

chibzu:
Looks like all of Xenia's critiques were addressed and it's looking good to me.

PR #534, opened 2016-05-17
fix playhead centering

mistermendez:
FYI @Jgkelle26


Jgkelle26:
Just an FYI, I changed the values from 50% to their current because the playhead was not centered after making adjustments to the scrubber-bar-padding and scrubber-bar

mistermendez:
try w/ this update, playhead should be centered w/ 50%

Jgkelle26:
giphy 4

PR #544, opened 2016-05-24
[PBW-5275] [Alice][Live Stream] Showing NAN:NAN if Live Stream assets are scrubbed

mityaha:
* For html5 live HLS assets main_html5 reports duration == 0 and scrubbaer bar shouldn't be doing calculations


aldigun:
LGTM

PR #545, opened 2016-05-25
[PBW-5291] Close the email window if it doesn't redirect to a web cli…

chibzu:
…ent.

### DIFF
--- a/js/components/sharePanel.js
+++ b/js/components/sharePanel.js
@@ -65,7 +65,16 @@ var SharePanel = React.createClass({
     mailToUrl += "?subject=" + encodeURIComponent(this.props.contentTree.title);
     mailToUrl += "&body=" + encodeURIComponent(emailBody + location.href);
     //location.href = mailToUrl; //same window
-    window.open(mailToUrl, "_blank", "height=315,width=780"); //new window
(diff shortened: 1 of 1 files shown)


js/components/sharePanel.js
```diff
@@ -65,7 +65,16 @@ var SharePanel = React.createClass({
     mailToUrl += "?subject=" + encodeURIComponent(this.props.contentTree.title);
     mailToUrl += "&body=" + encodeURIComponent(emailBody + location.href);
     //location.href = mailToUrl; //same window
-    window.open(mailToUrl, "_blank", "height=315,width=780"); //new window
+    var emailWindow = window.open(mailToUrl, "email", "height=315,width=780"); //new window
+    setTimeout(function(){
+      try {
+         // If we can't access href, a web client has taken over and this will throw
+         // an exception, preventing the window from being closed.
+        var test = emailWindow.location.href;
```
mityaha: Shouldn't here be a check for test being defined?

chibzu: This line is just fishing for a security exception, it doesn't actually matter what the value of href is. Test isn't needed anywhere else.

mityaha:
LG

PR #546, opened 2016-05-25
[PBW-5292] Handle boolean options parameters as booleans and not stri…

chibzu:
…ngs.

### DIFF
--- a/iframe.html
+++ b/iframe.html
@@ -44,7 +44,13 @@
           }
           if (typeof currentLevel === 'string')
             throw("Exception: Invalid query parameter");
-          currentLevel[keySplits[0]] = decodeURIComponent(match[2]);
+
+          if (decodeURIComponent(match[2]).toLowerCase() === "true")
+            currentLevel[keySplits[0]] = true;
+          else if (decodeURIComponent(match[2]).toLowerCase() === "false")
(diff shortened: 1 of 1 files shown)


iframe.html
```diff
@@ -44,7 +44,13 @@
...
           if (typeof currentLevel === 'string')
             throw("Exception: Invalid query parameter");
-          currentLevel[keySplits[0]] = decodeURIComponent(match[2]);
+
+          if (decodeURIComponent(match[2]).toLowerCase() === "true")
+            currentLevel[keySplits[0]] = true;
+          else if (decodeURIComponent(match[2]).toLowerCase() === "false")
+            currentLevel[keySplits[0]] = false;
+          else
+            currentLevel[keySplits[0]] = decodeURIComponent(match[2]);
```
mityaha: In the last "else" case parameters will still be strings - isn't it a contradiction?  What are possible values in this case?

chibzu: At the point we get to the if statements, currentLevel should be the object we're writing a key/value pair to, so I switched that to make sure that currentLevel is an object to be more concise. Then keySplits[0] is the last token from the query param and match[2] is the value of from the query param.

aldigun:
LGTM

PR #548, opened 2016-05-31
[PBW-5252] Unsubscribe from playback_ready on error.

brianfwong:


aldigun:
This is not right. This is essentially blocking the player.
The player has to be able to recover from errorScreen if a new embedCode is set.
This need to pass setEmbedCode with a working embedCode. I'm pretty sure PLAYBACK_READY is still needed to unblock player.

brianfwong:
@aldigun Moved PLAYBACK_READY subscription to subscribeBasicPlaybackEvents as a counterpoint to being in unsubscribe on errors. This ensures that if there is something to recover to, the player can do so.
Firing off something like:
pp.setEmbedCode("Axamo5MDE6xNoMVoLcOff-alUnRBrLbV")
after the player has hit the error screen now works, with the player able to recover and play the error-free stream.

js/controller.js
```diff
@@ -126,7 +126,6 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       this.mb.subscribe(OO.EVENTS.THUMBNAILS_FETCHED, 'customerUi', _.bind(this.onThumbnailsFetched, this));//xenia: to be replaced by a more appropriate event
       this.mb.subscribe(OO.EVENTS.AUTHORIZATION_FETCHED, 'customerUi', _.bind(this.onAuthorizationFetched, this));
       this.mb.subscribe(OO.EVENTS.ASSET_CHANGED, 'customerUi', _.bind(this.onAssetChanged, this));
-      this.mb.subscribe(OO.EVENTS.PLAYBACK_READY, 'customerUi', _.bind(this.onPlaybackReady, this));
```
aldigun: Let's not remove this. This is a critical piece of event. Do you know what will happen if you set 2 subscribers at the same time on the same event? I think it will not do a double call. Can you make sure that only a single call is made?

brianfwong: Yes I removed it not wanting the double call. But it was a little worrisome. In the past I've had double calls around in in-work code with no 'apparent' problems but if such code were to go live it would likely create an analytics mess. I'll make sure to put things into a conditional to prevent double calls.

brianfwong:
@aldigun Added a status around manipulating PLAYBACK_READY subscription to prevent it from being double subscribed.

brianfwong:
@aldigun PBW-5252 pull-request code has been rewritten and updated based on earlier feedback to use a different solution, and needs re-review.

js/controller.js
```diff
@@ -115,7 +115,12 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       "forceControlBarVisible": false,
       "timer": null,
       "errorCode": null,
-      "isSubscribed": false,
+
+      "isSubscribed": {
```
aldigun: Almost there. But I think we can have 2 separate flags. That means you do not need to create an Object out of isSubscribed, and make isSubscribed as-is.

brianfwong: @aldigun Ok. I've moved it to two separate flags, leaving the "isSubscribed" flag as it is on master.

brianfwong:
@aldigun  I've moved it to two separate flags, leaving the "isSubscribed" flag as it is on master.

aldigun:
LGTM

PR #554, opened 2016-06-08
update skin config

rchalooyala:


rchalooyala:
Self merge, just updating skin config

PR #555, opened 2016-06-09
[PBW-5401] move cc switch container so title doesnt overlap

rchalooyala:


Jgkelle26:
http://gph.is/1peMWbm

Jgkelle26:
giphy 10

PR #565, opened 2016-06-15
Responsive design edits

rchalooyala:


scss/responsive/_small.scss
```diff
@@ -71,18 +77,35 @@
...
+        top: calc(50% + 42px);
+      }
       .oo-selection-container-title {
         margin-bottom: 0px !important;
       }
     }
   }
+  //CC Preview Panel
+  .oo-preview-panel {
+    display: none;
```
aldigun: preview display is set to none?

rchalooyala: Correct, we are removing the cc-preview on small and extra small devices because there is just not enough room.

PR #568, opened 2016-06-16
Pbi 1808

aeng7:
Support for live ads, where ad time remaining is calculated based on unix time rather than playheads

### DIFF
--- a/js/components/adPanel.js
+++ b/js/components/adPanel.js
@@ -140,4 +140,21 @@ var AdPanel = React.createClass({
     );
   }
 });
+
+AdPanel.defaultProps = {
+  currentPlayhead: 0,
+  adVideoDuration: 0,
+  adStartTime: 0,
+  currentAdsInfo: {
+    numberOfAds: 0,
+    skipAdButtonEnabled: false,
+    currentAdItem: {
+      hasClickUrl: false,
+      name: "",
+      indexInPod: 0,
+      isLive: false
+    }
+  }
+};
+
 module.exports = AdPanel;
--- a/js/components/adPanel.js
+++ b/js/components/adPanel.js
@@ -72,7 +72,16 @@ var AdPanel = React.createClass({
       adPlaybackInfo = adPlaybackInfo + ": (" + currentAdIndex + "/" + totalNumberOfAds + ")";
     }
 
-    var remainingTime = Utils.formatSeconds(Math.max(0, parseInt(this.props.adVideoDuration - this.props.currentPlayhead)));
+    var isLive = this.props.currentAdsInfo.currentAdItem.isLive;
+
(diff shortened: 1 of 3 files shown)


js/components/adPanel.js
```diff
@@ -72,7 +72,16 @@ var AdPanel = React.createClass({
       adPlaybackInfo = adPlaybackInfo + ": (" + currentAdIndex + "/" + totalNumberOfAds + ")";
     }
 
-    var remainingTime = Utils.formatSeconds(Math.max(0, parseInt(this.props.adVideoDuration - this.props.currentPlayhead)));
+    var isLive = this.props.currentAdsInfo.currentAdItem.isLive;
```
pilievOoyala: need a big old if check. I know it's not being done earlier in the function but we should start adding it in.

js/components/adPanel.js
```diff
@@ -72,7 +72,16 @@ var AdPanel = React.createClass({
...
+
+    var remainingTime;
+    if (isLive) {
+      remainingTime = parseInt((this.props.adStartTime + this.props.adVideoDuration * 1000 - new Date().getTime())/1000);
```
pilievOoyala: can you surround these remainingTime calculations with a check for this.props?  Also, what if the number don't exist and we get a NaN. Can you do a check at the end and set it to 0 if it's NaN

js/controller.js
```diff
@@ -637,6 +638,11 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         this.state.isPlayingAd = true;
         this.state.currentAdsInfo.currentAdItem = adItem;
         this.state.playerState = CONSTANTS.STATE.PLAYING;
+        if (adItem.isLive) {
```
pilievOoyala: need a check at the beginning of this to make sure this.state exists.

js/skin.js
```diff
@@ -212,6 +212,7 @@ var Skin = React.createClass({
               responsiveView={this.state.responsiveId}
               componentWidth={this.state.componentWidth}
               videoQualityOptions={this.state.videoQualityOptions}
```
pilievOoyala: same thing here.  Sorry for making you do this but we have enough problems with null checks. May as well fix ones we see.

js/components/adPanel.js
chibzu: this.props is guaranteed to exist thanks to React. You can also use defaultProps to guarantee that there's something in that field, rather than adding if statements and null checks everywhere if data consistency is a concern.

js/controller.js
chibzu: Same thing here, this.state exists and can be populated with getInitialState

chibzu:
LGTM

PR #571, opened 2016-06-20
[PBW-5352] Fixes setting pp.setTargetBitrate(ID) through console

brianfwong:
See also:
https://git.corp.ooyala.com/projects/PBW/repos/mjolnir/pull-requests/88/overview
Pull request in mjolnir for this same issue

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -86,7 +86,8 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
 
       "videoQualityOptions": {
         "availableBitrates": null,
-        "selectedBitrate": null
+        "selectedBitrate": null,
+        "showVideoQualityPopover":false,
       },
 
       "volumeState": {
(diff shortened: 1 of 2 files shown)


js/controller.js
```diff
@@ -1163,6 +1165,27 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+      this.renderSkin({
+          "videoQualityOptions": {
+            "availableBitrates": this.state.videoQualityOptions.availableBitrates,
+            "selectedBitrate": this.state.videoQualityOptions.selectedBitrate,
+            "showVideoQualityPopover":this.state.videoQualityOptions.showVideoQualityPopover
+          }
+        });
+      if(this.state.videoQualityOptions.showVideoQualityPopover == true) {
+        this.toggleVideoQualityPopOver();
+      }
```
chibzu: Why is the state being toggled after the render call?

brianfwong: The render call was not changing the visible info, but merely updating what should change on the next redraw. The button needs to be toggled off and on for the change to be seen.

chibzu: Second thing, why the new event and not using SET_TARGET_BITRATE? I don't see any difference in how the events are being differentiated.

brianfwong: This was some time ago, but from the ticket I have this:

On calling pp.setTargetBitrate(1), the control directly goes to video controller without updating the skin. For this we have to publish a new event for sendVideoQualityChangeEvent() function in the controller.js which will set the new bitrate.
Tested creating an event listener in controller.js and published the event from setTargetBitrate() and also initialized the new event listener in the Constants.js of bit wrapper but getting an error.

So we made a new event because SET_TARGET_BITRATE was not available in this scope.

chibzu: All right, sounds good.

mistermendez:
FYI this broke the control bar unit test -

`FAIL  tests/components/controlBar-test.js`

PR #584, opened 2016-07-01
[PBW-5535] Thumbnails from previous asset are shown on the next asset without thumbnails

mityaha:
* Clear thumbnails data in onEmbedCodeChanged callback


chibzu:
LGTM

PR #585, opened 2016-07-05
[PBW-3876] Allow share screen to work on the START screen(When the vi…

brianfwong:
[PBW-3876] Allow share screen to work on the START screen(When the video is loading)


chibzu:
LGTM

PR #587, opened 2016-07-06
used controlbar height from skin.json

shruthi17:
This is related to : https://github.com/ooyala/skin-config/pull/145


mistermendez:
👍

config
```diff
@@ -1 +1 @@
-Subproject commit 64d81ceeb72ff87e4f941466632af9d5a9847a97
+Subproject commit 2fd9ad1f0ac89789e6c64df477661a399f970d08
```
mityaha: Why this change?  Did something else has changed in config repository that requires its update here?

PR #601, opened 2016-07-15
[PBW-5131] More work

mityaha:
Fixed a bug when on mouse down without moving smaill images were staggered on top of central image, fixed unit test problem with padding parameter

### DIFF
--- a/js/components/thumbnailCarousel.js
+++ b/js/components/thumbnailCarousel.js
@@ -15,17 +15,18 @@ var ThumbnailCarousel = React.createClass({
       thumbnailHeight: 0,
       centerThumbnailWidth: 0,
       centerThumbnailHeight: 0,
-      thumbnailPadding: 0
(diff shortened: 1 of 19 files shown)


js/components/thumbnailCarousel.js
```diff
@@ -27,19 +27,22 @@ var ThumbnailCarousel = React.createClass({
                    centerThumbnailHeight: carousel.clientHeight});
   },
 
+  componentWillReceiveProps: function(nextProps) {
```
mistermendez: why is this here? It will always set `this.props.update` to false

mityaha: This is a workaround for conflict with test code, I was telling you about it on Friday.  This method is called by real code, but it is not called by test code.  So, after it is set to false, real code will be just comparing properties, i.e. do the right job.

mityaha:
Looks good

PR #606, opened 2016-07-19
[PBW-5528]Discovery Up next appears fix after starting video playback

shruthi17:
Also, tested on the devices on which this manifested earlier and are not reproducible further.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -419,7 +419,8 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       }
       this.state.upNextInfo.timeToShow = timeToShow;
 
-       if (duration!==0 && duration- currentPlayhead <= timeToShow &&
+      if (duration !==0 &&
+        duration - currentPlayhead <= timeToShow &&
         !this.state.upNextInfo.countDownCancelled &&
         this.state.isPlayingAd !== true &&
         this.state.upNextInfo.upNextData !== null && (this.state.playerState === CONSTANTS.STATE.PLAYING || this.state.playerState === CONSTANTS.STATE.PAUSE)) {
--- a/js/controller.js
+++ b/js/controller.js
@@ -406,7 +406,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       }
     },
 
-    howshowUpNextScreenWhenReady: function(currentPlayhead, duration) {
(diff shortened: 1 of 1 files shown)


js/controller.js
```diff
@@ -406,7 +406,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       }
     },
 
-    showUpNextScreenWhenReady: function(currentPlayhead, duration) {
+    howshowUpNextScreenWhenReady: function(currentPlayhead, duration) {
```
mistermendez: TYPO: this should be `showUpNextScreenWhenReady`

js/controller.js
```diff
@@ -420,7 +420,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
 
       this.state.upNextInfo.timeToShow = timeToShow;
 
-      if (duration - currentPlayhead <= timeToShow &&
+if (duration!==0 &&- currentPlayhead <= timeToShow &&
```
mistermendez: Should this be?

```
if (duration !== 0 &&
  duration - currentPlayhead <= timeToShow &&
```

mistermendez: Also fix the indentation

js/controller.js
shruthi17: will fix it, thanks:)

js/controller.js
shruthi17: yeah it has to be ,  if (duration !== 0 && duration - currentPlayhead <= timeToShow &&

will fix it.

js/controller.js
```diff
@@ -417,10 +417,9 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         // time to show is based on percentage of duration from the beginning
         timeToShow = (1 - parseInt(stringTimeToShow)/100) * duration;
       }
-
       this.state.upNextInfo.timeToShow = timeToShow;
 
-      if (duration - currentPlayhead <= timeToShow &&
+       if (duration!==0 && duration- currentPlayhead <= timeToShow &&
```
mistermendez: - remove extra space in front of `if`
- add spaces around operator, should be `duration !== 0`
- keep each condition on separate line

mistermendez:
👍

PR #610, opened 2016-07-21
[PBW-5633] [V4] Thumbnails sometimes jump to position that is not aligned with mouse pointing at their center

mityaha:
If mouse is moved slowly, Thumbnail object may go away and a new one will be created.  This object doesn't *know* thumbnail width, so it is zero during initial rendering, which results in wrong placing.  Solution is to save last thumbnail width and pass it as a parameter to next Thumbnail object.  Store this value in Thubmnail static method.


mityaha:
Looks good

PR #613, opened 2016-07-21
[PBW-5653] [V4] [Thumbnail Scrubbing] First thumbnail scrubbing images are appearing again if we scrub near the end of video

mityaha:
Fixed initial value of selectedPositon


mistermendez:
👍

PR #616, opened 2016-07-25
[PBW-5524] Vertical scroll bar on CC options is not getting display b…

brianfwong:
…y default


mistermendez:
Scrollbar should use browser default not white background.

I recommend resolving PBW-5524 w/o dev. @Jgkelle26 can you weigh in here.

screen shot 2016-08-01 at 4 04 46 pm

PR #617, opened 2016-07-26
[PBF-665] Adds awareness of live-button click to events.

brianfwong:
See matching pull requests that will be in mjolnir and html5-video-plugins.


chibzu:
Looks good.

PR #618, opened 2016-07-28
[PBW-5708] verify saved caption lang availability

mistermendez:


js/controller.js
```diff
@@ -1222,6 +1222,10 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
     },
 
     setClosedCaptionsLanguage: function(){
+      //if saved language not in available languages, set to first available language
+      if (this.state.closedCaptionOptions.language == null || !_.contains(this.state.closedCaptionOptions.availableLanguages.languages, this.state.closedCaptionOptions.language)) {
+        this.state.closedCaptionOptions.language = this.state.closedCaptionOptions.availableLanguages.languages[0];
```
mityaha: Is it always a sure thing that array this.state.closedCaptionOptions.availableLanguages.languages is not empty?  Don't you also want to check whether it is not empty?

mistermendez: No, the contains() function would just return false.

mistermendez:
@mityaha does this also fix PBW-5709 ?

mityaha:
I haven't built with this change, I'll try.

js/controller.js
aldigun: Shouldn't we default to the language set in skin.json first, then fallback to first language.

aldigun: Oh nvm, during controller creation, language selection is already extended to contain the skin.json setting.

mistermendez: yup done here - https://github.com/ooyala/html5-skin/pull/618/files#diff-ceb739c986db75454cd0d5c8f22f2928L228

aldigun: LGTM.

aldigun:
This defintely fixes PBW-5708 and PBW-5712

PR #623, opened 2016-08-02
[PBW-5734] Prevent Discovery Up-next from showing at video start afte…

brianfwong:
Prevent Discovery Up-next from showing at video start after pre-roll.


mistermendez:
👍

PR #625, opened 2016-08-03
Remove a redundant element

kuu:


mistermendez:
This`oo-state-screen-selectable` is nested in the `oo-interactive-container`.

It's used to prevent the `oo-interactive-container` from blocking user selection.

PR #627, opened 2016-08-04
[PBW-5070] Close button now picks up iconstyle json config

brianfwong:


mistermendez:
The highlight config is for the control bar buttons only not the close button.

I recommend resolving PBW-5070 w/o dev.

We should also consider moving highlight style into the CSS. There is way too much code to achieve this using JSON and inline style instead of CSS :hover.

PR #630, opened 2016-08-11
[PBW-5518] sync skin with default logo and update unit test

mistermendez:
⚠️  must merge with https://github.com/ooyala/skin-config/pull/152

html5-skin is design to [exclude logo](https://github.com/ooyala/html5-skin/blob/stable/js/components/controlBar.js#L400) if there is no imageResource url.

This approach allows customers to continue to use custom logo if needed.


chibzu:
Looks good.

PR #634, opened 2016-08-15
Consolidate source files into 'src' dir

mistermendez:
Moved the following into `src` dir:
    * js/
    * scss/
    * assets/
    * iframe.html
    * amp_iframe.html


mistermendez:
Let's merge this after August release. I don't want to complicate cherry picking.

mistermendez:
It's best to close this to avoid major conflicts.

PR #648, opened 2016-08-31
[PBW-5804] Fix for making volume level disappear after some seconds i…

nagavenivn:
…n Android

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -1362,7 +1362,20 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
 
     showVolumeSliderBar: function() {
       this.state.volumeState.volumeSliderVisible = true;
+      if (Utils.isAndroid()) {
+        this.startHideVolumeSliderTimer();
+      }
       this.renderSkin();
+    },
+
+    startHideVolumeSliderTimer: function() {
+        this.cancelTimer();
+        var timer = setTimeout(function() {
+          if(this.state.volumeState.volumeSliderVisible === true){
+            this.hideVolumeSliderBar();
+          }
+        }.bind(this), 3000);
+        this.state.timer = timer;
     },
 
     startHideControlBarTimer: function() {


js/controller.js
```diff
@@ -1362,9 +1362,22 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
       this.state.volumeState.volumeSliderVisible = true;
+      if (Utils.isAndroid()) {
+        this.startHideVolumeSliderTimer();
+      }
       this.renderSkin();
     },
 
+    startHideVolumeSliderTimer: function() {
+        this.cancelTimer();
+        var timer = setTimeout(function() {
```
pilievOoyala: Is there any way to cancel this timeout?  what if we need to show it a second into this timeout?

js/controller.js
```diff
@@ -1362,9 +1362,22 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+          if(this.state.volumeState.volumeSliderVisible === true){
+            this.hideVolumeSliderBar();
+          }
+        }.bind(this), 3000);
+        this.state.timer = timer;
```
pilievOoyala: can you check if a timer already exists, if it does then cancel it.  Then do the assignment.

SainiBasu: There was no timer to hide volume slider, hence we added this timer. This timer starts when the volume slider becomes visible and as soon as it times out (i.e it gets cancelled) it will hide the volume slider. Control bar also uses the same logic.

js/controller.js
SainiBasu: This timer is different from the one discovery up next uses. The user might want to cancel the discovery timer explicitly as he may not want to see the next video auto-playing. But here the timer need not be cancelled explicitly as it starts once the slider becomes visible and it should auto hide as soon as the timer times out. The user intervention is not required here.

js/controller.js
brianfwong: @pilievOoyala If you are asking about a null check on the call to this.cancelTimer(), there is one inside of that function:

`cancelTimer: function() {
      if (this.state.timer !== null){
        clearTimeout(this.state.timer);
        this.state.timer = null;
      }
    },`

js/controller.js
brianfwong: @pilievOoyala The timer here will be cancelled anytime the user hits the volume button, which triggers show, cancels 'this' timer, and then starts the hide timer.
- Is that the timeout you are referring to?

js/controller.js
pilievOoyala: That's what I was looking for.  Thanks!

js/controller.js
pilievOoyala: I was worried about unexpected state transitions resulting in timers being left around and not canceled.  But now I see that it's handled.  So you can disregard my concern.

mistermendez:
👍

PR #649, opened 2016-09-01
Pbw 5841

nagavenivn:
Setting the timer to flase when discovery icon on the end screen is clicked and 
making it true for other scenarios on end screen.


mistermendez:
👍

PR #650, opened 2016-09-06
[PBW-5876] - Player will now display muted icon when initialVolume is set to 0

jpfranco:
Eventually, I think it would be better to mimic the behavior of the HTML5 video API in which `muted` and `volume` properties are handled separately. This would allow the user to set an initial volume even if the player starts out muted.

For the scope of this ticket, I think the changes I made should be sufficient though.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -376,8 +376,14 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       this.renderSkin({"contentTree": contentTree});
     },
 
-    onVolumeChanged: function (event, newVolume){
+    onVolumeChanged: function (event, newVolume) {
+      if (newVolume == 0) {
+        this.state.volumeState.muted = true;
+      } else {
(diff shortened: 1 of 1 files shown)


mistermendez:
should this be merged into `master` instead of `stable`?

js/controller.js
```diff
@@ -376,8 +376,14 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       this.renderSkin({"contentTree": contentTree});
     },
 
-    onVolumeChanged: function (event, newVolume){
+    onVolumeChanged: function (event, newVolume) {
+      if (newVolume == 0) {
```
pilievOoyala: This should be <= 0.  Just in case we get a negative number.

jpfranco:
@mistermendez right, sorry. I'm still not familiar with the project's branches. I'll create another pull request with the right branch in a minute.

PR #651, opened 2016-09-06
[PBW-5876] - Player will now display muted icon when initialVolume is set to 0

jpfranco:
Redid the pull request because I had the wrong branch. Also added Peter's suggested changes.

Original comment:

Eventually, I think it would be better to mimic the behavior of the HTML5 video API in which `muted `and `volume` properties are handled separately. This would allow the user to set an initial volume even if the player starts out muted.

For the scope of this ticket, I think the changes I made should be sufficient though.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -393,8 +393,15 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       this.renderSkin({"contentTree": this.state.contentTree});
     },
 
-    onVolumeChanged: function (event, newVolume){
-      this.state.volumeState.volume = newVolume;
+    onVolumeChanged: function (event, newVolume) {
+      if (newVolume <= 0) {
+        this.state.volumeState.muted = true;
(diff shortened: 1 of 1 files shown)


mityaha:
Looks good.  I agree with your remark that it is a design flaw that there is no explicit mute API.  It would be great if you added some explanation why moving code from setVolume to onVolumeChanged fixed the problem

jpfranco:
@mityaha sure. So there is a `VOLUME_CHANGED` event fired initially when the value of `initialVolume` is set.  On the previous version of the code, the volume change handler of the skin didn't have the logic to update the `muted` state of the player, that's why the volume was set to zero, but the mute icon wasn't shown.

I moved the rendering logic on `setVolume` to `onVolumeChanged` in order to keep it in a single place. The skin should react to volume changes triggered by both the skin's `setVolume` and other external events, so this seems like a better place for this logic to be.

mityaha:
Looks good, thanks.

PR #653, opened 2016-09-08
[PBW-5864] Restoring main video's closed caption settings after ad plays

jpfranco:
This is the flow that was causing the issue:

1. On Firefox video preloading is enabled, so bitmovin plugin initializes on page load and fires `CAPTIONS_FOUND_ON_PLAYING`.
2. The `video controller` handles this event and then fires `CLOSED_CAPTIONS_INFO_AVAILABLE`.
3. The skin handles this event and uses its info to populate the CC menu.
4. When the video is played, the preroll video is loaded and bitmovin again fires `CAPTIONS_FOUND_ON_PLAYING`, this time with empty captions.
5. The `video controller` again fires `CLOSED_CAPTIONS_INFO_AVAILABLE`, which causes the skin to clear the captions that it had for the main video.
6. When the main video regains focus, the skin is stuck with the captions from the ad (which are empty in this case). 

It's also possible that this issue can occur after a mid-roll ad if the ad itself has subtitles. It seems that technically the bitmovin plugin could fire a `CAPTIONS_FOUND_ON_PLAYING` during mid-roll initialization and cause the skin to clear its closed captions data. I wasn't able to reproduce this locally though.

The solution I implemented was to have the skin keep track of CC info for all of the different video id's that have been active. When the `VC_VIDEO_ELEMENT_IN_FOCUS` event is received by the skin, it now makes sure that it has loaded the captions that match the currently active video.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -73,6 +73,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         "skipAdButtonEnabled": false
       },
 
+      "closedCaptionsInfoCache": {},
       "closedCaptionOptions": {
         "enabled": null,
         "language": null,
@@ -781,6 +782,9 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
 
     onVideoElementFocus: function(event, source) {
       this.focusedElement = source;
+      // Make sure that the skin uses the captions that correspond
+      // to the newly focused video element
+      this.setClosedCaptionsInfo(source);
       if (source == OO.VIDEO.MAIN) {
         this.state.pluginsElement.removeClass("oo-showing");
         this.state.pluginsClickElement.removeClass("oo-showing");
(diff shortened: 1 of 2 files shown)


mityaha:
Looks very reasonable! Jeoffrey, can you please review it as the owner of the skin closed captioning code?

mistermendez:
👍

js/controller.js
```diff
@@ -833,11 +839,11 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
 
-    onClosedCaptionsInfoAvailable: function(event, languages) {
-      this.state.closedCaptionOptions.availableLanguages = languages;
-      if (this.state.closedCaptionOptions.enabled){
-        this.setClosedCaptionsLanguage();
-      }
+    onClosedCaptionsInfoAvailable: function(event, info) {
+      // Store info in cache in order to be able to restore it
+      // if this video element looses and then regains focus (like when an ad plays)
+      this.state.closedCaptionsInfoCache[info.videoId] = info;
```
pilievOoyala: you need to check if info is valid and not null/undefined

PR #656, opened 2016-09-15
[PBW-5843] - Allowing clicks to pass through pause icon on ad screen

jpfranco:
**Edit: I'm going to include another fix, please don't merge yet.**

So the fix is basically just allowing clicks to pass through the pause icon which blocks the video element even when it fades out.  This fix will not work on IE10 or lower, but considering [the list of supported browsers](http://support.ooyala.com/developers/documentation/concepts/pbv4_environments.html) I think this should be ok.


mistermendez:
nice 👍

PR #657, opened 2016-09-16
[PBW-5306] fix playhead position for pre and mid roll ads

mistermendez:
1.) Afer ad plays or skip button is clicked, ensure playhead is updated with main video playhead, duration, buffer values
2.) Ensure buffer indicator does not exceed 100% of scrubber bar

### DIFF
--- a/js/components/scrubberBar.js
+++ b/js/components/scrubberBar.js
@@ -196,7 +196,7 @@ var ScrubberBar = React.createClass({
       backgroundColor: this.props.skinConfig.controlBar.scrubberBar.backgroundColor
     };
     var bufferedIndicatorStyle = {
-      width: (parseFloat(this.props.buffered) / parseFloat(this.props.duration)) * 100 + "%",
+      width: Math.min((parseFloat(this.props.buffered) / parseFloat(this.props.duration)) * 100, 100) + "%",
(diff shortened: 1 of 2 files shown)


js/components/scrubberBar.js
```diff
@@ -196,7 +196,7 @@ var ScrubberBar = React.createClass({
       backgroundColor: this.props.skinConfig.controlBar.scrubberBar.backgroundColor
     };
     var bufferedIndicatorStyle = {
-      width: (parseFloat(this.props.buffered) / parseFloat(this.props.duration)) * 100 + "%",
+      width: Math.min((parseFloat(this.props.buffered) / parseFloat(this.props.duration)) * 100, 100) + "%",
```
pilievOoyala: if one of these two parseFloat returns a bad value (NaN) will it continue to work?

pilievOoyala: You may want to do the parseFloat before, then if the numerator is NaN, make it 0 and if the denominator is NaN make it 1.

mistermendez: Yup, it'll work, I need to use the same logic as the playhead indicator - https://github.com/ooyala/html5-skin/blob/stable/js/components/scrubberBar.js#L203

mityaha:
Looks good!

PR #658, opened 2016-09-20
[PBW-5843] Fixed adPanel's z-index when Pulse plugin is active

jpfranco:
These changes should fix two issues:
- Ads can't be clicked and therefore can't be paused/unpaused on iPad.
- Clicking on an ad on desktop pauses the ad but doesn't trigger a click-through.

Both of these issues only occur when using the **Pulse** ad plugin.

### DIFF
--- a/js/components/adPanel.js
+++ b/js/components/adPanel.js
@@ -133,6 +133,7 @@ var AdPanel = React.createClass({
     return (
       <div className="oo-ad-screen-panel">
         {spinner}
+        <div className="oo-ad-screen-panel-click-layer"></div>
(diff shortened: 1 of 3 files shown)


js/components/adPanel.js
```diff
@@ -133,6 +133,7 @@ var AdPanel = React.createClass({
     return (
       <div className="oo-ad-screen-panel">
         {spinner}
+        <div className="oo-ad-screen-panel-click-layer"></div>
```
jpfranco: I used a separate element here because, if I changed the z-index of `oo-ad-screen-panel`, I would invalidate the z-index of `oo-ad-top-bar`. With this approach I was able to fix the issue without changing the z-index of `oo-ad-top-bar`.

mistermendez:
👍

PR #660, opened 2016-09-23
[PBI-1970]

dchauOoyala:
- Update OO.ERROR.VTC.UNSUPPORTED_ENCODING error description to use OO.ERROR.PLAYBACK.STREAM error's description
- Remove OO.ERROR.PLAYBACK.STREAM error since the message was never published due to an unsupported encoding error, but more appropriately describes the above VTC error message.


mistermendez:
👍

dchauOoyala:
Shoot, I meant to close this. We decided to keep the streams error message, but follow through with the change for the unsupported encoding error message.

PR #662, opened 2016-09-26
[PBW-5959] add hasOwnProperty check to prevent prototype.js bug

mistermendez:


js/views/adScreen.js
```diff
@@ -135,8 +135,10 @@ var AdScreen = React.createClass({
     };
 
     var playbackControlItems = [];
-    for(var item in playbackControlItemTemplates) {
-      playbackControlItems.push(playbackControlItemTemplates[item]);
+    for (var item in playbackControlItemTemplates) {
```
mityaha: Wouldn't for (var item = 0; item < playbackControlItemTemplates.length; item++) be more effective?

mityaha: Or is playbackControlItemTemplates an object, not array?

mistermendez: right it's an object

mityaha:
Looks good

PR #665, opened 2016-10-12
[PBW-6068] removing blur when upNext or Discovery element is clicked

ksunika:
fixes blurred ads after clicking on upNext or Discovery element

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -1205,6 +1205,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
     },
 
     sendDiscoveryClickEvent: function(selectedContentData, isAutoUpNext) {
+      this.state.pluginsElement.removeClass("oo-overlay-blur");
       this.state.upNextInfo.showing = false;
       if (isAutoUpNext){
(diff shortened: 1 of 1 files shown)


chibzu:
Do we know why the onWillPlaySingleAd method doesn't remove the blur here?

pilievOoyala:
Is there an onExit method that would be a better place to put this. In case there is a way to change the video without clicking?

ksunika:
@chibzu, onWillPlaySingleAd removes 'oo-blur' but not 'oo-overlay-blur'. Looks like the latter was added to blur overlay ads that video might have.

ksunika:
@pilievOoyala, we don't have a single onExit method here. There are two ways to leave that screen through UI, one is by clicking close and another is by selecting a video. They are handled separately. 

If the video is changed after upNext countdown, it goes through the same  sendDiscoveryClickEvent.

If the video is changed through setEmbedCode, the blurring problem does not occur since the flow is different.

chibzu:
Got it, looks fine by me.

mistermendez:
👍

PR #667, opened 2016-10-19
[PBW-6113] - Fix for CC's being truncated on iPhone

jpfranco:
Looks like the CC's where being squeezed between the `top` and `bottom` properties on smaller screens.


mistermendez:
👍🏽

PR #671, opened 2016-10-25
Bundle skin.json with html5-skin

mistermendez:
1.) Import and bundle skin.json AND all JSON lang files (en, es, zh)
2.) Replace jQuery ajax requests ($.getJSON) with imported skin.json and JSON lang files
3.) update unit test to work with updates
4.) remove config/skin.json param in html files
5.) our html files now require charset utf-8 because language strings are loaded from external js file instead of ajax, this is standard for all multi-lingal sites
6.) if config (skin.json) param or lang jsons (en, es, zh) exist load via ajax for backwards compatibility, else use defaults bundled in app
7.) must be merged with - https://github.com/ooyala/skin-config/pull/154

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -7,7 +7,8 @@ var React = require('react'),
     CONSTANTS = require('./constants/constants'),
     AccessibilityControls = require('./components/accessibilityControls'),
     Fullscreen = require('screenfull'),
-    Skin = require('./skin');
+    Skin = require('./skin'),
+    SkinJSON = require('../config/skin');
 
 OO.plugin("Html5Skin", function (OO, _, $, W) {
   //Check if the player is at least v4. If not, the skin cannot load.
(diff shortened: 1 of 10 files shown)


mistermendez:
NOTE: The bundle tool we use, Browserify cannot dynamically bundle files at runtime.

The lang JSON files in config (en, es, zh) are still loaded dynamically at runtime w/ ajax.

We'll need another bundle tool like Webpack to dynamically bundle. Or do static bundle for just those 3 lang files - would not work well for customers with other languages.

mistermendez:
closing to investigate dynamic bundle options

mistermendez:
This is now [PLAYER-16](https://jira.corp.ooyala.com/browse/PLAYER-16) in JIRA

tests/controller-test.js
```diff
@@ -217,11 +226,16 @@ OO = {
...
     Html5Skin.externalPluginSubscription.call(controllerMock);
 
-    // test player state
-    Html5Skin.onPlayerCreated.call(controllerMock, 'customerUi', 1, {skin:{config:{}}}, {});
+    //test player state
+    var tempSkin = controllerMock.skin;
+    Html5Skin.onPlayerCreated.call(controllerMock, 'customerUi', elementId, {skin:{config:{}}}, {});
+    Html5Skin.loadConfigData.call(controllerMock, {skin:{config:{}}}, {});
+    Html5Skin.createPluginElements.call(controllerMock);
+    controllerMock.skin = tempSkin; //reset skin, onPlayerCreated updates skin
```
jpfranco: I think that tempSkin would point to the same object as controllerMock.skin, so we'd probably need to clone it in order for this to work as a reset.

tests/controller-test.js
```diff
@@ -217,11 +226,16 @@ OO = {
...
 
+    var tempMainVideoElement = controllerMock.state.mainVideoElement;
     Html5Skin.onVcVideoElementCreated.call(controllerMock, 'customerUi', {videoId: OO.VIDEO.MAIN});
-    controllerMock.state.mainVideoElement = {addClass: function(a) {}, removeClass: function(a) {}, get: function(a) { return { webkitSupportsFullscreen: true, webkitEnterFullscreen: function() {}, webkitExitFullscreen: function() {}, addEventListener: function(a,b) {}}}}
+    controllerMock.state.mainVideoElement = tempMainVideoElement;
```
jpfranco: Same comment as above, I think this might need a clone.

js/controller.js
```diff
@@ -831,6 +779,85 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+      //override state settings with defaults from skin config and possible local storage settings
+      $.extend(true, this.state.closedCaptionOptions, SkinJSON.closedCaptionOptions, settings.closedCaptionOptions);
+      this.state.config = SkinJSON;
+
+      //load config language json if exist
+      if (SkinJSON.localization.availableLanguageFile) {
+        SkinJSON.localization.availableLanguageFile.forEach(function(languageObj){
+          if (languageObj.languageFile) {
+            $.getJSON(languageObj.languageFile, function(data) {
+              Localization.languageFiles[languageObj.language] = data;
```
jpfranco: I'm not entirely sure but it looks like, by the time the getJSON callback is called, the value of languageObj will no longer be the same as it was when getJSON itself was called.

mistermendez: languageObj isn't modified it's read-only. Only the Localization object is updated if lang file is specified in config file.

mistermendez:
@jpfranco81 made updates to unit test

js/controller.js
jpfranco: Oh nevermind, the problem I'm thinking about would happen if it were a regular for loop. Should be good 👍

jpfranco:
lg 👍

PR #674, opened 2016-10-27
[PBW-5932] Implemented text customization for discovery screen

jpfranco:
I also fixed an issue that prevented the end screen's countdown timer from being disabled. I was playing with default props but it turns out that they are applied with a shallow merge, so if the main property is not undefined (like this.props.skinConfig) the nested property defaults don't get applied. I added a helper function to prevent these changes from crashing if the user forgets to define a property on the skin.

### DIFF
--- a/js/components/utils.js
+++ b/js/components/utils.js
@@ -264,6 +264,33 @@ var Utils = {
   },
 
   /**
+   * Safely gets the value of an object's nested property.
+   *
+   * @function getPropertyValue
+   * @param {Object} object - The object we want to extract the property form
+   * @param {String} propertyPath - A path that points to a nested property in the object with a form like 'prop.nestedProp1.nestedProp2'
+   * @param {Object} defaltValue - (Optional) A default value to return when the property is undefined
+   * @return {Object} - The value of the nested property, the default value if nested property was undefined
+   */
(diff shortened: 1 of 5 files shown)


js/controller.js
```diff
@@ -1231,9 +1231,6 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         "custom" : { "source" : screen}
       };
       this.mb.publish(OO.EVENTS.DISCOVERY_API.SEND_DISPLAY_EVENT, eventData);
-      if (screen == CONSTANTS.SCREEN.END_SCREEN) {
```
jpfranco: I need a second opinion about this. I couldn't figure out why this is overridden instead of using the value defined on the skin.json, but this prevented the end screen's countdown timer from being disabled.

js/components/utils.js
```diff
@@ -264,6 +264,33 @@ var Utils = {
   },
 
   /**
+   * Safely gets the value of an object's nested property.
```
mistermendez: very useful, reduces need for null checks

js/controller.js
mistermendez: this looks like a needed improvement 👍

mistermendez:
👍

PR #675, opened 2016-10-28
[PBW-6065] - Automatically setting docUrl to frame's parent url when not explicitly set

jpfranco:
This will automatically get the parent frame's url and use it for analytics, without the customer having to set it via `docUrl`. Passing `docUrl` through the query string will override this.

### DIFF
--- a/iframe.html
+++ b/iframe.html
@@ -27,6 +27,8 @@
     var docUrl = undefined;
     if (!!queryParams.match(/docUrl=([^&]*)/)) {
       playerParam["docUrl"] = decodeURIComponent(queryParams.match(/docUrl=([^&]*)/)[1]);
+    } else {
+      playerParam["docUrl"] = document.referrer;
     }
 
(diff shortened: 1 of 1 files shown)


iframe.html
```diff
@@ -27,6 +27,8 @@
     var docUrl = undefined;
     if (!!queryParams.match(/docUrl=([^&]*)/)) {
       playerParam["docUrl"] = decodeURIComponent(queryParams.match(/docUrl=([^&]*)/)[1]);
+    } else {
+      playerParam["docUrl"] = document.referrer;
```
pilievOoyala: I've seen some customer complain about referrer being null. Can you put in an extra check? If it is output something to the console so we can see it.

jpfranco: Ok sure, I'll add that.

jpfranco: @pilievOoyala done.

pilievOoyala:
Looks good!

PR #678, opened 2016-11-01
[PBW-5939] watermark related changes 

ksunika:
should work with https://github.com/ooyala/skin-config/pull/156

to test: https://goo.gl/svljJF

### DIFF
--- a/js/components/watermark.js
+++ b/js/components/watermark.js
@@ -1,0 +1,148 @@
+/**
+ * Watermark component
+ *
+ * @module Watermark
+ */
+var React = require('react'),
+    CONSTANTS = require('../constants/constants'),
+    ClassNames = require('classnames'),
+    Utils = require('./utils');
+
+var Watermark = React.createClass({
+
+  getInitialState: function() {
+    this.isMobile = this.props.controller.state.isMobile;
+    return null;
+  },
+
+  handleMouseUp: function(event) {
+    // handle watermark clicked on desktop
+    if (!this.isMobile) {
+      event.stopPropagation(); // W3C
+      event.cancelBubble = true; // IE
+
+      if (this.props.playerState == CONSTANTS.STATE.PLAYING){
+        this.props.controller.togglePlayPause();
+      }
+      this.props.controller.state.accessibilityControlsEnabled = true;
+      this.handleWatermarkClick();
+    }
+    // for mobile, touch is handled in handleTouchEnd
+  },
+
+  handleTouchEnd: function(event) {
+    event.preventDefault();//to prevent mobile from propagating click to discovery shown on pause
+    if (this.props.playerState == CONSTANTS.STATE.PLAYING){
+        this.props.controller.togglePlayPause();
+    }
+    this.handleWatermarkClick();
+  },
+
+  handleWatermarkClick: function() {
+    window.open(this.props.skinConfig.general.watermark.clickUrl, this.props.skinConfig.general.watermark.target);
+  },
+
+  render: function() {
+    var watermarkUrl = this.props.skinConfig.general.watermark.imageResource.url;
+    var clickUrl = this.props.skinConfig.general.watermark.clickUrl;
+
+    if (!watermarkUrl) {
+      return null;
+    }
+    else {
+      var watermarkStyle = {};
+      watermarkStyle.opacity = this.props.skinConfig.general.watermark.transparency;
+      if (this.props.skinConfig.general.watermark.scalingOption == "height") {
+        watermarkStyle.height = this.props.skinConfig.general.watermark.scalingPercentage + '%';
+      }
+      else if (this.props.skinConfig.general.watermark.scalingOption == "width") {
+        watermarkStyle.width = this.props.skinConfig.general.watermark.scalingPercentage + '%';
+      }
+      else if (this.props.skinConfig.general.watermark.scalingOption == "default") {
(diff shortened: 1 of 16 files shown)


ksunika:
@mistermendez I addressed the comments

js/components/watermark.js
```diff
@@ -0,0 +1,159 @@
...
+    ClassNames = require('classnames');
+
+var Watermark = React.createClass({
+
+  getInitialState: function() {
+    this.isMobile = this.props.controller.state.isMobile;
+    return null;
+  },
+
+  handleMouseUp: function(event) {
```
mistermendez: do we still need this?

js/components/watermark.js
```diff
@@ -0,0 +1,159 @@
...
+      if (this.props.playerState == CONSTANTS.STATE.PLAYING){
+        this.props.controller.togglePlayPause();
+      }
+      this.props.controller.state.accessibilityControlsEnabled = true;
+      this.handleWatermarkClick();
+    }
+    // for mobile, touch is handled in handleTouchEnd
+  },
+
+  handleTouchEnd: function(event) {
```
mistermendez: do we still need handleTouchEnd?

js/components/watermark.js
```diff
@@ -0,0 +1,159 @@
...
+    'oo-blur': this.props.nonClickable,
+    'oo-watermark': true
+  });
+
+    var watermarkImage = <img className={watermarkImageClass} style={watermarkImageStyle} src={watermarkUrl} ref='watermarkImage'/>;
+    if (this.props.nonClickable)
+      return (<div className={watermarkClass} ref='watermark' style={watermarkStyle}>{watermarkImage}</div>);
+    else
+      return (<a className={watermarkClass} ref='watermark' style={watermarkStyle} href={this.props.skinConfig.general.watermark.clickUrl}
+            target={this.props.skinConfig.general.watermark.target} onMouseUp={this.handleMouseUp} onTouchEnd={this.handleTouchEnd}>{watermarkImage}</a>);
```
mistermendez: we can remove `onMouseUp` and `onTouchEnd ` here

js/components/watermark.js
```diff
@@ -0,0 +1,159 @@
...
+ *
+ * @module Watermark
+ */
+var React = require('react'),
+    CONSTANTS = require('../constants/constants'),
+    ClassNames = require('classnames');
+
+var Watermark = React.createClass({
+
+  getInitialState: function() {
```
mistermendez: we no longer need `getInitialState`

js/components/watermark.js
```diff
@@ -0,0 +1,159 @@
...
+        scalingOption: React.PropTypes.string,
+        scalingPercentage: React.PropTypes.number,
+        nonClickable: React.PropTypes.bool
+      })
+    })
+  }),
+  controller: React.PropTypes.shape({
+    togglePlayPause: React.PropTypes.func,
+    state: React.PropTypes.shape({
+      isMobile: React.PropTypes.bool
```
mistermendez: we can remove isMobile

js/components/watermark.js
```diff
@@ -0,0 +1,159 @@
...
+
+  handleTouchEnd: function(event) {
+    event.preventDefault();//to prevent mobile from propagating click to discovery shown on pause
+    if (this.props.playerState == CONSTANTS.STATE.PLAYING){
+        this.props.controller.togglePlayPause();
+    }
+    this.handleWatermarkClick();
+  },
+
+  handleWatermarkClick: function() {
```
ksunika: you're right, not needed any more

js/components/watermark.js
ksunika: @mistermendez, we use mouseUp and touchEnd to pause the player if playing

js/views/contentScreen.js
```diff
@@ -45,12 +46,12 @@ var ContentScreen = React.createClass({
 
     return (
       <div className={this.props.screenClassName}>
+        <Watermark {...this.props} controlBarVisible={false} nonClickable={true}/>
```
mistermendez: do we need the `nonClickable ` param? should `general.watermark.clickUrl` determine if the watermark is clickable or not?

js/views/endScreen.js
```diff
@@ -37,6 +38,8 @@ var EndScreen = React.createClass({
 
         <a className="oo-state-screen-selectable" onClick={this.handleClick}></a>
```
mistermendez: we can also remove all `oo-state-screen-selectable` that are children of `oo-interactive-container`.

now that users can click through `oo-interactive-container` we don't need duplicate `oo-state-screen-selectable`

mistermendez:
we can also remove all `oo-state-screen-selectable` elements that are children of `oo-interactive-container`.

now that users can click through `oo-interactive-container` we don't need duplicate `oo-state-screen-selectable`

js/views/contentScreen.js
ksunika: @mistermendez, the watermark is not clickable from CC, share, bitrate selection screens, but we still show it.

mistermendez: why is it not clickable on those screens? how does the user determine click-ability?

an empty `general.watermark.clickUrl` param should also determine click-ability.

ksunika: Agree on the empty clickUrl, I lost that check along the way.

About the content screen: since the watermark can be placed in different parts of the screen, if we make it fully functional on CC or share screens, it might interfere with other elements (both visually and functionally). Currently it looks like it is behind the content screen, blurred with the video.

mistermendez:
👍

PR #681, opened 2016-11-11
[PLAYER-15] adding title and description to end screen

ksunika:
should work with https://github.com/ooyala/skin-config/pull/159

### DIFF
--- a/js/views/endScreen.js
+++ b/js/views/endScreen.js
@@ -2,16 +2,35 @@
   END SCREEN
 *********************************************************************/
 var React = require('react'),
+    ReactDOM = require('react-dom'),
     ClassNames = require('classnames'),
     ControlBar = require('../components/controlBar'),
     Watermark = require('../components/watermark'),
-    Icon = require('../components/icon');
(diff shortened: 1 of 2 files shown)


js/views/endScreen.js
```diff
@@ -27,11 +46,40 @@ var EndScreen = React.createClass({
       opacity: this.props.skinConfig.endScreen.replayIconStyle.opacity
     };
 
+    var titleStyle = {
+      color: this.props.skinConfig.startScreen.titleFont.color
```
ksunika: We are reusing the color in pause screen, so I thought it makes sense if all three screens have the same color.

js/views/endScreen.js
```diff
@@ -27,11 +46,40 @@ var EndScreen = React.createClass({
...
+    };
+    var descriptionStyle = {
+      color: this.props.skinConfig.startScreen.descriptionFont.color
```
ksunika: We are reusing the color in pause screen, so I thought it makes sense if all three screens have the same color.

mistermendez:
👍

PR #682, opened 2016-11-14
[PLAYER-168] using showEmbedTab to show or hide the embed tab

ksunika:
should work with https://github.com/ooyala/skin-config/pull/160


mistermendez:
Should this param be `activeTab` to control which tab is active?

mistermendez:
- works with skin.json param updates
- able to control which tabs display and which is active tab
- unit tests pass

mistermendez:
👍

PR #683, opened 2016-11-14
[PLAYER-213] Publish SET_EMBED_CODE with playerParam on discovery click

dchauOoyala:
The lack of passing the playerParam caused an issue where the ad set code would not get propagated to subsequently clicked discovery videos.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -1244,7 +1244,8 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         this.renderSkin();
         this.mb.publish(OO.EVENTS.PAUSE);
         if (selectedContentData.clickedVideo.embed_code){
-          this.mb.publish(OO.EVENTS.SET_EMBED_CODE, selectedContentData.clickedVideo.embed_code);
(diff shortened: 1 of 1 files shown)


js/controller.js
```diff
@@ -1244,7 +1244,8 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         this.renderSkin();
         this.mb.publish(OO.EVENTS.PAUSE);
         if (selectedContentData.clickedVideo.embed_code){
-          this.mb.publish(OO.EVENTS.SET_EMBED_CODE, selectedContentData.clickedVideo.embed_code);
+          this.mb.publish(OO.EVENTS.SET_EMBED_CODE, selectedContentData.clickedVideo.embed_code,
+                          this.state.playerParam);
```
pilievOoyala: Does this cover the case where discovery auto plays the next video?

dchauOoyala: No it does not because the logic for discovery auto plays is covered in the onPlayed method

pilievOoyala:
After chatting with you, the other code does pass the playerParams. So this was just missed here.

PR #687, opened 2016-11-16
[PLAYER-235] removing outline in Firefox

ksunika:


pilievOoyala:
LGTM.

pilievOoyala:
Firefox is the only browser we see the outline in. The changes apply to all browsers but it really only effect firefox.

PR #688, opened 2016-11-17
[PLAYER-253] allowing clicks to pass through spinner

ksunika:
fixes watermark not clickable when spinner is shown.


mistermendez:
👍

PR #695, opened 2016-11-30
[PLAYER-303] take description from contentTree so it updates on every…

brianfwong:
… new start screen, such as when scrubbing through playlist entries.

### DIFF
--- a/js/views/startScreen.js
+++ b/js/views/startScreen.js
@@ -41,6 +41,14 @@ var StartScreen = React.createClass({
   },
 
   render: function() {
+    if (!this.props.contentTree) {
+      this.props.contentTree = StartScreen.defaultProps.contentTree;
+    } 
+
+    if (!this.props.skinConfig) {
+      this.props.contentTree = StartScreen.defaultProps.skinConfig;
+    } 
+
     //inline style for config/skin.json elements only
     var titleStyle = {
       color: this.props.skinConfig.startScreen.titleFont.color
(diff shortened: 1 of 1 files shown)


js/views/startScreen.js
```diff
@@ -89,7 +89,7 @@ var StartScreen = React.createClass({
 
     var titleMetadata = (<div className={titleClass} style={titleStyle}>{this.props.contentTree.title}</div>);
     var iconName = (this.props.controller.state.playerState == CONSTANTS.STATE.END ? "replay" : "play");
-    var descriptionMetadata = (<div className={descriptionClass} ref="description" style={descriptionStyle}>{this.state.descriptionText}</div>);
+    var descriptionMetadata = (<div className={descriptionClass} ref="description" style={descriptionStyle}>{this.props.contentTree.description}</div>);
```
pilievOoyala: can you add a check to the beginning of this function to make sure content tree exists.  If not sets it to default empty object.  Ideally do it for skinconfig too.

nagavenivn: @pilievOoyala  Added the check as per your review comments.

mistermendez: this.props.contentTree.description has a [defaultProp](https://github.com/ooyala/html5-skin/blob/stable/js/views/startScreen.js#L169). 

@pilievOoyala with React.js defaultProps add stability to components and prevents us from having to add a bunch of null checks everywhere.

Also explained here - https://github.com/ooyala/html5-skin/pull/568#r67448769

mistermendez: @brianfwong we also need to leave this has a state variable. This update would break the description being truncated to 2 lines when the width of the player is resized. Use `componentWillReceiveProps()` to update `this.state.descriptionText` when `this.props.contentTree.description` is updated.

mistermendez: Try something like 
```
    if (nextProps.contentTree.description != this.props.contentTree.description) {
      this.handleResize();
    }
  },
```

mistermendez:
do we also need to add this to Pause and End screen?

nagavenivn:
This fix breaks the description being truncated to 2 lines when the width of the player is resized. Will be analyzing this again. 

Closing this pull request

PR #696, opened 2016-12-02
v3 to v4 deep merge updates

mistermendez:
* filter out disabled buttons (from v3 metadata)
* gracefully merge array of object and maintain sort order - used for buttons array in config JSON


ksunika:
👍 +1:

PR #700, opened 2016-12-07
reliably set this.state.mainVideoElement across all video plugins

mistermendez:
* created `findMainVideoElement()` function to reliably find and set `this.state.mainVideoElement`
* controller has been updated to find mainVideoElement for `onVcVideoElementCreated` and `onPlaying` events
* `bit_wrapper.js` video plugin creates nested video/object element after video is played. `bit_wrapper.js` needs to be updated to create main video element during plugin initialization 
* `this.state.mainVideoMediaType` is also reliably set
* removed some dependency on jQuery

### DIFF
--- a/js/constants/constants.js
+++ b/js/constants/constants.js
@@ -117,7 +117,9 @@ module.exports = {
 
   MEDIA_TYPE: {
     HTML5: "html5",
-    FLASH: "flash"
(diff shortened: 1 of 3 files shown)


mityaha:
Looks good

PR #701, opened 2016-12-08
background image validation

mistermendez:
undefined live stream promo image on start screen was causing console error.

added image validation to all background images.

### DIFF
--- a/js/components/utils.js
+++ b/js/components/utils.js
@@ -467,23 +467,15 @@ var Utils = {
   },
 
   /**
-   * Checks if image loads and is valid
+   * Checks if string is valid
    *
-   * @function isValidImage
-   * @param {String} src - absolute path to image src
-   * @returns {Boolean} true if image is valid, false if not
+   * @function isValidString
+   * @param {String} src - string to be validated
+   * @returns {Boolean} true if string is valid, false if not
    */
-  isValidImage: function(src) {
-    if (src) {
-      var isValidImage = true;
-      var img = new Image();
-      img.src = src;
-
-      // check if error occurs while loading img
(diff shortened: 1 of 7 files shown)


js/components/utils.js
```diff
@@ -465,6 +466,29 @@ var Utils = {
...
 
+  /**
+   * Checks if image loads and is valid
+   *
+   * @function isValidImage
+   * @param {String} src - absolute path to image src
+   * @returns {Boolean} true if image is valid, false if not
+   */
+  isValidImage: function(src) {
+    if (src) {
```
dchauOoyala: Would it be better here to check that `src` is a string and also not empty? Anything of truthy value will pass here (i.e. 1, a non-empty object, etc)

mistermendez: good catch will update logic. also noticed return boolean isn't being updated by onerror function.

js/components/utils.js
```diff
@@ -465,6 +466,21 @@ var Utils = {
...
 
+  /**
+   * Checks if string is valid
+   *
+   * @function isValidString
+   * @param {String} src - string to be validated
+   * @returns {Boolean} true if string is valid, false if not
+   */
+  isValidString: function(src) {
+    if (src && (typeof src === 'string' || src instanceof String)) {
```
dchauOoyala: consider simplifying this to: `return (src && (typeof src === 'string' || src instanceof String))`

mistermendez:
actual image validation (done via network request) would need to be done asynchronously and overkill for this issue. simplified logic to src string validation.

dchauOoyala:
LGTM

PR #702, opened 2016-12-08
[PLAYER-303] Fix for description does not getting udpated on videos l…

nagavenivn:
…oaded from playlist

### DIFF
--- a/js/views/startScreen.js
+++ b/js/views/startScreen.js
@@ -25,6 +25,10 @@ var StartScreen = React.createClass({
     this.handleResize();
   },
 
+  componentWillReceiveProps: function(nextProps) {
+    this.handleResize();
+  },
+
   handleResize: function() {
     if (ReactDOM.findDOMNode(this.refs.description)){
       this.setState({
--- a/js/views/startScreen.js
+++ b/js/views/startScreen.js
@@ -25,14 +25,17 @@ var StartScreen = React.createClass({
     this.handleResize();
   },
 
-  componentWillReceiveProps: function() {
-    this.handleResize();
+  componentWillReceiveProps: function(nextProps) {
+    if (nextProps.contentTree.description != this.props.contentTree.description) {
+      this.handleResize(nextProps);
+    }
   },
 
-  handleResize: function() {
+  handleResize: function(nextProps) {
+    var description = nextProps ? nextProps.contentTree.description : this.props.contentTree.description;
     if (ReactDOM.findDOMNode(this.refs.description)){
       this.setState({
-        descriptionText: Utils.truncateTextToWidth(ReactDOM.findDOMNode(this.refs.description), this.props.contentTree.description)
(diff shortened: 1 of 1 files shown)


pilievOoyala:
@mistermendez Can you double check this?  Looks fine to me.

js/views/startScreen.js
```diff
@@ -25,6 +25,10 @@ var StartScreen = React.createClass({
     this.handleResize();
   },
 
+  componentWillReceiveProps: function() {
```
nagavenivn: @mistermendez I tried with this condition before creating pull request and it will not work. Reason is in handleResize(), this .setState.descriptionText is set based on this.props.contentTree.description and not based on nextProps.  So when ever there is a change in description it won't take affect.

nagavenivn: @mistermendez By doing below also can fix the problem. But we will be duplicating the code.

    if (nextProps.contentTree.description == this.props.contentTree.description) {
      if (ReactDOM.findDOMNode(this.refs.description)){
        this.setState({
          descriptionText: Utils.truncateTextToWidth(ReactDOM.findDOMNode(this.refs.description), nextProps.contentTree.description)
        });
      }
    }
  },

mistermendez: Gotcha. Let's add a param to `handleResize()` to update state with nextProps. Will this work?
```
    if (nextProps.contentTree.description != this.props.contentTree.description) {
      this.handleResize(nextProps.contentTree.description);
    }
  },

  handleResize: function(text) {
    var description = text ? text : this.props.contentTree.description;
    if (ReactDOM.findDOMNode(this.refs.description)){
      this.setState({
        descriptionText: Utils.truncateTextToWidth(ReactDOM.findDOMNode(this.refs.description), description)
      });
    }
  },
```

We should always check for prop changes because this method gets called a lot, even if props have not changed. See https://facebook.github.io/react/docs/react-component.html#componentwillreceiveprops

nagavenivn: @mistermendez: Passing param to  handleResize() also will not work.  When you change the player size, then nextProps.contentTree.description and this.props.contentTree.description will be same. In this scenario handleResize() will not be called and description will be shown in more than 2 lines
screen shot 2016-12-09 at 10 09 19 am

nagavenivn: @mistermendez handleResize() is also called from ResizeMixin() and this calls with this.handleResize(nextProps); So I modified accordingly in startscreen.js and pushed the code changes

    if (nextProps.contentTree.description != this.props.contentTree.description) {
      this.handleResize(nextProps);
    }
  },

  handleResize: function(nextProps) {
    var description = nextProps ? nextProps.contentTree.description : this.props.contentTree.description;
    if (ReactDOM.findDOMNode(this.refs.description)){
      this.setState({
        descriptionText: Utils.truncateTextToWidth(ReactDOM.findDOMNode(this.refs.description), description)
      });
    }
  },

mistermendez:
sweet 🍬

PR #703, opened 2016-12-09
onSkinMetaDataFetched and deep merge updates

mistermendez:
* we are using this [deep merge lib](https://github.com/KyleAMathews/deepmerge/blob/master/index.js#L64)
* the custom `arrayDeepMerge` function is passed as a callback to deepmerge function and only used if there is an array in the object that needs to be merged
* Also see - https://github.com/KyleAMathews/deepmerge#arraymerge
* controller updated to subscribe to `OO.EVENTS.SKIN_METADATA_FETCHED` event, returns v3 metadata, deep merges with `skin.json`
* `arrayDeepMerge` function updated to replace buttons by default, prepend after flexible space is also another option
* buttons disabled if `location: "none"`
* `swap` no longer used but still an optional param

### DIFF
--- a/js/components/utils.js
+++ b/js/components/utils.js
@@ -431,34 +431,38 @@ var Utils = {
    * @function arrayDeepMerge
    * @param {Array} target - An array that will receive new items if additional items are passed
    * @param {Array} source - An array containing additional items to merge into target
-   * @param {Object} optionsArgument - optional parameters passed, i.e. arrayMerge function
+   * @param {Object} optionsArgument - optional parameters passed, i.e. arrayMerge, swap, unionBy, clone
    * @returns {Array} new merged array with items from both target and source
    */
   arrayDeepMerge: function(target, source, optionsArgument) {
+    var targetArray = optionsArgument.swap ? source : target;
+    var sourceArray = optionsArgument.swap ? target : source;
     var self = this;
     var destination = [];
-    destination = source.slice();
-    target.forEach(function(targetItem, i) {
+    destination = sourceArray.slice();
+
+    targetArray.forEach(function(targetItem, i) {
       if (typeof destination[i] === 'undefined') {
         destination[i] = self._cloneIfNecessary(targetItem, optionsArgument);
       }
       else if (self._isMergeableObject(targetItem)) {
         // custom merge for buttons array, used to maintain source sort order
-        if (targetItem.name) {
-          source.forEach(function(sourceItem, j) {
+        if (targetItem[optionsArgument.unionBy]) {
+          sourceArray.forEach(function(sourceItem, j) {
             //gracefully merge buttons by name
-            if (targetItem.name === sourceItem.name) {
-              destination[j] = DeepMerge(targetItem, sourceItem, optionsArgument);
+            if (targetItem[optionsArgument.unionBy] === sourceItem[optionsArgument.unionBy]) {
+              var targetObject = optionsArgument.swap ? sourceItem : targetItem;
+              var sourceObject = optionsArgument.swap ? targetItem : sourceItem;
+              destination[j] = DeepMerge(targetObject, sourceObject, optionsArgument);
             }
           });
         }
         // default array merge
         else {
-          destination[i] = DeepMerge(targetItem, source[i], optionsArgument);
+          destination[i] = DeepMerge(targetItem, sourceArray[i], optionsArgument);
         }
-
       }
-      else if (source.indexOf(targetItem) === -1) {
+      else if (sourceArray.indexOf(targetItem) === -1) {
         destination.push(self._cloneIfNecessary(targetItem, optionsArgument));
       }
     });
--- a/js/components/utils.js
+++ b/js/components/utils.js
@@ -432,21 +432,25 @@ var Utils = {
    * @function arrayDeepMerge
    * @param {Array} target - An array that will receive new items if additional items are passed
    * @param {Array} source - An array containing additional items to merge into target
-   * @param {Object} optionsArgument - optional parameters passed, i.e. arrayMerge, swap, unionBy, clone, arrayFusion
+   * @param {Object} optionsArgument - parameters passed to parent DeepMerge function, i.e. -
+   *        arrayMerge - https://github.com/KyleAMathews/deepmerge#arraymerge
+   *        clone - https://github.com/KyleAMathews/deepmerge#clone
+   *        arrayUnionBy - key used to compare Objects being merged, i.e. button name
+   *        arrayFusion - method used to merge arrays ['replace', 'prepend']
+   *        arraySwap - swaps target/source
    * @returns {Array} new merged array with items from both target and source
    */
   arrayDeepMerge: function(target, source, optionsArgument) {
-    // replaces target with source, no merge
+    // returns source, no merge
     if (optionsArgument.arrayFusion === 'replace') {
       return source;
     }
 
-    var targetArray = optionsArgument.swap ? source : target;
-    var sourceArray = optionsArgument.swap ? target : source;
+    var targetArray = optionsArgument.arraySwap ? source : target;
+    var sourceArray = optionsArgument.arraySwap ? target : source;
     var self = this;
-    var uniqueSourceArray = sourceArray.slice();
-    var destination = [];
-    destination = targetArray.slice();
+    var uniqueSourceArray = sourceArray.slice(); //array used to keep track of objects that do not exist in target
+    var destination = targetArray.slice();
 
     sourceArray.forEach(function(sourceItem, i) {
       if (typeof destination[i] === 'undefined') {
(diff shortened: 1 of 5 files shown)


js/components/utils.js
```diff
@@ -431,34 +431,38 @@ var Utils = {
...
    * @returns {Array} new merged array with items from both target and source
    */
   arrayDeepMerge: function(target, source, optionsArgument) {
+    var targetArray = optionsArgument.swap ? source : target;
+    var sourceArray = optionsArgument.swap ? target : source;
     var self = this;
     var destination = [];
-    destination = source.slice();
-    target.forEach(function(targetItem, i) {
+    destination = sourceArray.slice();
```
pilievOoyala: do we need to validate that the sourceArray is actually an array, since anything can be passed in?

js/components/utils.js
```diff
@@ -431,34 +431,38 @@ var Utils = {
    * @function arrayDeepMerge
    * @param {Array} target - An array that will receive new items if additional items are passed
    * @param {Array} source - An array containing additional items to merge into target
-   * @param {Object} optionsArgument - optional parameters passed, i.e. arrayMerge function
+   * @param {Object} optionsArgument - optional parameters passed, i.e. arrayMerge, swap, unionBy, clone
    * @returns {Array} new merged array with items from both target and source
    */
   arrayDeepMerge: function(target, source, optionsArgument) {
+    var targetArray = optionsArgument.swap ? source : target;
```
mistermendez: I tried that and it didn't maintain button sort order. we also need to swap target/source in the nested deep merge on line 456

mistermendez: also this `arrayDeepMerge` function is only used to merge an array of objects. actual objects in the main deep merge still need to maintain target/source hierarchy. this function changes how the array list of objects are looped to maintain correct sort order while maintaining target/source hierarchy of original merge.

will retest while working on unit test

js/components/utils.js
```diff
@@ -431,34 +431,38 @@ var Utils = {
...
         // custom merge for buttons array, used to maintain source sort order
-        if (targetItem.name) {
-          source.forEach(function(sourceItem, j) {
+        if (targetItem[optionsArgument.unionBy]) {
+          sourceArray.forEach(function(sourceItem, j) {
             //gracefully merge buttons by name
-            if (targetItem.name === sourceItem.name) {
-              destination[j] = DeepMerge(targetItem, sourceItem, optionsArgument);
+            if (targetItem[optionsArgument.unionBy] === sourceItem[optionsArgument.unionBy]) {
+              var targetObject = optionsArgument.swap ? sourceItem : targetItem;
```
pilievOoyala: If you do the swap in the beginning of the function, doesn't this switch it back to the original?

js/components/utils.js
pilievOoyala: sounds good

js/components/utils.js
mistermendez: the [deep merge lib](https://github.com/KyleAMathews/deepmerge/blob/master/index.js#L64) we are using does that.

this custom `arrayDeepMerge` function is passed as a callback to deepmerge function and only used if there is an array in the object that needs to be merged.

Also see - https://github.com/KyleAMathews/deepmerge#arraymerge

js/components/utils.js
mistermendez: yea, to maintaining target/source hierarchy of original merge

mistermendez: For example - `DeepMerge(skinJSON, v3MetaData)` if these two object have the same key, `v3MetaData` will override `skinJSON` and maintain the sort order of `v3MetaData`. We need to maintain sort order of `skinJSON`. 

The main purpose of this function is to maintain sort of target array while still maintaining target/source hierarchy of original merge.

mistermendez:
Closing until unit tests are complete

mistermendez:
unit tests are complete, this is ready for review

js/components/utils.js
```diff
@@ -432,37 +432,78 @@ var Utils = {
    * @function arrayDeepMerge
    * @param {Array} target - An array that will receive new items if additional items are passed
    * @param {Array} source - An array containing additional items to merge into target
-   * @param {Object} optionsArgument - optional parameters passed, i.e. arrayMerge function
+   * @param {Object} optionsArgument - optional parameters passed, i.e. arrayMerge, swap, unionBy, clone, arrayFusion
    * @returns {Array} new merged array with items from both target and source
    */
   arrayDeepMerge: function(target, source, optionsArgument) {
+    // replaces target with source, no merge
+    if (optionsArgument.arrayFusion === 'replace') {
```
pilievOoyala: since optionsArgument is optional, this will throw an erro since it could be null.

js/components/utils.js
```diff
@@ -432,37 +432,78 @@ var Utils = {
    * @function arrayDeepMerge
    * @param {Array} target - An array that will receive new items if additional items are passed
    * @param {Array} source - An array containing additional items to merge into target
-   * @param {Object} optionsArgument - optional parameters passed, i.e. arrayMerge function
+   * @param {Object} optionsArgument - optional parameters passed, i.e. arrayMerge, swap, unionBy, clone, arrayFusion
```
pilievOoyala: are these all the options?

js/components/utils.js
```diff
@@ -432,37 +432,78 @@ var Utils = {
...
    * @returns {Array} new merged array with items from both target and source
    */
   arrayDeepMerge: function(target, source, optionsArgument) {
+    // replaces target with source, no merge
```
aeng7: We may want to update this comment as target is untouched.

js/components/utils.js
```diff
@@ -432,37 +432,78 @@ var Utils = {
...
+    if (optionsArgument.arrayFusion === 'replace') {
+      return source;
+    }
+
+    var targetArray = optionsArgument.swap ? source : target;
+    var sourceArray = optionsArgument.swap ? target : source;
     var self = this;
+    var uniqueSourceArray = sourceArray.slice();
     var destination = [];
```
aeng7: We can just define var destination = targetArray.slice(); since that is the very next line.

js/components/utils.js
aeng7: Can we include some documentation on what each of these options do?

tests/components/utils-test.js
```diff
@@ -231,4 +236,43 @@ describe('Utils', function () {
...
+    var metaDataSettings = {"closedCaptionOptions":{"fontSize":"Large","windowColor":"Green"},"buttons":{"desktopContent":[{"name":"share","location":"controlBar","whenDoesNotFit":"moveToMoreOptions","minWidth":45,"enabled":true},{"name":"volume","location":"controlBar","whenDoesNotFit":"keep","minWidth":45,"enabled":true},{"name":"fullscreen","location":"controlBar","whenDoesNotFit":"keep","minWidth":55,"enabled":true},{"name":"quality","location":"controlBar","whenDoesNotFit":"moveToMoreOptions","minWidth":45,"enabled":true}]},"general":{"accentColor":"#ffbb00","watermark":{"imageResource":{"url":"http://ak.c.ooyala.com/Uzbm46asiensk3opIgwfFn5KFemv/watermark147585568"},"position":"top-left","clickUrl":"","transparency":0.51,"scalingOption":"none","scalingPercentage":0}},"shareScreen":{"shareContent":["social","ooyala"],"socialContent":["twitter","lisa","google+","jason"]}};
```
pilievOoyala: can you space all these out in to separate lines so it's clearer what's been tested?

js/components/utils.js
```diff
@@ -432,37 +432,78 @@ var Utils = {
...
+            // gracefully merge buttons by name
+            if (sourceItem[optionsArgument.unionBy] === targetItem[optionsArgument.unionBy]) {
+              var targetObject = optionsArgument.swap ? sourceItem : targetItem;
+              var sourceObject = optionsArgument.swap ? targetItem : sourceItem;
+              destination[j] = DeepMerge(targetObject, sourceObject, optionsArgument);
+
+              // prunes uniqueSourceArray to unique items not in target
+              if (optionsArgument.arrayFusion === 'prepend' && uniqueSourceArray && uniqueSourceArray.length) {
+                for (var x in uniqueSourceArray) {
+                  if (uniqueSourceArray[x][optionsArgument.unionBy] == sourceItem[optionsArgument.unionBy]) {
```
aeng7: Should we do triple equals here?

tests/components/utils-test.js
```diff
@@ -231,4 +236,43 @@ describe('Utils', function () {
     isValidString = Utils.isValidString(src);
     expect(isValidString).toBeTruthy();
   });
-});
+
+  it('tests deep merge', function () {
+    var localSettings = {"closedCaptionOptions":{"windowColor":"Yellow","enabled":true, "backgroundOpacity":"0.2","textOpacity":"1"}};
```
pilievOoyala: Is there a test for arrayDeepMerge?

js/components/utils.js
```diff
@@ -432,37 +432,78 @@ var Utils = {
    * @function arrayDeepMerge
    * @param {Array} target - An array that will receive new items if additional items are passed
    * @param {Array} source - An array containing additional items to merge into target
-   * @param {Object} optionsArgument - optional parameters passed, i.e. arrayMerge function
+   * @param {Object} optionsArgument - optional parameters passed, i.e. arrayMerge, swap, unionBy, clone, arrayFusion
    * @returns {Array} new merged array with items from both target and source
    */
   arrayDeepMerge: function(target, source, optionsArgument) {
```
aeng7: Can we have a walkthrough on this function? It will help understanding and could potentially catch small things.

js/components/utils.js
mistermendez: I added more info to the description of this PR that links to deep merge lib and how this function is used.

We don't call this function directly, it's passed as a callback using `arrayMerge` param in `optionsArgument` object. When this function is called `optionsArgument` will always at least have `arrayMerge`.

Will update doc language to remove "optional" for this param

js/components/utils.js
mistermendez: yes, will update

js/components/utils.js
mistermendez: good point, will update

mistermendez:
@aeng7 @pilievOoyala updates from comments made, ready for review

aeng7:
LGTM!

PR #704, opened 2016-12-15
[PLAYER-212] - Showing Discovery countdown after exiting fullscreen on iOS

jpfranco:
[Link to ticket](https://jira.corp.ooyala.com/browse/PLAYER-212)

These changes are meant to mimic the way V3 handles autoplay for Discovery videos on iOS. Since we can't show the *Up Next Panel* on fullscreen, we wait until the user exits fullscreen and then we display the countdown. The discovery screen's counter will be used if either `upNext.showUpNext` or `discoveryScreen.showCountDownTimerOnEndScreen` are configured.

This should be merged along with these other PR's:
https://git.corp.ooyala.com/projects/PBW/repos/mjolnir/pull-requests/16
https://github.com/ooyala/html5-video-plugins/pull/258

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -282,6 +282,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       this.state.videoQualityOptions.availableBitrates = null;
       this.state.videoQualityOptions.selectedBitrate = null;
       this.state.closedCaptionOptions.availableLanguages = null;
+      this.state.closedCaptionOptions.cueText = null;
       this.state.closedCaptionsInfoCache = {};
       this.state.discoveryData = null;
       this.state.thumbnails = null;
--- a/js/controller.js
+++ b/js/controller.js
@@ -40,6 +40,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       "screenToShow": null,
       "playerState": null,
       "discoveryData": null,
+      "forceCountDownTimerOnEndScreen": false,
       "isPlayingAd": false,
       "adOverlayUrl": null,
       "showAdOverlay": false,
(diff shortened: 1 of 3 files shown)


js/skin.js
```diff
@@ -300,4 +301,4 @@ Skin.defaultProps = {
 
 };
 
-module.exports = Skin;
\ No newline at end of file
+module.exports = Skin;
```
pilievOoyala: I think we need to get the new line, just in case.

pilievOoyala:
@mistermendez Can you double check this as well. I just want to make sure we do clean up on the state correctly.

js/controller.js
```diff
@@ -990,6 +991,19 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
       this.state.fullscreen = false;
+      var showUpNext = this.skin.props.skinConfig.upNext.showUpNext || this.skin.props.skinConfig.discoveryScreen.showCountDownTimerOnEndScreen;
+
+      // [PLAYER-212]
+      // We can't show UI such as the "Up Next" countdown on fullscreen iOS. If a countdown
+      // is configured, we wait until the user exits fullscreen and then we display it.
+      if (showUpNext && this.state.playerState === CONSTANTS.STATE.END) {
+        this.state.forceCountDownTimerOnEndScreen = true;
+        this.sendDiscoveryDisplayEvent("endScreen");
+        this.state.pluginsElement.addClass("oo-overlay-blur");
```
jpfranco: @pilievOoyala I think this is handled when the discovery screen is closed by `toggleDiscoveryScreen`, right @mistermendez ?

js/skin.js
jpfranco: This is done automatically by Atom, let me check to see if I can make it not do that.

jpfranco: @pilievOoyala btw, my text editor added a new line, it didn't have one before.

jpfranco:
Guys I made another small update to fix an issue that I had missed. The current caption that is being displayed needs to be cleared when switching videos, otherwise it will get stuck now that the video element is no longer disposed.

PR #705, opened 2016-12-16
[PBW-6265] removing transition animation to fix the problem in Safari

ksunika:


pilievOoyala:
Can we remove this conditionally based on browser?

ksunika:
@pilievOoyala, there are two possible ways to do it:
1) add one more class and programmatically apply transition property when it's not Safari (not through CSS)
2) use one of the hacky ways to separate Safari and Chrome for the CSS (though it's not a very clean way, and not always works)

However, the functionality we are losing here is not even that noticeable and was never a formal requirement (showed to Steve, he agreed). I'm not sure it's worth the trouble of having it at all.
What do you think?

pilievOoyala:
Since Steve is fine with it, then I have no problem with it.

PR #707, opened 2017-01-06
[PLAYER-20] fixes problem with social content arrays from backlot not…

ksunika:
… respected; removing `url` content since we don`t have the tab

There are two changes, one makes sure that if embedded skin config has shareScreen.shareContent == ['social', 'embed'], but the backlot has only shareScreen.shareContent == ['social'], the result is the backlot array, not the embedded.

The second change is for removing the 'url' from the list of shareContent values. As I understand, we might use it later, but now share screen can't handle it.


mistermendez:
👍  LGTM. @ksunika can you review my latest update and merge if you approve.

I added a param for array merge that defaults to 'replace'.

ksunika:
looks good to me!

PR #708, opened 2017-01-09
update skin to subscribe to external API event to change caption language

mistermendez:
⚠️  must merge with - https://git.corp.ooyala.com/projects/PBW/repos/mjolnir/pull-requests/174/overview


mityaha:
Looks good.  In general, I recommend to split one big controller.js unit test into a small tests each verifying that certain calls are made, events are triggered, states are set.

mistermendez:
@mityaha validation added to unit test

PR #709, opened 2017-01-11
[mediaset-10]

aeng7:
-the ad panel will now calculate the ad countdown based on a new adVideoPlayhead/currentAdPlayhead properties
-Fixes mediaset issue 3 (not 10, branch named incorrectly)


aeng7:
How should I add unit tests for this change?

mistermendez:
this has an existing unit test, as long as it doesn't break, that is sufficient.

👍  LGTM

PR #713, opened 2017-01-20
[PBW-6348] - Disabling pointer cursor on non-clickable time display

jpfranco:
[Ticket](https://jira.corp.ooyala.com/browse/PBW-6348)

Since the time display isn't clickable, a customer suggested that we shouldn't display a pointer cursor over it.


ksunika:
LGTM

PR #714, opened 2017-01-21
[PLAYER-329] - Triggering RELATED_VIDEOS_DISPLAYED event when Up Next panel is shown

jpfranco:
[Ticket](https://jira.corp.ooyala.com/browse/PLAYER-329)

The analytics backend requires that a `Related Videos Displayed` event be fired before a `Related Video Selected`, otherwise the latter event will be discarded. On V4 `Related Videos Displayed` was only getting fired when the Discovery screen was shown, which means that all the Discovery related data was getting lost if the _Up Next_ panel was used.

These changes trigger a `Related Videos Displayed` event when the _Up Next_ panel is first shown.


mistermendez:
👍

PR #715, opened 2017-01-24
[PLAYER-423] moving moving externalPluginSubscription to earlier

ksunika:
The timing of subscribing to OO.EVENTS.DISCOVERY_API.RELATED_VIDEOS_FETCHED was too late, sometimes we were subscribing *after* the even was fired. Hence, the discovery was not usable.

I am moving the subscription up, to the earliest point where discovery plugin is already loaded.

[PLAYER-423](https://jira.corp.ooyala.com/browse/PLAYER-423)


mityaha:
Looks good

PR #718, opened 2017-01-31
sanitize config data to prevent empty array from replacing player config

mistermendez:


js/components/utils.js
```diff
@@ -529,6 +529,21 @@ var Utils = {
...
   },
 
+  /**
+   * Returns sanitized config data
+   *
+   * @function sanitizeConfigData
+   * @param {Object} data to be sanitized
+   * @returns {Object} data if data is valid, empty object if not
+   */
+  sanitizeConfigData: function(data) {
```
pilievOoyala: The name for this is a generic kind of name and hints at it being useful for any data. Do we not accept arrays at all in the skin params config? If not I would rename it.

js/components/utils.js
```diff
@@ -529,6 +529,21 @@ var Utils = {
...
+    if (data && (!Array.isArray(data))) {
+      return data
+    } else {
+      return {};
```
pilievOoyala: For our debugging purposes can we get a debug log statement in the negative case. To at least let the developers or our team know that bad data is coming in from somewhere.

tests/controller-test.js
```diff
@@ -242,6 +242,7 @@ OO = {
     Html5Skin.onPlayerCreated.call(controllerMock, 'customerUi', elementId, {skin:{config:{}}}, persistentSettings);
     Html5Skin.onSkinMetaDataFetched.call(controllerMock, 'customerUi', {});
     Html5Skin.loadConfigData.call(controllerMock, 'customerUi', {skin:{config:{}}}, {}, {}, {});
```
pilievOoyala: Do we need tests for skin:{config:[]} as well?  And {inline:{}}, just in case. Or do those already exist.

tests/components/utils-test.js
```diff
@@ -237,6 +237,30 @@ describe('Utils', function () {
...
+
+    data = '';
+    sanitizedConfigData = Utils.sanitizeConfigData(data);
+    expect(sanitizedConfigData).toEqual({});
+
+    data = undefined;
+    sanitizedConfigData = Utils.sanitizeConfigData(data);
+    expect(sanitizedConfigData).toEqual({});
+
+    data = [];
```
pilievOoyala: what about an array that has data in it?

tests/components/utils-test.js
```diff
@@ -237,6 +237,30 @@ describe('Utils', function () {
     expect(isValidString).toBeTruthy();
   });
 
+  it('tests sanitizeConfigData', function () {
+    var data = null;
+    var sanitizedConfigData = Utils.sanitizeConfigData(data);
+    expect(sanitizedConfigData).toEqual({});
+
+    data = '';
```
pilievOoyala: how about a string that isn't empty? Then it won't be falsy. Also, what if data is a number.

js/components/utils.js
mistermendez: we don't accept playerParam.skin.inline being an `[]`

an Object with array is still acceptable `{"buttons": [1, 2, 3]}`

`sanitizeData` is generic `sanitizeConfigData` is just for CONFIG data

js/components/utils.js
mistermendez: sure

tests/controller-test.js
mistermendez: sure will add test for those

tests/components/utils-test.js
mistermendez: array with data is still invalid. are you suggesting we add a test for array with data?

tests/components/utils-test.js
mistermendez: Strings and numbers are valid, the data type we need to prevent here are Arrays.

The deepmerge function already checks if value is a "[mergeable Object](https://github.com/KyleAMathews/deepmerge/blob/master/index.js#L11)"

mistermendez:
@pilievOoyala updates made

tests/components/utils-test.js
pilievOoyala: Yes. Just in case it somehow passed when it shouldn't

tests/components/utils-test.js
pilievOoyala: Can you add tests that assure that non empty string and numbers return something valid.

js/components/utils.js
pilievOoyala: OK sounds good.

tests/components/utils-test.js
mistermendez: okay added

tests/components/utils-test.js
mistermendez: yup, added

pilievOoyala:
LGTM. Thank you!

PR #719, opened 2017-02-03
[PLAYER-556] Reworked to accept all player parameters

mityaha:


amp_iframe.html
```diff
@@ -4,67 +4,76 @@
...
-    // Parse docUrl
-    var docUrl = undefined;
-    if (!!queryParams.match(/docUrl=([^&]*)/)) {
-      playerParam["docUrl"] = decodeURIComponent(queryParams.match(/docUrl=([^&]*)/)[1]);
-    }
+    function queryStringToJSON(qs) {
+      var pairs = qs.split('&');
+      var result = {};
+      pairs.forEach(function(pair) {
+        var pair = pair.split('=');
```
mistermendez: should `var pair` have another name? the argument to the function is also `pair`

amp_iframe.html
```diff
@@ -4,67 +4,76 @@
...
+            result[key].push(value);
+          } else {
+            result[key] = [result[key], value];
           }
-          if (typeof currentLevel !== 'object')
-            throw("Exception: Invalid query parameter");
+        } else {
+          result[key] = value;
+        }
+      });
```
mistermendez: semicolon not needed here

iframe.html
```diff
@@ -13,62 +13,66 @@
...
-    } else if (document.referrer) {
-      playerParam["docUrl"] = document.referrer;
-    } else {
-      console.log("iframe.html - document.referrer was not set");
-    }
+    function queryStringToJSON(qs) {
+      var pairs = qs.split('&');
+      var result = {};
+      pairs.forEach(function(pair) {
+        var pair = pair.split('=');
```
mistermendez: should `var pair` have another name? the argument to the function is also `pair`

iframe.html
```diff
@@ -13,62 +13,66 @@
...
-            currentLevel[keySplits[0]] = true;
-          else if (paramValue.toLowerCase() === "false")
-            currentLevel[keySplits[0]] = false;
-          else if (paramValue[0] === "[" &&
-            paramValue[paramValue.length -1] === "]")
-            currentLevel[keySplits[0]] = JSON.parse(decodeURIComponent(match[2]));
-          else
-            currentLevel[keySplits[0]] = decodeURIComponent(match[2]);
+      return JSON.parse(JSON.stringify(result));
+    };
```
mistermendez: semicolon not needed here

amp_iframe.html
```diff
@@ -4,67 +4,76 @@
...
+        console.log("iframe.html - document.referrer was not set");
       }
     }
 
     OO.ready(function() {
-      window.pp = OO.Player.create('container', queryParams.match(/ec=([^&]*)/)[1], playerParam);
+      window.pp = OO.Player.create('container', playerParams.ec, playerParams);
 
-      //Enable cross-iframe control
+        //Enable cross-iframe control
```
mistermendez: extra space not needed

amp_iframe.html
```diff
@@ -4,67 +4,76 @@
...
-      playerParam["docUrl"] = decodeURIComponent(queryParams.match(/docUrl=([^&]*)/)[1]);
-    }
+    function queryStringToJSON(qs) {
+      var pairs = qs.split('&');
+      var result = {};
+      pairs.forEach(function(pair) {
+        var pair = pair.split('=');
+        var key = pair[0];
+        var value = decodeURIComponent(pair[1] || '');
+        if (value === "true") {
```
mistermendez: would switch statement work better than if/else here? 

the logic is the same but switch is more efficient than else/if. see - http://stackoverflow.com/questions/449273/why-the-switch-statement-and-not-if-else

amp_iframe.html
```diff
@@ -4,67 +4,76 @@
...
+            source: "<iframe width='640' height='480' frameborder='0' allowfullscreen src='//player.ooyala.com/static/v4/stable/latest/skin-plugin/iframe.html?ec=<ASSET_ID>&pbid=<PLAYER_ID>&pcode=<PUBLISHER_ID>'></iframe>"
+          }
         }
       }
-      catch (e) {
-        throw("Exception: Invalid query parameter");
+    };
+    playerParams["playerBrandingId"] = playerParams.pbid;
+    playerParams["skin"] = skin;
+    delete playerParams.pbid;
```
mistermendez: why do we delete here?

iframe.html
```diff
@@ -13,62 +13,66 @@
...
+            source: "<iframe width='640' height='480' frameborder='0' allowfullscreen src='//player.ooyala.com/static/v4/stable/latest/skin-plugin/iframe.html?ec=<ASSET_ID>&pbid=<PLAYER_ID>&pcode=<PUBLISHER_ID>'></iframe>"
+          }
         }
       }
-      catch (e) {
-        throw("Exception: Invalid query parameter");
+    };
+    playerParams["playerBrandingId"] = playerParams.pbid;
+    playerParams["skin"] = skin;
+    delete playerParams.pbid;
```
mistermendez: why do we delete here?

mityaha: We may leave it, but this will be unused parameter, it's cleaner this way I believe.

amp_iframe.html
mityaha: Why not?

amp_iframe.html
mityaha: It doesn't look to me that switch here is feasible as these are some of conditions are logical, not string comparisons.

amp_iframe.html
mityaha: Good point, fixed

iframe.html
mityaha: Fixed

iframe.html
mityaha: Fixed

amp_iframe.html
mityaha: Fixed

iframe.html
```diff
@@ -13,62 +13,66 @@
...
-      console.log("iframe.html - document.referrer was not set");
-    }
+    function queryStringToJSON(qs) {
+      var pairs = qs.split('&');
+      var result = {};
+      pairs.forEach(function(pair) {
+        var pair = pair.split('=');
+        var key = pair[0];
+        var value = decodeURIComponent(pair[1] || '');
+        if (value === "true") {
```
mityaha: See my answer in another similar comment

amp_iframe.html
mityaha: We may leave it, but this will be unused parameter, it's cleaner this way I believe.

amp_iframe.html
mistermendez: it's a function

amp_iframe.html
mistermendez: coolio

iframe.html
mistermendez: okay

mistermendez:
👍

PR #720, opened 2017-02-09
[PLAYER-570] Fixed problem for youtube when video element is created …

mityaha:
…later

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -267,6 +267,8 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         if (Utils.isIE10()) {
           videoElement.attr("controls", "controls");
         }
+      } else {
+        OO.log("Could not find main video element")
       }
 
       if (params.videoId === OO.VIDEO.MAIN) {
@@ -1595,14 +1597,14 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
 
     //find descendant video element
     findMainVideoElement: function(element) {
-      var elements = [];
-      //use actual element
-      if (element[0]) {
-        element = element[0];
-      }
-
       //find html5 video
       try {
+        var elements = [];
+        //use actual element
+        if (element[0]) {
+          element = element[0];
+        }
+
         if (element.tagName && element.tagName.toLowerCase().indexOf(CONSTANTS.MEDIA_TYPE.VIDEO) != -1) {
           this.state.mainVideoMediaType = CONSTANTS.MEDIA_TYPE.HTML5;
         }
--- a/js/controller.js
+++ b/js/controller.js
@@ -256,19 +256,23 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
 
     onVcVideoElementCreated: function(event, params) {
       var videoElement = params.videoElement;
-      videoElement = this.findMainVideoElement(videoElement);
+      if (params.videoId === OO.VIDEO.MAIN) {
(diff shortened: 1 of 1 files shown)


js/controller.js
```diff
@@ -1598,28 +1602,32 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       }
```
pilievOoyala: we need an extra check in here to see if 'element' is valid when it's passed in.

pilievOoyala:
Just one minor remark since you are making changes in here anyway.

js/controller.js
```diff
@@ -256,19 +256,23 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+        if (Utils.isIE10()) {
+          videoElement.attr("controls", "controls");
+        }
       }
 
       if (params.videoId === OO.VIDEO.MAIN) {
-        this.state.mainVideoElement = videoElement;
+        if (videoElement) {
+          this.state.mainVideoElement = videoElement;
+        }
```
aeng7: Is this block redundant with lines 259-261 above?

js/controller.js
```diff
@@ -1598,28 +1602,32 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+        else if (element.getElementsByTagName(CONSTANTS.MEDIA_TYPE.OBJECT).length) {
+          elements = element.getElementsByTagName(CONSTANTS.MEDIA_TYPE.OBJECT);
+          if (elements.length) {
+            element = elements[0];
+            this.state.mainVideoMediaType = CONSTANTS.MEDIA_TYPE.FLASH;
+          }
+        }
+        return element;
+      } catch(err) {
+        return null;
```
aeng7: Should we output a console log here for easier debugging in the future?

js/controller.js
mityaha: No, it is not, because we want to overwrite the value mainVideoElement in case it was found.

js/controller.js
mityaha: What extra check do you have in mind?

js/controller.js
mistermendez: could we remove lines 259-261?

js/controller.js
pilievOoyala: I put this commetn in the wrong place.

js/controller.js
```diff
@@ -1598,28 +1602,32 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       }
 
       //find html5 video
-      if (element.tagName && element.tagName.toLowerCase().indexOf(CONSTANTS.MEDIA_TYPE.VIDEO) != -1) {
```
pilievOoyala: At the beginning of this function (outside of what the diff show by default) we access element[0] without checking if element is null.  Need an extra null check there.

js/controller.js
mityaha: No, we want to save initial value of mainVideoElement because on line 484 this value will be reused in the next attempt to obtain mainVideoElement

js/controller.js
mityaha: Good point, will do

js/controller.js
mityaha: Good point, I moved it into try/catch block

mityaha:
Closed it, as these fixes don't address the problem of accessing main video element from youtube video plugin.

PR #724, opened 2017-02-14
[PLAYER-556] Better parsing of URL parameters

mityaha:


amp_iframe.html
```diff
@@ -12,39 +12,50 @@
...
-    }
+      var nvp = query.split('&');
+      var data = {};
+      for (var i = 0 ; i < nvp.length ; i++) {
+        var pair = nvp[i].split('=');
+        var name = decodeURIComponent(pair[0]);
+        var value = decodeURIComponent(pair[1]);
+
+        var path = name.match(/(^[^\[]+)(\[.*\]$)?/);
+        var first = path[1];
```
pilievOoyala: what happens if path contains 0 elements? Need a check.

iframe.html
```diff
@@ -13,39 +13,50 @@
...
+
+      var nvp = query.split('&');
+      var data = {};
+      for (var i = 0 ; i < nvp.length ; i++) {
+        var pair = nvp[i].split('=');
+        var name = decodeURIComponent(pair[0]);
+        var value = decodeURIComponent(pair[1]);
+
+        var path = name.match(/(^[^\[]+)(\[.*\]$)?/);
+        var first = path[1];
```
pilievOoyala: Need to check that path actually contains elements here as well.

pilievOoyala: Don't we want to start with the 0th element, not 1?

mityaha: key-value: key is element #0 of pair, value is element #1, value may be encoded, so we want to decode it, so we deal with element #1 here, element #0 doesn't require decoding

amp_iframe.html
mityaha: Added

iframe.html
mityaha: No, zeroth element returns whole expression, matching parts start with index 1

amp_iframe.html
```diff
@@ -12,39 +12,56 @@
...
-        } 
-
-        if (result[key]) {
-          if (Array.isArray(result[key])) {
-            result[key].push(value);
-          } else {
-            result[key] = [result[key], value];
+    function queryStringToJSON(query) {
+      var setValue = function(root, path, value) {
+        if (path.length > 1){
```
pilievOoyala: I think it needs to be path && path.length.  What is path is null?

mityaha: Path has already been verified by code that calls setValue, this is actually redundant check altogether.

pilievOoyala:
LGTM

PR #725, opened 2017-02-14
[PLAYER-584] Fixed rendering CCs that contain new line character

mityaha:


js/components/textTrackPanel.js
```diff
@@ -127,8 +129,8 @@ var TextTrackPanel = React.createClass({
                   this.props.closedCaptionOptions.textEnhancement
                 )
               }
-              >
-              {this.props.cueText}
+             >
+             <span dangerouslySetInnerHTML={Utils.createMarkup(this.props.cueText)}></span>
```
pilievOoyala: What makes this dangerous and why are we using it?

mityaha: This is a question to @mistermendez, he is going for it, we use this pattern in a couple other places

mistermendez: https://facebook.github.io/react/docs/dom-elements.html#dangerouslysetinnerhtml

`dangerouslySetInnerHTML` is React's replacement for using `innerHTML` in the browser DOM. In general, setting HTML from code is risky because it's easy to inadvertently expose your users to a [cross-site scripting (XSS)](https://en.wikipedia.org/wiki/Cross-site_scripting) attack. So, you can set HTML directly from React, but you have to type out `dangerouslySetInnerHTML` and pass an object with a` __html` key, to remind yourself that it's dangerous.

mistermendez:
👍 LGTM

pilievOoyala:
Sidenote: We need a 007 emoji.

PR #728, opened 2017-02-17
[player-395] showing flash alerts when flash is disabled in the browser

ksunika:
[PLAYER-395](https://jira.corp.ooyala.com/browse/PLAYER-395).


aeng7:
Can we add some unit tests for this feature?

js/controller.js
```diff
@@ -668,6 +697,11 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         width: ""
       });
       this.state.forceControlBarVisible = (this.state.pluginsElement.children().length > 0);
+
+      if (this.state.midVideo) {
+        this.mb.publish(OO.EVENTS.PAUSE);
```
aeng7: How come we pause the video here (and resume above in adsPlayed)?

ksunika: Until we enable the flash, the call to pause the main content doesn't happen, so we have the flash alert visible and the main video playing in the background unless we pause it.

js/controller.js
```diff
@@ -656,6 +680,11 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       this.state.isPlayingAd = false;
       this.state.pluginsElement.removeClass("oo-showing");
       this.state.pluginsClickElement.removeClass("oo-showing");
+      this.showResponsiveDiv();
+      this.changeIMAiframeOpacity(0);
```
ksunika: I think it's because we manually change the opacity to 1 before that, it doesn't change back to 0 by itself.

PR #721, opened 2017-02-10
Make sure player destroy removes attached keyboard eventlisteners

mariusberg:
Seems to be an issue when video returns an error on load. And the keyboard events will still be attached even if the player is unmounted.

PR #729, opened 2017-02-17
Make sure player destroy removes attached keyboard eventlisteners

mariusberg:
Seems to be an issue when video returns an error on load. And the keyboard events will still be attached even if the player is unmounted.

Resubmit https://github.com/ooyala/html5-skin/pull/721 to master per request.


pilievOoyala:
Looks good. Thank you!

aeng7:
LGTM as well.

PR #733, opened 2017-02-28
[fix-unit-test]

aeng7:
-added mock cleanUpEventListeners function


aeng7:
Walked through changes with Mike. @lental approves.

PR #736, opened 2017-03-02
[PBF-808] Don't show discovery, quality, CC, share and more options w…

nagavenivn:
…hen ooyala ad is playing

### DIFF
--- a/tests/components/controlBar-test.js
+++ b/tests/components/controlBar-test.js
@@ -417,6 +417,80 @@ describe('ControlBar', function () {
     expect(toggleScreenClicked).toBe(true);
   });
 
+  it('hides closed caption button if ooyala ad is playing', function() {
+    var mockController = {
+      state: {
+        isMobile: false,
+        volumeState: {
+          volume: 1
+        },
+        closedCaptionOptions: {availableLanguages: true},
+        videoQualityOptions: {
+          availableBitrates: null
+        },
+        isOoyalaAds: true
+      }
+    };
+
+    var oneButtonSkinConfig = Utils.clone(skinConfig);
+    oneButtonSkinConfig.buttons.desktopContent = [
+      {"name":"closedCaption", "location":"controlBar", "whenDoesNotFit":"moveToMoreOptions", "minWidth":35 }
+    ];
+
+    var mockProps = {
+      isLiveStream: false,
+      controller: mockController,
+      skinConfig: oneButtonSkinConfig
+    };
+
+    var DOM = TestUtils.renderIntoDocument(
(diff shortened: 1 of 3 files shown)


nagavenivn:
@aeng7 Could you please do CR?

tests/components/controlBar-test.js
```diff
@@ -600,6 +816,95 @@ describe('ControlBar', function () {
...
+      <ControlBar {...mockProps} controlBarVisible={true}
+        componentWidth={100}
+        playerState={CONSTANTS.STATE.PLAYING}
+        isLiveStream={mockProps.isLiveStream} />
+    );
+
+    optionsButton = TestUtils.scryRenderedDOMComponentsWithClass(DOM, 'oo-more-options');
+    expect(optionsButton.length).toBe(1);
+    buttons = TestUtils.scryRenderedDOMComponentsWithClass(DOM, 'oo-play-pause');
+    expect(buttons.length).toBeLessThan(5);
```
aeng7: Where does less than 5 come from?

tests/components/controlBar-test.js
```diff
@@ -600,6 +816,95 @@ describe('ControlBar', function () {
...
+      <ControlBar {...mockProps} controlBarVisible={true}
+        componentWidth={100}
+        playerState={CONSTANTS.STATE.PLAYING}
+        isLiveStream={mockProps.isLiveStream} />
+    );
+
+    optionsButton = TestUtils.scryRenderedDOMComponentsWithClass(DOM, 'oo-more-options');
+    expect(optionsButton.length).toBe(0);
+    buttons = TestUtils.scryRenderedDOMComponentsWithClass(DOM, 'oo-play-pause');
+    expect(buttons.length).toBe(1);
```
aeng7: Why do we expect buttons length to be 1 here?

nagavenivn: @aeng7 Because of the entry in oneButtonSkinConfig.buttons.desktopContent is to keep even when it doesn't fit the size
    oneButtonSkinConfig.buttons.desktopContent = [
      {"name":"playPause", "location":"controlBar", "whenDoesNotFit":"keep", "minWidth":35 },

tests/components/controlBar-test.js
nagavenivn: @aeng7 This test case is based on the existing one "shows/hides the more options button appropriately".  I just added isOoyalaAds: true in the controller options. As far as my understanding there are 5 entries in the oneButtonSkinConfig.buttons.desktopContent = [ and one of them will go in the more option button based on the size of componentWidth={100} in the DOM

PR #737, opened 2017-03-03
[pbw-6443-transition]

aeng7:
-instead of the start screen, the loading screen will be shown when transitioning from an Ooyala ad to the main content


dchauOoyala:
LGTM

PR #740, opened 2017-03-07
[player-395b]

aeng7:
-fixed an issue where IMA overlays were made transparent


jpfranco:
@aeng7 I probably should've caught this on the previous PR, but on line 1687 we have this:

`var IMAiframe = $("iframe[src^='http://imasdk.googleapis.com/']")[0];`

It looks like this might break when using `https` and in general seems a bit brittle. Maybe we can check for both protocols, although this would still break if Google changes that URL in the future.

aeng7:
I just sent out an email to IMA SDK support asking how to best access the iframe. We'll await their response.

PR #742, opened 2017-03-22
[PLAYER-509]

dchauOoyala:
- Update references to stable/latest to production/latest
- Fix bad references


dchauOoyala:
I could not completely remove the word, `stable`, from use because we still have a `stable` branch where the Jenkins job manually updates the player file references to a specific `stable/$branch` version.

rcvrgs:
@dchauOoyala, for which cases you say that you weren't able to remove stable? I see you changed them all in the files you changed.

dchauOoyala:
Only this line:
https://github.com/ooyala/html5-skin/blame/master/README.md#L154

But I think that is okay.

rcvrgs:
In that line, at the end it says "we recommend using latest stable for production", it is under the hot to contribute section, so do you mean that they should be using stable to create new branches for PRs instead of master? If that's the case the it would be better if it's like "we recommend using latest stable to contribute".

dchauOoyala:
I think that this comment is to clarify that users should use `stable` for "production" purposes. As in if they are self hosting, they should use the skin files under `stable`, whereas development and pull requests should be done under `master`, as is mentioned in the paragraph above it.

README.md
```diff
@@ -165,4 +165,4 @@ Before submitting a pull request, please make sure the following is done:
 3. Ensure the test suite passes (`npm test`).
 
 ###Style Guide
```
rcvrgs: This isn't part of your PR, but I see that we need a space between the ### and Style, if we put it together it will render the ### characters and not the Style Guide as a header.

rcvrgs: That also applies for the ##How to Contribute and ###Submitting a Pull Request, but I don't know why I can't comment those lines.

rcvrgs:
Ok, thanks for the comments @dchauOoyala. I feel confused now xD do customers need to be using production or stable? Or would it be the same for them?

PR #743, opened 2017-03-27
Fix the issue in Safari where 'CC' icon is always displayed

kuu:
[Issue] 'CC' icon is always displayed in Safari.
[Cause] In Safari, CLOSED_CAPTIONS_INFO_AVAILABLE event is always issued even when the assets does not have CC.
[Fix] Ignore the event if it does not contain a valid language code.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -951,6 +951,11 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       if (!info || !info.videoId || !info.languages) {
         return;
       }
+      // A workaround for Safari's bug
+      var languages = info.languages;
+      if (languages && languages.length === 1 && languages[0] === 'CC') {
+        return;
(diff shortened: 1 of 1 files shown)


jpfranco:
Thanks for taking the time to contribute @kuu. Unfortunately these changes would result in a false positive that would cause the CC button to be hidden for videos that do have in-stream (or in-manifest) captions.

You can see what I mean by testing with the following embed code: Ztb2l1ZzravsTPzgVqX0taL-bvO-sLRD

Please note that pull requests need to be based off the master branch. We've updated the readme file of the project in order to clarify that.

kuu:
@jpfranco To disable CC icon using skin config, I needed to change code.
Please review the following the separate [PR](https://github.com/ooyala/html5-skin/pull/760).

PR #746, opened 2017-04-07
[PLAYER-760] Disabling anamorphic video fix when ads are active

jpfranco:
[Ticket](https://jira.corp.ooyala.com/browse/PLAYER-760)

The fix for distorted anamorphic videos on Firefox was being applied to both main content and ads. After these changes the fix will no longer be active during ad playback, so the ads will no longer get stretched.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -1485,9 +1485,11 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         // Only enable anamorphic videos fix if video actually requires it
         if (isAnamorphic === true || isAnamorphic === 'true') {
           this.state.mainVideoInnerWrapper.addClass('oo-anamorphic');
+          OO.log("Anamorphic video fix: ON");
         }
       } else {
         this.state.mainVideoInnerWrapper.removeClass('oo-anamorphic');
+        OO.log("Anamorphic video fix: OFF");
       }
     },
 
--- a/js/controller.js
+++ b/js/controller.js
@@ -1485,11 +1485,11 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         // Only enable anamorphic videos fix if video actually requires it
         if (isAnamorphic === true || isAnamorphic === 'true') {
           this.state.mainVideoInnerWrapper.addClass('oo-anamorphic');
-          OO.log("Anamorphic video fix: ON");
+          OO.log('Anamorphic video fix: ON');
         }
       } else {
         this.state.mainVideoInnerWrapper.removeClass('oo-anamorphic');
-        OO.log("Anamorphic video fix: OFF");
+        OO.log('Anamorphic video fix: OFF');
       }
     },
 
--- a/js/controller.js
+++ b/js/controller.js
@@ -1474,6 +1474,13 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       this.mb.publish(OO.EVENTS.SAVE_PLAYER_SETTINGS, this.state.persistentSettings);
     },
 
+    /**
+     * Used to enable or disable the CSS workaround that prevents anamorphic videos from
+     * being distorted on Firefox. The fix will only be enabled if ots_stretch_to_output
+     * is set to true in the player attributes.
+     * Note that currently the oo-anamorphic class has effect only on Firefox.
+     * @param {boolen} enabled A value that determines whether to enable or disable the anamorphic videos CSS fix.
+     */
     trySetAnamorphicFixState: function(enabled) {
       if (!this.state || !this.state.mainVideoInnerWrapper) {
         return;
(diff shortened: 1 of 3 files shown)


tests/controller-test.js
```diff
@@ -244,6 +245,7 @@ OO = {
     Html5Skin.onPlayerCreated.call(controllerMock, 'customerUi', elementId, {skin:{config:{}}}, persistentSettings);
     Html5Skin.onSkinMetaDataFetched.call(controllerMock, 'customerUi', {});
     Html5Skin.onAttributesFetched.call(controllerMock, 'customerUi', {"attributes":{"provider":{"ots_stretch_to_output":"true"}}});
+    Html5Skin.trySetAnamorphicFixState.call(controllerMock, true);
```
jpfranco: I think our unit tests for the skin controller need some improvement. I spent some time trying to write some tests for the scenarios in this ticket, but it looks like we don't currently have the setup required to test the actual logic of the controller. 

I think that creating a new test environment for the controller will need to be handled as a separate task.

js/controller.js
```diff
@@ -1472,6 +1474,25 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       this.mb.publish(OO.EVENTS.SAVE_PLAYER_SETTINGS, this.state.persistentSettings);
     },
 
+    trySetAnamorphicFixState: function(enabled) {
```
aeng7: Should we start some JS docs for this file? I know the standard was not to include them, but I feel it will prove helpful for maintenance.

tests/controller-test.js
aeng7: I had the same problem when writing unit tests for some Ooyala Ads changes. Can you check out the:

"describe('Controller testing Ooyala Ads', function () {"

section below. Would we be able to call the function and check the mock controller state like what was done in the 'Controller testing Ooyala Ads' tests?

js/controller.js
jpfranco: Sounds good, I'll add that.

jpfranco:
@aeng7 I implemented your suggestions. I'm still not entirely happy about the unit tests, but they should at least cover the basic scenarios now.

jpfranco:
Closing for small update...

aeng7:
@jpfranco Yep, agreed. We need a unit test rehaul for skin at some point. Right now its pretty much testing one happy path scenario as a whole.

js/controller.js
```diff
@@ -1472,6 +1474,32 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       this.mb.publish(OO.EVENTS.SAVE_PLAYER_SETTINGS, this.state.persistentSettings);
     },
 
+    /**
+     * Used to enable or disable the CSS workaround that prevents anamorphic videos from
+     * being distorted on Firefox. The fix will only be enabled if ots_stretch_to_output
+     * is set to true in the player attributes.
+     * Note that currently the oo-anamorphic class has effect only on Firefox.
+     * @param {boolen} enabled A value that determines whether to enable or disable the anamorphic videos CSS fix.
```
dchauOoyala: typo: `boolean`

js/controller.js
```diff
@@ -1472,6 +1474,32 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+     */
+    trySetAnamorphicFixState: function(enabled) {
```
dchauOoyala: I think this `trySetAnamorphicFixState` would be easier understood as two separate functions because the function does not necessarily do what the function name suggests depending on the `enabled` argument. Perhaps we should split it into `enableAnamorphicFixState` and `disableAnamorphicFixState`? That way, it is also immediately clear to the reader what the function is doing as opposed to some ambiguity of `trySetAnamorphicFixState(true)` and `trySetAnamorphicFixState(false)`

js/controller.js
jpfranco: Good catch, I'll update this

js/controller.js
jpfranco: I think the current code strikes a good balance between readability and code reuse. Do you think it would help if we used a different name for this function? In my mind the name means that you try to set the state of the fix (either enabled or disabled), but it can probably be improved.

dchauOoyala: Ahh I see, for some reason I didn't really register the "state" part of the function name. Makes more sense now. What do you think about `setAnamorphicFixState` as opposed to `trySetAnamorphicFixState`?

dchauOoyala: LGTM otherwise

jpfranco: I added the `try` because the fix is not applied unless it is configured for that video. If it's confusing I can change it though.

dchauOoyala: I see, that makes sense. I just completely read the jsdoc

dchauOoyala:
LGTM

PR #748, opened 2017-04-28
Fix issue with breakepoint blindspots

mariusberg:
If the container/componentWidth is hitting within the offset ranges 1px unhanded spacing. The player will fail. As the mixin returns false for all responsive breakpoints. E.G componentWidth = 559.xxx 

By some magic we where hitting this for playback in our app on iPad landscape view.


jpfranco:
Hello @mariusberg, sorry about the delay. This update has been recreated in #856 and has been merged to master. We're planning to include this update on our next release if it passes QA.

PR #757, opened 2017-06-09
[PLAYER-1016] Basic accessibility support

jpfranco:
[Ticket](https://jira.corp.ooyala.com/browse/PLAYER-1016)

Partial accessibility support which includes:

- Keyboard navigation for the big play button and a subset of the control bar buttons.
- ARIA labels for these same buttons

**IMPORTANT: We are waiting for feedback from ManU, so please review but don't merge yet**

### DIFF
--- a/js/components/utils.js
+++ b/js/components/utils.js
@@ -305,6 +305,53 @@ var Utils = {
     return isFinite(Number(property)) ? Number(property) : 0;
   },
 
+
+  /**
+   * Determines whether an element contains a class or not.
+   * TODO:
+   * classList.contains is much better for this purpose, but our current version
+   * of React Test Utils generates events with a null classList, which results in
+   * broken unit tests.
+   *
+   * @param {DOMElement} element The DOM element which we want to check
+   * @param {String} className The name of the class we want to match
+   * @return {Boolean} True if the element contains the given class, false otherwise
+   */
+  hasClass: function(element, className) {
+    if (!element) {
+      return false;
+    }
+    return (' ' + element.className + ' ').indexOf(' ' + className + ' ') > -1;
+  },
+
+  /**
+   * Returns the icon element associated with an event (usually mouseover or mouseout),
+   * which can be either the event's target element itself or a child of the target element.
+   * The icon is matched with a class name.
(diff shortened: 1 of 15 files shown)


js/components/utils.js
```diff
@@ -305,6 +305,53 @@ var Utils = {
...
+   *
+   * @param {DOMElement} element The DOM element which we want to check
+   * @param {String} className The name of the class we want to match
+   * @return {Boolean} True if the element contains the given class, false otherwise
+   */
+  elementHasClass: function(element, className) {
+    if (!element) {
+      return false;
+    }
+    return (' ' + element.className + ' ').indexOf(' ' + className + ' ') > -1;
```
aeng7: What is the purpose of the extra spaces (' ') appended and prepended to the classname?

js/components/utils.js
```diff
@@ -314,9 +361,7 @@ var Utils = {
   highlight: function(target, opacity, color) {
     target.style.opacity = opacity;
     target.style.color = color;
-    target.style.WebkitFilter = "drop-shadow(0px 0px 3px rgba(255,255,255,0.8))";
-    target.style.filter = "drop-shadow(0px 0px 3px rgba(255,255,255,0.8))";
-    target.style.msFilter = "progid:DXImageTransform.Microsoft.Dropshadow(OffX=0, OffY=0, Color='#fff')";
+    target.style.textShadow = "0px 0px 3px rgba(255, 255, 255, 0.5), 0px 0px 6px rgba(255, 255, 255, 0.5), 0px 0px 9px rgba(255, 255, 255, 0.5)";
```
aeng7: Does this need to match the highlight-base mixin defined in the _mixins.scss file? If so, is there a way to link them together? If not, can we add a comment to help with maintenance in case we need to change this style in the future?

aeng7:
Should this be merged into master instead of stable?

js/components/utils.js
jpfranco: This is in order to prevent a false positive when a class name contains the string that you're testing. For example, if you have an element with a class `fancyclass`, then doing `hasClass('fancy')` should return `false`

jpfranco:
Wrong branch, closing.

PR #758, opened 2017-06-09
[PLAYER-1016] Basic accessibility support

jpfranco:
[Ticket](https://jira.corp.ooyala.com/browse/PLAYER-1016)

Partial accessibility support which includes:

- Keyboard navigation for the big play button and a subset of the control bar buttons.
- ARIA labels for these same buttons

### DIFF
--- a/js/components/accessibilityControls.js
+++ b/js/components/accessibilityControls.js
@@ -23,8 +23,17 @@ AccessibilityControls.prototype = {
     var currentTime;
     var newPlayheadTime;
     var newVolume;
+    var targetTagName;
 
-    if (e.keyCode === CONSTANTS.KEYCODES.SPACE_KEY){
+    if (e.target && typeof e.target.tagName === "string") {
+      targetTagName = e.target.tagName.toLowerCase();
+    }
+
+    // We override the default behavior when the target element is a button (pressing
+    // the spacebar on a button should activate it).
+    // Note that this is not a comprehensive fix for all clickable elements, this is
+    // mostly meant to enable keyboard navigation on control bar elements.
+    if (e.keyCode === CONSTANTS.KEYCODES.SPACE_KEY && targetTagName !== "button") {
       e.preventDefault();
       this.controller.togglePlayPause();
     }
--- a/js/components/utils.js
+++ b/js/components/utils.js
@@ -305,6 +305,53 @@ var Utils = {
     return isFinite(Number(property)) ? Number(property) : 0;
   },
 
+
+  /**
+   * Determines whether an element contains a class or not.
+   * TODO:
+   * classList.contains is much better for this purpose, but our current version
+   * of React Test Utils generates events with a null classList, which results in
+   * broken unit tests.
+   *
+   * @param {DOMElement} element The DOM element which we want to check
+   * @param {String} className The name of the class we want to match
+   * @return {Boolean} True if the element contains the given class, false otherwise
+   */
+  hasClass: function(element, className) {
+    if (!element) {
+      return false;
+    }
+    return (' ' + element.className + ' ').indexOf(' ' + className + ' ') > -1;
+  },
+
+  /**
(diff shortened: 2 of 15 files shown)


js/components/utils.js
```diff
@@ -305,6 +305,53 @@ var Utils = {
...
+   *
+   * @param {DOMElement} element The DOM element which we want to check
+   * @param {String} className The name of the class we want to match
+   * @return {Boolean} True if the element contains the given class, false otherwise
+   */
+  elementHasClass: function(element, className) {
+    if (!element) {
+      return false;
+    }
+    return (' ' + element.className + ' ').indexOf(' ' + className + ' ') > -1;
```
jpfranco: Original comment: "What is the purpose of the extra spaces (' ') appended and prepended to the classname?"

jpfranco: This is in order to prevent a false positive when a class name contains the string that you're testing. For example, if you have an element with a class `fancyclass`, then doing `hasClass('fancy')` should return false

js/components/utils.js
```diff
@@ -314,9 +361,7 @@ var Utils = {
   highlight: function(target, opacity, color) {
     target.style.opacity = opacity;
     target.style.color = color;
-    target.style.WebkitFilter = "drop-shadow(0px 0px 3px rgba(255,255,255,0.8))";
-    target.style.filter = "drop-shadow(0px 0px 3px rgba(255,255,255,0.8))";
-    target.style.msFilter = "progid:DXImageTransform.Microsoft.Dropshadow(OffX=0, OffY=0, Color='#fff')";
+    target.style.textShadow = "0px 0px 3px rgba(255, 255, 255, 0.5), 0px 0px 6px rgba(255, 255, 255, 0.5), 0px 0px 9px rgba(255, 255, 255, 0.5)";
```
jpfranco: Original comment: "Does this need to match the highlight-base mixin defined in the _mixins.scss file? If so, is there a way to link them together? If we cannot link them together, can we add a comment to help with maintenance in case we need to change this style in the future?"

jpfranco: Yes, ideally this should match the mixin. There is no way to link them as far as I know, so I will add comments to both files

js/components/utils.js
aeng7: Nice!

jpfranco:
@aeng7 @pilievOoyala I added the comments and made a few small changes. Please check when you get a chance.

js/components/accessibilityControls.js
```diff
@@ -23,8 +23,17 @@ AccessibilityControls.prototype = {
     var currentTime;
     var newPlayheadTime;
     var newVolume;
+    var targetTagName;
 
-    if (e.keyCode === CONSTANTS.KEYCODES.SPACE_KEY){
+    if (e.target && typeof e.target.tagName === "string") {
```
pilievOoyala: if target isn't a string, is it still valid?

jpfranco: This `typeof` is mostly just checking for the existence of `tagName`. It really shouldn't be anything other than a string, but I'm checking it this way just to make sure it doesn't crash the `toLowerCase` below if there's a weird browser bug.

js/controller.js
```diff
@@ -240,6 +241,8 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       //initial DOM manipulation
       this.state.mainVideoContainer.addClass('oo-player-container');
       this.state.mainVideoInnerWrapper.addClass('oo-player');
+      this.state.mainVideoInnerWrapper.attr('aria-label', CONSTANTS.ARIA_LABELS.VIDEO_PLAYER);
+      this.state.mainVideoInnerWrapper.attr('tabindex', '-1');
```
pilievOoyala: This is to hide the button highlighting, right? Can you put in a note?

pilievOoyala:
LGTM

PR #760, opened 2017-06-19
Don't show CC icon when it's disabled in the settings

kuu:
This way providers can prevent CC icon from always being displayed in Mac Safari.


aldigun:
We will close this PR for now.

PR #761, opened 2017-06-21
[PLAYER-509a] Change candidate/latest references to production/latest

aeng7:


aeng7:
The results from this PR may not function correctly. Don't merge yet.

aeng7:
We should think of a way to handle these routes. We can edit the build script to modify the routes at build time, but we have an issue when we want to promote a build to prod or prod/latest, because ideally we would only want to copy, not edit then copy.

pilievOoyala:
We probably need to change the build scripts to fix this. Closing this PR.

PR #763, opened 2017-07-15
[PLAYER-1257] Perceived Player Performance: Disabling loading spinner during player load

jpfranco:
[Ticket](https://jira.corp.ooyala.com/browse/PLAYER-1257])

This will essentially remove the loading spinner while the player is loading. The start screen UI will be rendered as soon as possible. If the user clicks on Play before the player is ready, the spinner will be shown at this point and playback will start automatically after `PLAYBACK_READY`

### DIFF
--- a/js/constants/constants.js
+++ b/js/constants/constants.js
@@ -26,6 +26,10 @@ module.exports = {
     ERROR_SCREEN: "errorScreen"
   },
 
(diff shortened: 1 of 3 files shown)


jpfranco:
Wrong branch, closing.

PR #764, opened 2017-07-15
[PLAYER-1257] Perceived Player Performance: Disabling loading spinner during player load

jpfranco:
[Ticket](https://jira.corp.ooyala.com/browse/PLAYER-1257])

This will essentially remove the loading spinner while the player is loading. The start screen UI will be rendered as soon as possible. If the user clicks on Play before the player is ready, the spinner will be shown at this point and playback will start automatically after `PLAYBACK_READY`

### DIFF
--- a/js/constants/constants.js
+++ b/js/constants/constants.js
@@ -26,6 +26,10 @@ module.exports = {
     ERROR_SCREEN: "errorScreen"
   },
 
+  CUSTOM_EVENTS: {
+    INITIAL_PLAY_REQUESTED: 'html5SkinInitialPlayRequested'
+  },
+
   SKIN_TEXT: {
     LEARN_MORE: "Learn More",
     CLOSED_CAPTION_PREVIEW: "CLOSED CAPTION PREVIEW",
--- a/tests/controller-test.js
+++ b/tests/controller-test.js
@@ -180,6 +180,7 @@ OO = {
       exitFullWindow: function() {},
       exitFullWindowOnEscKey: function() {},
       onBuffered: function() {},
+      onInitialPlayRequested: function() {},
       unsubscribeBasicPlaybackEvents: function() {},
       resetUpNextInfo: function(a) {},
       showUpNextScreenWhenReady: function(a,b) {},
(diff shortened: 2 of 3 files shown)


js/constants/constants.js
```diff
@@ -26,6 +26,10 @@ module.exports = {
     ERROR_SCREEN: "errorScreen"
   },
 
+  CUSTOM_EVENTS: {
```
pilievOoyala: Are you using Custom_Events to hide events from being public? I think the custom event being named the same thing as the regular event may be confusing.  Should we name it DEFFERED_INITIAL_PLAY_REQUESTED?

tests/controller-test.js
```diff
@@ -554,6 +555,43 @@ OO = {
...
+      });
+
+      it('should raise INITIAL_PLAY_REQUESTED event when toggling play/pause on Start Screen', function() {
+        Html5Skin.onPlayerCreated.call(controllerMock, 'customerUi', 'elementId', {});
+        Html5Skin.togglePlayPause.call(controllerMock);
+        expect(messageBusSpy.calledOnce).toBe(true);
+        expect(messageBusSpy.calledWith(CONSTANTS.CUSTOM_EVENTS.INITIAL_PLAY_REQUESTED)).toBe(true);
+      });
+
+      it('should publish INITIAL_PLAY after INITIAL_PLAY_REQUESTED', function() {
```
aeng7: Can we add a unit test to test the dependency of INITIAL_PLAY on PLAYBACK_READY via INITIAL_PLAY_REQUESTED?

js/constants/constants.js
jpfranco: I used custom events in order to avoid adding a dependent of `INITIAL_PLAY`, in case this caused any issues. The regular event's name is `INITIAL_PLAY` though, do you want to change it to `DEFERRED_INITIAL_PLAY`?

tests/controller-test.js
jpfranco: I though about adding this, but there's no easy way of doing it with the current unit test setup. This would be like testing the message bus though, so this should technically be covered in mjolnir already.

aeng7: Alright, sounds good to me.

aeng7:
LGTM

js/constants/constants.js
jpfranco: But yeah, I'm also using custom events so that they're private. This new event is internal to the skin so the customer shouldn't be using it.

pilievOoyala: I'm cool with the custom events.  @aeng7 Which var name do think is better INITIAL_PLAY_REQUESTED or DEFERED_INITIAL_PLAY?

aeng7: I think INITIAL_PLAY_REQUESTED is fine. This gets triggered when the user touches/clicks the screen, which is a user request for initial playback.

PR #766, opened 2017-07-21
[PLAYER-1257]  Hiding everything but thumbnail until PLAYBACK_READY

jpfranco:
We will still be removing the spinner during startup, but instead of allowing the user to click on Play before `PLAYBACK_READY`, we will only show the thumbnail until the player is ready to play.


aeng7:
LGTM

PR #768, opened 2017-07-25
add z-index to overlay container when overlay showing and div not empty

ocundale:


pilievOoyala:
Apologies for the late response. But do you have a test case where we can see the original problem and how this fixes it?

tr-coull:
Thanks for the reply. We are currently testing a potential temporary workaround for the issue in our VPAID code which changes the z-index of the _.oo-player-skin-plugins.oo-showing_ div to be 12005 to move forward while this issue is open, but it is not a sustainable solution we can maintain. 

However, this has revealed another separate issue with this test player (related to the Google IMA) which means the nonlinear advert will only show for a split second before being hidden. We have been in touch with Steve Paddon at Ooyala separately about this and are awaiting further information on a solution.

If you look at the test player _https://featured.dugout.com/preroll-test/coull.html_ you will see the fix to change the z-index of the previously-obstructed div running, but because of the other GIMA issue, it will then be removed quickly. 

To see the original issue again, you could use a local proxy to re-write the param **pid** to be **10989** instead of **13675** on requests to **quasar.coull.com**. The z-index workaround we released will then not run so you will see the original z-index issue occurring, and also the GIMA issue which causes the nonlinear ad and VPAID code to be removed from the page. If you are unable to test with a re-write, let us know and we will try to restore a test page highlighting the original issue without our temporary workaround again.

aldigun:
We will close this PR for now.

PR #769, opened 2017-07-25
Tooltips integration

LuisGerardoAngel:
Feature required by ManU
Jira ticket FR-101

### DIFF
--- a/js/components/controlBar.js
+++ b/js/components/controlBar.js
@@ -335,7 +335,7 @@ var ControlBar = React.createClass({
         aria-label={playPauseAriaLabel}
         autoFocus={this.props.controller.state.playPauseButtonFocused}>
         <Icon {...this.props} icon={playIcon} style={dynamicStyles.iconCharacter} 
-          className="oo-tooltip oo-tooltip-right" dataTip={playPauseTooltip}/>
+          className="oo-tooltip oo-tooltip-right" tooltip={playPauseTooltip}/>
       </button>,
 
       "live": <a className={liveClass}
@@ -355,7 +355,7 @@ var ControlBar = React.createClass({
           aria-label={volumeAriaLabel}>
           <Icon {...this.props} icon={volumeIcon} ref="volumeIcon"
             style={this.props.skinConfig.controlBar.iconStyle.inactive}
-            className="oo-tooltip oo-tooltip-center" dataTip={volumeTooltip}/>
+            className="oo-tooltip oo-tooltip-center" tooltip={volumeTooltip}/>
         </button>
         {volumeControls}
       </div>,
@@ -370,7 +370,7 @@ var ControlBar = React.createClass({
         onClick={this.handleMoreOptionsClick} key="moreOptions" aria-hidden="true">
         <Icon {...this.props} icon="ellipsis" style={dynamicStyles.iconCharacter}
           onMouseOver={this.highlight} onMouseOut={this.removeHighlight}
-          className="oo-tooltip oo-tooltip-center" dataTip={CONSTANTS.TOOLTIPS.VQ}/>
+          className="oo-tooltip oo-tooltip-center" tooltip={CONSTANTS.TOOLTIPS.MORE_OPTIONS}/>
       </a>,
 
       "quality": (
@@ -379,7 +379,7 @@ var ControlBar = React.createClass({
           <a className={qualityClass} onClick={this.handleQualityClick} style={selectedStyle} aria-hidden="true">
             <Icon {...this.props} icon="quality" style={dynamicStyles.iconCharacter}
               onMouseOver={this.highlight} onMouseOut={this.removeHighlight}
-              className="oo-tooltip oo-tooltip-center" dataTip={CONSTANTS.TOOLTIPS.VQ}/>
+              className="oo-tooltip oo-tooltip-center" tooltip={CONSTANTS.TOOLTIPS.VIDEO_QUALITY}/>
           </a>
         </div>
       ),
@@ -388,7 +388,7 @@ var ControlBar = React.createClass({
         onClick={this.handleDiscoveryClick} key="discovery" aria-hidden="true">
         <Icon {...this.props} icon="discovery" style={dynamicStyles.iconCharacter}
           onMouseOver={this.highlight} onMouseOut={this.removeHighlight}
-          className="oo-tooltip oo-tooltip-center" dataTip={CONSTANTS.TOOLTIPS.DISCOVER}/>
+          className="oo-tooltip oo-tooltip-center" tooltip={CONSTANTS.TOOLTIPS.DISCOVER}/>
       </a>,
 
       "closedCaption": (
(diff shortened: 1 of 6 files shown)


scss/components/_tooltips.scss
```diff
@@ -0,0 +1,77 @@
+.oo-tooltip::before {
+  content: attr(data-tip) ;
+  font-family: "Roboto Condensed", sans-serif;
+  text-shadow: none;
+  position:absolute;
+  z-index: 99999999;
```
jpfranco: We should use a `$zindex` variable from _variables.scss here. We should also avoid using an arbitrarily large z-index since the customer might struggle to get their own popup menus, etc. on top of the player elements. I recommend testing with the value of `$zindex-popover`, if this is displayed correctly with ads then I think we could use that value.

scss/components/_tooltips.scss
```diff
@@ -0,0 +1,77 @@
...
+
+.oo-tooltip.oo-tooltip-right::before {
+  left: 0%;
+  right: inherit;
+}
+
+.oo-tooltip::after {
+  content: "";
+  position: absolute;
+  z-index: 99999999;
```
jpfranco: Same comment as above, we should switch to a variable.

js/constants/constants.js
```diff
@@ -135,6 +135,21 @@ module.exports = {
...
+  TOOLTIPS: {
+    PLAY: "Play",
+    PAUSE: "Pause",
+    REPLAY: "Replay",
+    MUTE: "Mute",
+    UNMUTE: "Unmute",
+    SHARE: "Share",
+    DISCOVER: "Discover",
+    CC: "Closed Captions",
+    VQ: "Video Quality",
```
jpfranco: Can we use the full names here such as `VIDEO_QUALITY`? I think it will make it easier to understand when referencing outside this file.

jpfranco:
@LuisGerardoAngel please check to see if you can add unit tests similar to those added for ARIA labels in controlBar-test.js.

jpfranco:
Found a couple of issues:

- The tooltips don't seem to work on Firefox and IE11 on elements that are buttons (it works with the anchor tags)
- The "show more" button that is shown when the player is small shows the wrong label. Also, the icons on the "show more" menu don't have tooltips, although I'm not sure that this is part of the requirements.
screen shot 2017-07-25 at 12 47 08 pm
screen shot 2017-07-25 at 12 47 16 pm

scss/components/_tooltips.scss
```diff
@@ -0,0 +1,77 @@
...
+@keyframes OOTooltip {
+  0% {opacity: 0;}
+  25% {opacity: 0;}
+  50% {opacity: 0;}
+  100% {opacity: 0.8;}
+}
+
+.oo-tooltip:hover::after {
+  opacity: 0.8;
+  animation: OOTooltip 2s;
```
jpfranco: I think that this transition might be a bit too slow. The first time I tried it I though the tooltips weren't working. I suggest reviewing with someone from UX.

scss/components/_tooltips.scss
```diff
@@ -0,0 +1,77 @@
...
+  text-shadow: none;
+  position:absolute;
+  z-index: 99999999;
+  background: black;
+  color: white;
+  border-radius: 3px;
+  opacity: 0;
+  font-size: 15px;
+  padding: 8px 20px 8px 20px;
+  bottom: 59px;
```
jpfranco: Is it possible to make this a function of `$control-bar-height` so that the offset doesn't break if we change the style of the control bar? A fixed value will also make it harder for customers who have custom skin styles.

js/components/controlBar.js
```diff
@@ -403,7 +417,8 @@ var ControlBar = React.createClass({
         key="fullscreen"
         tabIndex="0"
         aria-label={fullscreenAriaLabel}>
-        <Icon {...this.props} icon={fullscreenIcon} style={dynamicStyles.iconCharacter} />
+        <Icon {...this.props} icon={fullscreenIcon} style={dynamicStyles.iconCharacter}
+          className="oo-tooltip oo-tooltip-left" dataTip={fullscreenTooltip}/>
```
jpfranco: `oo-tooltip-left` should be dynamic based on whether the button is the last element in the control bar. The user can customize the elements in the control bar via the skin.json, so fullscreen will not always be the last element.

js/components/icon.js
```diff
@@ -25,14 +25,16 @@ Icon.propTypes = {
   icon: React.PropTypes.string,
   skinConfig: React.PropTypes.object,
   className: React.PropTypes.string,
-  style: React.PropTypes.object
+  style: React.PropTypes.object,
+  dataTip: React.PropTypes.string
```
jpfranco: Would this be clearer if we named it `toolTip`? I realize that `data` is needed for the attribute, but the parameter would probably be better understood as `toolTip`.

LuisGerardoAngel:
Thanks for the response. I'll check it out and let you know the changes.

scss/components/_tooltips.scss
LuisGerardoAngel: The requirement was a 1 second delay before display the tooltip but I will ask again, it seems wrong to me too.

scss/components/_tooltips.scss
LuisGerardoAngel: Fix

scss/components/_tooltips.scss
LuisGerardoAngel: Fix

js/constants/constants.js
LuisGerardoAngel: Fix

scss/components/_tooltips.scss
LuisGerardoAngel: Fix

js/components/icon.js
LuisGerardoAngel: Fix

scss/components/_tooltips.scss
```diff
@@ -0,0 +1,88 @@
...
+}
+
+.oo-tooltip.oo-tooltip-left::before {
+  left: inherit;
+  right: 0%;
+}
+
+.oo-tooltip.oo-tooltip-center::before {
+  left: 50%;
+  transform: translate(-50%);
```
jpfranco: Can we use the `vendor-prefixes` mixin here? Not actually needed for current browsers, but someone might be using an old version of Safari

scss/components/_tooltips.scss
```diff
@@ -0,0 +1,88 @@
...
+
+//tooltip
+$tooltip-opacity: 0.75;
+$tooltip-color: black;
+$tooltip-animation-duration: 0.5s;
+$tooltip-pointer-size: 5px;
+
+.oo-tooltip::before, .oo-tooltip::after{
+  position: absolute;
+  z-index: $zindex-popover;
```
jpfranco: I would recommend defining a `zindex-tooltip` var with the same value as `zindex-popover` so that the two are not tied together. Did this work fine with ads btw?

scss/components/_tooltips.scss
```diff
@@ -0,0 +1,88 @@
...
+  font-family: $font-family-roboto-condensed;
+  text-shadow: none;
+  background: $tooltip-color;
+  border-radius: 3px;
+  font-size: 15px;
+  padding: 8px 20px 8px 20px;
+  bottom: $tooltip-offset + $tooltip-pointer-size;
+}
+
+.oo-xsmall .oo-tooltip::before {
```
jpfranco: It looks like we have all the `oo-small`, etc. styles in a section of the corresponding size file. I'm not a big fan of this approach, but it might make sense to move these styles there for consistency.

jpfranco:
Another small issue I found, the tooltips will trigger if you hover over the scrubber bar (notice the mouse position):

screen shot 2017-07-27 at 6 04 27 pm

pilievOoyala:
@LuisGerardoAngel We have two checkins that are supposed to do the same thing (at least it seems like that to us), who's is supposed to go in for ManU?

 https://github.com/ooyala/html5-skin/pull/769 and https://github.com/ooyala/html5-skin/pull/772

PR #770, opened 2017-07-25
[PLAYER-1016] - Accessibility features

alfrejivi:
tab invoke control bar, highlight class on focus updated, "G" value descriptor for pause state addressed for screen readers. 
task details on this document: [PLAYER-1016] [https://docs.google.com/document/d/15XO8gCRKZSYwWBKmd1tsxGYO0fBnhnt_-eIyoRQin6I/edit ](url)

### DIFF
--- a/js/views/endScreen.js
+++ b/js/views/endScreen.js
@@ -35,7 +35,7 @@ var EndScreen = React.createClass({
   },
 
   blurOnMouseUp: function(evt) {
-    if (evt.currentTarget) {
+    if (evt.currentTarget && evt.currentTarget.blur) {
       evt.currentTarget.blur();
     }
   },
--- a/js/views/endScreen.js
+++ b/js/views/endScreen.js
@@ -34,6 +34,12 @@ var EndScreen = React.createClass({
     }
   },
 
+  blurOnMouseUp: function(evt) {
+    if (evt.currentTarget) {
+      evt.currentTarget.blur();
+    }
+  },
+
   handleClick: function(event) {
     // pause or play the video if the skin is clicked
     event.preventDefault();
@@ -100,6 +106,7 @@ var EndScreen = React.createClass({
 
         <button className={actionIconClass}
           onClick={this.handleClick}
+          onMouseUp={this.blurOnMouseUp}
           tabIndex="0"
           aria-label={CONSTANTS.ARIA_LABELS.REPLAY}>
           <Icon {...this.props} icon="replay" style={actionIconStyle}/>
--- a/js/views/pauseScreen.js
+++ b/js/views/pauseScreen.js
@@ -119,9 +119,9 @@ var PauseScreen = React.createClass({
           {this.props.skinConfig.pauseScreen.showDescription ? descriptionMetadata : null}
         </div>
 
-        <a className={actionIconClass} onClick={this.handleClick}>
+        <button className={actionIconClass} onClick={this.handleClick} aria-hidden="true">
           <Icon {...this.props} icon="pause" style={actionIconStyle}/>
-        </a>
+        </button>
 
         <div className="oo-interactive-container">
           {this.props.closedCaptionOptions.enabled ?
--- a/js/views/pauseScreen.js
+++ b/js/views/pauseScreen.js
@@ -119,7 +119,7 @@ var PauseScreen = React.createClass({
           {this.props.skinConfig.pauseScreen.showDescription ? descriptionMetadata : null}
         </div>
 
-        <button className={actionIconClass} onClick={this.handleClick} aria-hidden="true">
+        <button className={actionIconClass} onClick={this.handleClick} aria-hidden="true" tabIndex="-1">
           <Icon {...this.props} icon="pause" style={actionIconStyle}/>
         </button>
 
--- a/js/views/playingScreen.js
+++ b/js/views/playingScreen.js
@@ -73,6 +73,10 @@ var PlayingScreen = React.createClass({
     // show control bar on tab key navigation
     if(event.which === 9 || event.keyCode === 9) {
       this.showControlBar();
+      this.props.controller.startHideControlBarTimer();
+      if(this.props.controller.showControlBarByTab) {
+        this.props.controller.showControlBarByTab();
+      }
     }
   },
 
--- a/js/views/playingScreen.js
+++ b/js/views/playingScreen.js
@@ -71,12 +71,9 @@ var PlayingScreen = React.createClass({
 
   handleKeyPress: function(event) {
     // show control bar on tab key navigation
-    if(event.which === 9 || event.keyCode === 9) {
+    if (event.which === 9 || event.keyCode === 9) {
       this.showControlBar();
       this.props.controller.startHideControlBarTimer();
-      if(this.props.controller.showControlBarByTab) {
-        this.props.controller.showControlBarByTab();
-      }
     }
   },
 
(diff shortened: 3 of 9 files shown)


js/views/playingScreen.js
```diff
@@ -69,6 +69,13 @@ var PlayingScreen = React.createClass({
     // for mobile, touch is handled in handleTouchEnd
   },
 
+  handleKeyPress: function(event){
```
jpfranco: Can we add a unit test for this scenario?

js/views/playingScreen.js
```diff
@@ -69,6 +69,13 @@ var PlayingScreen = React.createClass({
...
+    // show control bar on tab key navegation
```
jpfranco: typo in navegation :p

js/views/playingScreen.js
```diff
@@ -69,6 +69,13 @@ var PlayingScreen = React.createClass({
...
+    if(event.which === 9 || event.keyCode === 9){
```
jpfranco: Small nitpick, can we add spaces before and after the parenthesis? Same for the function declaration above

scss/components/_control-bar.scss
```diff
@@ -81,6 +81,13 @@
 
     .oo-control-bar-item:focus .oo-icon {
       @include highlight-base();
+      @include highlight-tab();
+      opacity: 1 !important;
+    }
+
+    .oo-control-bar-item:hover .oo-icon {
+      @include highlight-base();
+      @include remove-highlight-tab();
```
jpfranco: Can we keep the focus styles when hovering? I think it's more accurate since hovering doesn't remove the focused state.

scss/mixins/_mixins.scss
```diff
@@ -13,6 +13,14 @@
   }
 }
 
+@mixin highlight-tab(){
+  box-shadow: inset 0 0 0 2px $blue-light;
+}
+
+@mixin remove-highlight-tab(){
```
jpfranco: Can we add spaces after the parenthesis for consistency?

jpfranco:
@alfrejivi I might've been the one to suggest this initially, but it looks like there is no way to hide the control bar without using the mouse after focusing with the tab key. This might be an issue for people who can see properly but use the keyboard to navigate due to having motor disabilities. We should review this aspect of the implementation with product.

scss/mixins/_mixins.scss
```diff
@@ -13,6 +13,14 @@
   }
 }
 
+@mixin highlight-tab(){
+  box-shadow: inset 0 0 0 2px $blue-light;
```
jpfranco: Should we remove the `inset` part so that the highlight is larger? The customer complained about the visibility of the focused style, so I think it's better to make it bigger.

scss/components/_control-bar.scss
jpfranco: I think it would probably be better to also remove the glow on `focus` and keep just the box shadow. This way when you hover over a focused element you will see both the focused and hover styles.

js/views/pauseScreen.js
```diff
@@ -119,9 +119,9 @@ var PauseScreen = React.createClass({
           {this.props.skinConfig.pauseScreen.showDescription ? descriptionMetadata : null}
         </div>
 
-        <a className={actionIconClass} onClick={this.handleClick}>
+        <button className={actionIconClass} onClick={this.handleClick} aria-hidden="true">
```
jpfranco: This needs a `tabindex="-1"` so that it doesn't get focus with a `tab`.

scss/components/_control-bar.scss
alfrejivi: I added the active state cause when you clicked on an icon the box-shadow appeared, then in the active state, I remove the box shadow.

jpfranco:
@alfrejivi the updates look good, just found a minor issue:

- On iPad the hover style gets stuck on the play/pause button after clicking on it. It looks like this wasn't the case before.

The remaining updates are:
- Unit tests
- Allow the control bar to auto-hide after focusing on a button with the `tab` key. Ideally this should use the same mechanism we use for auto-hide after hover.

tests/views/playingScreen-test.js
```diff
@@ -104,6 +106,9 @@ describe('PlayingScreen', function () {
 
     TestUtils.Simulate.mouseOver(screen);
     expect(over).toBe(true);
+
+    TestUtils.Simulate.keyUp(screen, {key: 'Tab', which: 9, keyCode: 9});
```
jpfranco: We should keep this as a separate unit test. Please refer to startScreen-tests.js for an example of how we want to structure unit tests moving forward. Unfortunately right now our tests in this project are vey inconsistent.

I suggest a test case like "should show control bar when pressing the tab key". This would include the simulated key press and then it would check that `startHideControlBarTimer` and `showControlBarByTab`

jpfranco: There are some examples of spying on callbacks on the 'Controller testing Anamorphic videos fix' in controller-test.js

js/views/playingScreen.js
```diff
@@ -69,6 +69,17 @@ var PlayingScreen = React.createClass({
     // for mobile, touch is handled in handleTouchEnd
   },
 
+  handleKeyPress: function(event) {
+    // show control bar on tab key navigation
+    if(event.which === 9 || event.keyCode === 9) {
```
jpfranco: Small nitpick, we should add spaces before the `(` to keep the style consistent

js/views/playingScreen.js
```diff
@@ -69,6 +69,17 @@ var PlayingScreen = React.createClass({
...
+      this.showControlBar();
+      this.props.controller.startHideControlBarTimer();
+      if(this.props.controller.showControlBarByTab) {
```
jpfranco: Is this for unit tests, or are there cases in which this function is not defined?

jpfranco:
It looks like the changes from controller.js are not showing in the diff. There might be some git issue.

scss/components/_state-screen.scss
```diff
@@ -129,11 +129,19 @@
     }
   }
 
-  .oo-action-icon:focus {
+  .oo-action-icon:focus {    
+    @include highlight-tab();
+  }
+
+  .oo-action-icon:hover {
```
jpfranco: I think it might be better to remove this hover style since it wasn't there on the previous version.

jpfranco:
@alfrejivi thanks for the updates, the PR looks good. Only had one comment and also found one last minor issue. The keyboard highlight is shown when clicking on the big replay button. You might be able to fix  it by using the `blurOnMouseUp` login from `controlBar.js`

alfrejivi:
@jpfranco which browser did u see last minor issue? I can't replicate it.

jpfranco:
@alfrejivi on Chrome. Try testing with the rest of the scripts pointed to candidate/latest.

js/views/endScreen.js
```diff
@@ -34,6 +34,12 @@ var EndScreen = React.createClass({
     }
   },
 
+  blurOnMouseUp: function(evt) {
+    if (evt.currentTarget) {
+      evt.currentTarget.blur();
```
pilievOoyala: Do we need to check if the blur function exists or is it guaranteed to be there?

jpfranco:
@alfrejivi the updates look good to me. Usually we require two approvals for each PR, so I'll ask someone else to review. We will not be merging this yet because we want to get the current release first, but once this is approved we can sync up to create a sandbox build so that QA can start testing.

jpfranco:
Closing this in favor of https://github.com/ooyala/html5-skin/pull/771

PR #773, opened 2017-08-09
PLAYER-1389 - prevent stuck on hover styles on mobile devices

alfrejivi:
Solution to the PLAYER-1389 ticket.


pilievOoyala:
@jpfranco  Can you take a quick look and see if you are okay with it too. So we have two approvals.

PR #776, opened 2017-08-17
Feat/tooltips v2

FernandoCuevas:


js/components/tooltip.js
```diff
@@ -0,0 +1,135 @@
...
+  },
+
+  onMouseLeave: function () {
+    this.setState({ visible: false });
+  }
+});
+
+Tooltip.propTypes = {
+  enabled: React.PropTypes.bool.isRequired,
+  text: React.PropTypes.string.isRequired,
```
aeng7: Should we use include a default for text below in defaultProps since it is required?

js/components/tooltip.js
```diff
@@ -0,0 +1,135 @@
...
+    left: pointerAlignment[alignment]
+  };
+}
+
+
+var Tooltip = React.createClass({
+
+  componentDidMount: function () {
+    this.onMouseOver = this.onMouseOver.bind(this);
+    this.onMouseLeave = this.onMouseLeave.bind(this);
```
aeng7: Can we bind these functions when they are defined instead?

js/components/tooltip.js
```diff
@@ -0,0 +1,135 @@
...
+
+    this.parentElement = (ReactDOM.findDOMNode(this) || {}).parentElement;
```
aeng7: Do we need the second half where we get the parentElement of an empty object? If the first part fails to find the DOM node, it should be falsy and the if statement in the next line will fail.

scss/components/_control-bar.scss
```diff
@@ -97,6 +97,11 @@
       cursor: auto;
     }
 
+    a.oo-control-bar-item
+    {
```
aeng7: I think the styling for this file is that the { should be on the line above.

js/components/tooltip.js
jpfranco: I added a fix for this, if it's falsy it will return the the object

aeng7:
Can we add some unit tests?

js/components/tooltip.js
jpfranco: Not sure we can do that since they are defined in the object literal.

js/components/tooltip.js
jpfranco: Done

js/components/controlBar.js
```diff
@@ -317,103 +319,146 @@ var ControlBar = React.createClass({
...
-            style={this.props.skinConfig.controlBar.iconStyle.inactive} />
+          aria-label={playPauseAriaLabel}
+          autoFocus={this.props.controller.state.playPauseButtonFocused}>
+          <Icon {...this.props} icon={playIcon} style={dynamicStyles.iconCharacter} />
+          <Tooltip enabled={isTooltipEnabled}
+            alignment={alignment}
+            responsivenessMultiplier={this.responsiveUIMultiple}
+            bottom={this.responsiveUIMultiple * this.props.skinConfig.controlBar.height}
+            text={this.props.playerState == CONSTANTS.STATE.PLAYING ? Utils.getLocalizedString(this.props.language, CONSTANTS.SKIN_TEXT.PAUSE, this.props.localizableStrings) :
+              this.props.playerState == CONSTANTS.STATE.END ? Utils.getLocalizedString(this.props.language, CONSTANTS.SKIN_TEXT.REPLAY, this.props.localizableStrings) : Utils.getLocalizedString(this.props.language, CONSTANTS.SKIN_TEXT.PLAY, this.props.localizableStrings)}>
```
aeng7: Do we need to consider the ERROR playerState?

js/components/controlBar.js
```diff
@@ -317,103 +319,146 @@ var ControlBar = React.createClass({
...
+      }).bind(this),
+
+      "quality": (function (alignment) {
+        return <div className="oo-popover-button-container" key="quality">
           {videoQualityPopover}
           <a className={qualityClass} onClick={this.handleQualityClick} style={selectedStyle} aria-hidden="true">
             <Icon {...this.props} icon="quality" style={dynamicStyles.iconCharacter}
-              onMouseOver={this.highlight} onMouseOut={this.removeHighlight}/>
+              onMouseOver={this.highlight} onMouseOut={this.removeHighlight} />
+            <Tooltip enabled={isTooltipEnabled} text={Utils.getLocalizedString(this.props.language, CONSTANTS.SKIN_TEXT.VIDEO_QUALITY, this.props.localizableStrings)} bottom={this.responsiveUIMultiple * this.props.skinConfig.controlBar.height} alignment={alignment}
```
aeng7: Minor thing, but we should standardize an order for the tooltip properties. Each of these buttons have the tooltip properties in a different order.

PR #791, opened 2017-09-04
Player 711

gramcha:


aeng7:
@gramcha can we add some unit tests?

gramcha:
@aeng7  I have added unit tests. Please check it out.

pilievOoyala:
Approved but need one more approval before merging.

js/controller.js
```diff
@@ -1452,14 +1452,14 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
     onChangeClosedCaptionLanguage: function(event, language) {
       var availableLanguages = this.state.closedCaptionOptions.availableLanguages;
-
       //validate language is available before update and save
-      if (language && availableLanguages && _.contains(availableLanguages.languages, language)) {
+        if (language && availableLanguages && (_.contains(availableLanguages.languages, language) || language === OO.CONSTANTS.CLOSED_CAPTIONS.NONE)) {
         this.state.closedCaptionOptions.language = this.state.persistentSettings.closedCaptionOptions.language = language;
-        var captionLanguage = this.state.closedCaptionOptions.enabled ? language : "";
-        var mode = this.state.closedCaptionOptions.enabled ? OO.CONSTANTS.CLOSED_CAPTIONS.HIDDEN : OO.CONSTANTS.CLOSED_CAPTIONS.DISABLED;
+        var captionLanguage = this.state.closedCaptionOptions.enabled && language !== OO.CONSTANTS.CLOSED_CAPTIONS.NONE ? language : "";
```
pilievOoyala: In both of these cases, what is the value of captionLanguage and mode if the first variable is false? Is it a valid value?

js/controller.js
```diff
@@ -1452,14 +1452,14 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
 
     onChangeClosedCaptionLanguage: function(event, language) {
       var availableLanguages = this.state.closedCaptionOptions.availableLanguages;
-
       //validate language is available before update and save
-      if (language && availableLanguages && _.contains(availableLanguages.languages, language)) {
+        if (language && availableLanguages && (_.contains(availableLanguages.languages, language) || language === OO.CONSTANTS.CLOSED_CAPTIONS.NONE)) {
```
aeng7: We'll need to go through the html5-common update process to update OO.CONSTANTS.CLOSED_CAPTIONS with the NONE value.

aeng7:
@gramcha We're going to revert this PR. We require two approvals to merge, and we typically don't self merge. Since we can't enforce that on github, we're going to make some changes to our approval process on github so that this doesn't happen again. There are also some unresolved comments, including a significant one:

"Can we define this constant (OO.CONSTANTS.CLOSED_CAPTIONS.NONE) elsewhere? It does not fit with the other values in OO.CONSTANTS.CLOSED_CAPTIONS, since we are comparing language against it. This will also help us avoid having to update html5-common, which is a complicated process."

Please reopen this PR and address the open comments. Thanks!

gramcha:
Sure @aeng7 I will reopen.

js/controller.js
gramcha: @pilievOoyala 
captionLanguage will have empty string if language === none.

PR #800, opened 2017-09-06
[PLAYER-1598] Accessibility support for scrubber bar and volume controls

jpfranco:
[Ticket](https://jira.corp.ooyala.com/browse/PLAYER-1598])

Enables keyboard navigation and adds ARIA attributes for the scrubber and volume bars.

**I will create a separate PR for remaining fixes and unit tests**


js/components/accessibilityControls.js
```diff
@@ -1,87 +1,126 @@
...
+    var sliderIsActive = document.activeElement && document.activeElement.getAttribute('role') === 'slider';
 
-    // We override the default behavior when the target element is a button (pressing
-    // the spacebar on a button should activate it).
-    // Note that this is not a comprehensive fix for all clickable elements, this is
-    // mostly meant to enable keyboard navigation on control bar elements.
-    if (e.keyCode === CONSTANTS.KEYCODES.SPACE_KEY && targetTagName !== "button") {
-      e.preventDefault();
-      this.controller.togglePlayPause();
+    switch (e.keyCode) {
```
aeng7: Nice job consolidating all these if else statements into a switch statement.

js/components/controlBar.js
```diff
@@ -48,6 +46,28 @@ var ControlBar = React.createClass({
...
+    if (!this.props.controller.state.focusedControl || !this.domNode) {
+      return;
+    }
+    var control = this.domNode.querySelector('[data-focus-id="' + this.props.controller.state.focusedControl + '"]');
+
+    if (control && typeof control.focus === 'function') {
+      control.focus();
+      // If we got to this point it means that play was triggered using the spacebar
+      // (since a click would've cleared the focused element) and we need to
+      // trigger control bar auto hide
```
aeng7: Is there a way to make the spacebar flow and the click flow the same?

jpfranco: Not in this case since the control bar timer is not started on click. When interacting with a mouse the timer is actually triggered on mouse leave. For keyboard controls all the interactions should auto show the controls.

js/components/utils.js
```diff
@@ -6,6 +6,127 @@
...
+      var isLiveNow = Math.abs(timeShift) < 1;
+
+      if (useNegativeDvrOffset) {
+        // We don't show current time unless there is a time shift when using
+        // negative DVR offset
+        currentTime = isLiveNow ? null : Utils.formatSeconds(timeShift);
+      } else {
+        // When not using negative DVR offset, DVR progress is shown in the usual
+        // "current time of total time" format, with total time set to the size of DVR window
+        currentTime = isLiveNow ? totalTime : Utils.formatSeconds(Math.ceil(duration + timeShift));
```
aeng7: Should we also check that total time is not null?

js/components/volumeControls.js
```diff
@@ -0,0 +1,195 @@
...
+var ClassNames = require('classnames');
+var Slider = require('./slider');
+var Utils = require('./utils');
+var CONSTANTS = require('../constants/constants');
+
+var VolumeControls = React.createClass({
+
+  handleVolumeClick: function(event) {
+    event.preventDefault();
+    var newVolume = parseFloat(event.target.dataset.volume);
```
aeng7: Do we need a null check for event.target.dataset?

js/components/volumeControls.js
```diff
@@ -0,0 +1,195 @@
...
+      var volumeClass = ClassNames({
+        'oo-volume-bar': true,
+        'oo-on': turnedOn
+      });
+      var barStyle = {
+        backgroundColor: this.props.skinConfig.controlBar.volumeControl.color ? this.props.skinConfig.controlBar.volumeControl.color : this.props.skinConfig.general.accentColor
+      };
+
+      volumeBars.push(
+        <a data-volume={(i + 1) / 10}
```
aeng7: Can we store (i+1)/10 in a var and use it here and above for the turnedOn var?

js/constants/constants.js
```diff
@@ -109,6 +109,17 @@ module.exports = {
     UNMUTE: "Unmute",
     FULLSCREEN: "Fullscreen",
     EXIT_FULLSCREEN: "Exit Fullscreen",
+    SEEK_SLIDER: "Seek slider",
+    VOLUME_SLIDER: "Volume slider",
+    VOLUME_PERCENT: "{volume}% volume",
+    TIME_DISPLAY: "{currentTime} of {totalTime}",
+    TIME_DISPLAY_LIVE: "Live video",
+    TIME_DISPLAY_DVR: "{currentTime} of {totalTime} live video"
```
aeng7: Can we store these macros (currentTime, totalTime, volume) as constants? If we do, we should update the macro replacement functions to use these constants as well.

js/constants/constants.js
```diff
@@ -120,6 +131,21 @@ module.exports = {
     ESCAPE_KEY: 27
   },
 
+  // KeyboardEvent's which and keyCode properties are deprecated.
+  // It's a good idea to use KeyboardEvent.key moving forward even though React
+  // synthetic events normalize event data.
+  KEY_VALUES: {
+    ENTER: "Enter",
+    TAB: "Tab",
+    SPACE: " ", // yep
```
aeng7: yep

js/controller.js
```diff
@@ -78,7 +78,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       "mainVideoPlayhead": 0,
       "adVideoPlayhead": 0,
       "focusedElement": null,
-      "playPauseButtonFocused": false,
+      "focusedControl": null, // Stores the id of the control bar element that is currently focused
```
aeng7: Is it possible for multiple controls to be focused somehow?

scss/components/_control-bar.scss
```diff
@@ -164,55 +164,6 @@
   }
 }
 
-a.oo-volume-bar {
```
aeng7: Were these styles effectively moved into the VolumeControls component?

js/controller.js
jpfranco: Nope, only one at a time, same as with `document.activeElement`

js/components/controlBar.js
aeng7: Got it.

scss/components/_control-bar.scss
jpfranco: Yes, since I moved the code into its own component I also separated the styles. Some are new though

js/components/volumeControls.js
jpfranco: Will do

js/components/volumeControls.js
jpfranco: I copied this straight from the old code, I think it's a good idea to add the check though

js/components/utils.js
jpfranco: Good catch

js/constants/constants.js
jpfranco: It's a bit complicated to reference other values from the same file since it's defined as an object literal. I'll leave this for another PR since I still have other stuff I need to fix

jpfranco:
@aeng7 implemented some of the suggestions and also added a fix for volume bars on small players.

aeng7:
LGTM

js/components/utils.js
aeng7: Are you addressing this in a future PR?

js/components/accessibilityControls.js
```diff
@@ -1,87 +1,126 @@
...
+        break;
+      default:
+        break;
     }
+  },
+
+  /**
+   * Increases or decreases the player volume by the specified percentage. Values beyond
+   * the minimum or maximum will be constrained to appropriate values.
+   *
```
pilievOoyala: we need the @public or @private

js/components/accessibilityControls.js
```diff
@@ -1,87 +1,126 @@
...
+  },
 
-    else if (e.keyCode === CONSTANTS.KEYCODES.RIGHT_ARROW_KEY || e.keyCode === CONSTANTS.KEYCODES.LEFT_ARROW_KEY){
-      e.preventDefault();
-      var shiftSign = 1; // positive 1 for fast forward, negative for rewind back
+  /**
+   * Seeks the video by the specified number of seconds. The direction of the playhead
+   * can be specified with the forward parameter. If a value exceeds the minimum or maximum
+   * seekable range it will be constrained to appropriate values.
+   *
```
pilievOoyala: need @public or @private

js/components/utils.js
jpfranco: @aeng7 yes I will include this in the next one.

js/components/controlBar.js
```diff
@@ -48,6 +46,28 @@ var ControlBar = React.createClass({
     window.removeEventListener('orientationchange', this.closePopovers);
   },
 
+  /**
+   * Restores the focus of a previously selected control bar item.
+   * This is needed as a workaround because switching between play and pause states
+   * currently causes the control bar to re-render.
```
pilievOoyala: @public or @private

js/components/accessibilityControls.js
jpfranco: Adding...

js/components/controlBar.js
```diff
@@ -203,12 +197,24 @@ var ControlBar = React.createClass({
...
-  changeVolumeSlider: function (event) {
-    var newVolume = parseFloat(event.target.value);
-    this.props.controller.setVolume(newVolume);
-    this.setState({
-      volumeSliderValue: event.target.value
-    });
+  /**
+   * Fires whenever an item is focused inside the control bar. Stores the id of
+   * the focused control.
+   *
```
pilievOoyala: @public or @private

js/components/controlBar.js
jpfranco: Since this is a component I don't think you could call it externally

js/components/accessibilityControls.js
jpfranco: Adding...

js/components/volumeControls.js
```diff
@@ -0,0 +1,207 @@
...
+    return CONSTANTS.ARIA_LABELS.VOLUME_PERCENT.replace('{volume}', this.getVolumePercent());
+  },
+
+  /**
+   * Builds the volume bar controls that are shown on desktop.
+   */
+  renderVolumeBars: function() {
+    var volumeBars = [];
+
+    for (var i = 0; i < 10; i++) {
```
pilievOoyala: Should the 10 be a constant? Do we have variable number of bars?

jpfranco: It's always ten since each one increases by 10%. I copied this from the old code, so I think it's not prepared to handle a variable number of bars.

pilievOoyala: Sounds good

PR #802, opened 2017-09-06
[player-1545] iOS 10 allows inline playback. Skin is now active for i…

pilievOoyala:
…OS 10+, so that if you come out of fullscreen it will be in the correct state. It also fixes ads now showing when playing inline.

### DIFF
--- a/js/components/utils.js
+++ b/js/components/utils.js
@@ -592,8 +592,8 @@ var Utils = {
   },
 
   /**
-   * [description]
-   * @return {[type]} [description]
+   * Returns whether the OS can render the skin
+   * @return {[boolean]} true if the OS can render the skin.
    */
   canRenderSkin: function() {
     var result = !(OO.isIphone && OO.iosMajorVersion < 10);
--- a/js/components/utils.js
+++ b/js/components/utils.js
@@ -362,7 +362,7 @@ var Utils = {
     target.style.opacity = opacity;
     target.style.color = color;
     // HEADSUP
-    // This is currently the same style as the one used in _mixins.scss. 
+    // This is currently the same style as the one used in _mixins.scss.
     // We should change both styles whenever we update this.
     target.style.textShadow = "0px 0px 3px rgba(255, 255, 255, 0.5), 0px 0px 6px rgba(255, 255, 255, 0.5), 0px 0px 9px rgba(255, 255, 255, 0.5)";
   },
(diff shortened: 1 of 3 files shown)


aldigun:
What happen with the skin on iOS10 if playsinline is not activated by the playerParam?

js/components/utils.js
```diff
@@ -591,6 +591,15 @@ var Utils = {
     }
   },
 
+  /**
+   * [description]
+   * @return {[type]} [description]
```
aeng7: [description]

js/controller.js
```diff
@@ -196,8 +196,8 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         }
 
         // ad events
-        if (!Utils.isIPhone()) {
-          //since iPhone is always playing in full screen and not showing our skin, don't need to render skin
+        if (Utils.canRenderSkin()) {
+          //since iPhone < iOS10 is always playing in full screen and not showing our skin, don't need to render skin
```
aeng7: We should probably move this comment to the canRenderSkin function (or have it in both places).

js/controller.js
```diff
@@ -598,8 +598,8 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         this.state.screenToShow = CONSTANTS.SCREEN.END_SCREEN;
         this.mb.publish(OO.EVENTS.END_SCREEN_SHOWN);
       }
-      if (Utils.isIPhone()){
-        //iPhone end screen is the same as start screen, except for the replay button
+      if (!Utils.canRenderSkin()){
```
aeng7: I'm leaning towards having the components decide what to render. This way, the screenToShow states will be consistent with desktop flow (which hopefully means less headaches for this feature), but the components can just choose to not render anything. @pilievOoyala @aldigun @jpfranco your thoughts on this?

aeng7:
LGTM so we can get this in for testing tonight. @pilievOoyala , please take a look at my comments for a future PR. Thanks!

js/controller.js
pilievOoyala: I with you on this for the most part, unless we need to show a completely different screen because it doesn't make sense on that platform.

jpfranco: I agree that the checks are ugly, but it looks like there is a bit more logic than just the rendering, so it might be tricky to migrate it to the components. We should definitely look for a way to refactor this, but for now I think it's safer to leave it as is.

PR #804, opened 2017-09-07
[PLAYER-1598] Scrubber bar and volume bar accessibility fixes

jpfranco:
Fixed a few remaining issues from the previous PR. 

I will make yet another PR for the remaining unit tests.


js/components/utils.js
```diff
@@ -106,19 +105,19 @@ var Utils = {
...
+        currentTime = isLiveNow ? totalTime : this.formatSeconds(Math.ceil(duration + timeShift));
       }
     } else {
       currentTime = currentPlayheadTime;
     }
     // Total time is not displayed when using negative DVR offset, only the
     // timeshift is shown
     if (useNegativeDvrOffset) {
-      totalTime = isLiveStream ? null : totalTime;
+      totalTime = isLiveStream ? '' : totalTime;
```
jpfranco: There is no good way of "fixing" the output when the input parameters are incorrect, so instead of handling all of the edge cases I'm now returning empty strings instead of `null`'s. This way at least the output values can be consistently treated as strings.

js/components/utils.js
```diff
@@ -66,36 +63,38 @@ var Utils = {
...
    * Live - DVR - useNegativeDvrOffset === false
    *  currentTime: Formatted value of the current playhead relative to max time shift
    *  totalTime: Formatted value of the total duration of the DVR window
+   * NOTE:
+   * Either property can be returned as an empty string if the parameters don't match the requirements.
    *
    * @function getTimeDisplayValues
    * @param {Number} currentPlayhead The current value of the playhead in seconds.
-   * @param {Number} duration The total duration of the video in seconds.
+   * @param {Number} duration The total duration of the video in seconds. Should be -0 or Infinity for Live videos with no DVR.
```
pilievOoyala: is -0 actually a thing or did you mean -1?

jpfranco: Yes, this actually what the playhead is set to in these cases. I don't know why it was done like this, but I didn't wan't to risk breaking anything by changing it.

pilievOoyala: Sounds good.

js/components/accessibilityControls.js
```diff
@@ -50,7 +50,7 @@ AccessibilityControls.prototype = {
         break;
       case CONSTANTS.KEYCODES.LEFT_ARROW_KEY:
       case CONSTANTS.KEYCODES.RIGHT_ARROW_KEY:
-        if (!sliderIsActive && !this.controller.state.isPlayingAd) {
```
aeng7: Does something else prevent the ad from being seeked?

js/components/accessibilityControls.js
```diff
@@ -88,6 +88,30 @@ AccessibilityControls.prototype = {
   },
 
   /**
+   * Determines whether or not the controller is in a state that allows seeking the video.
+   * @private
+   * @return {Boolen} True if seeking is possible, false otherwise.
```
aeng7: Boolen is misspelled, only fix if there are actual fixes to apply

js/components/accessibilityControls.js
```diff
@@ -88,6 +88,30 @@ AccessibilityControls.prototype = {
...
+   */
+  canSeek: function() {
+    var seekingEnabled = false;
+    switch (this.controller.state.screenToShow) {
+      case CONSTANTS.SCREEN.PLAYING_SCREEN:
+      case CONSTANTS.SCREEN.PAUSE_SCREEN:
+      case CONSTANTS.SCREEN.END_SCREEN:
```
aeng7: Does End Screen cover replay, discovery, etc?

jpfranco: These are the only three screens that show a progress bar (end has the replay button and the progress bar). We don't allow seeking on discovery, share or any of the other modals.

aeng7: Got it.

aeng7:
LGTM

PR #812, opened 2017-09-11
Feature/player 1413

IvanSakharovskii:
Watch monoscopic video 360 on MAC and Windows desktops


js/controller.js
```diff
@@ -259,6 +269,19 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+        this.mb.subscribe(OO.EVENTS.DIRECTION_CHANGED, 'customerUi', _.bind(this.setViewingDirection, this));
+      }
+    },
+
+    getVrParams: function(){
+      var playerParam = this.state.playerParam;
+      var bitWrapper = playerParam ? playerParam['bit-wrapper'] : null;
+      var isVr = !!bitWrapper && !!bitWrapper.source && !!bitWrapper.source.vr;
+
+      return isVr ? _.extend({}, bitWrapper.source.vr) : false;
```
pilievOoyala: Please do not call functions that change values of variables inside of a turnary statement. Stick this inside of an if statement. Otherwise using the debugger to debug this will be much harder.

js/controller.js
```diff
@@ -245,7 +253,9 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       // Setting the tabindex will let some screen readers recognize this element as a group
       // identified with the ARIA label above. We set it to -1 in order to prevent actual keyboard focus
       this.state.mainVideoInnerWrapper.attr('tabindex', '-1');
-      this.state.mainVideoInnerWrapper.append("<div class='oo-player-skin'></div>");
+
+      var $ooPlayerSkin = $('.oo-player-skin');
+      !$ooPlayerSkin.length && this.state.mainVideoInnerWrapper.append("<div class='oo-player-skin'></div>");
```
pilievOoyala: Please change this to an if statement.  LIke my other comment below, we don't call functions that change the value of variables within checks like this to allow for easier debugging.

js/views/playingScreen.js
```diff
@@ -156,14 +158,22 @@ var PlayingScreen = React.createClass({
...
     return (
-    <div className="oo-state-screen oo-playing-screen"
-         ref="PlayingScreen"
-         onMouseOver={this.showControlBar}
-         onMouseOut={this.hideControlBar}
-         onMouseMove={this.handlePlayerMouseMove}
-         onKeyDown={this.handleKeyDown}>
-
-      <div className="oo-state-screen-selectable" onMouseUp={this.handlePlayerMouseUp} onTouchEnd={this.handleTouchEnd}></div>
+    <div
```
pilievOoyala: The alignment here is off.  We use tabs of 2 spaces but this is 3 spaces.

tests/views/playingScreen-test.js
```diff
@@ -37,16 +38,62 @@ describe('PlayingScreen', function () {
     TestUtils.Simulate.mouseUp(screen[0]);
     expect(clicked).toBe(true);
   });
+  it('creates a PlayingScreen and checks mouseDown, mouseUp with video360', function() {
```
pilievOoyala: need an extra new line between tests.

tests/views/playingScreen-test.js
```diff
@@ -37,16 +38,62 @@ describe('PlayingScreen', function () {
...
+    var isTouched = false
+      , isStartHideControlBarTimer = false;
```
pilievOoyala: can you use var on the second line as well?  Otherwise you can technically just do 

var isTouched = false
      , isStartHideControlBarTimer = false
      , mockController = {

And the current team tries to avoid multiple variable definitions using ','

tests/views/playingScreen-test.js
```diff
@@ -37,16 +38,62 @@ describe('PlayingScreen', function () {
...
+        upNextInfo: {
+          showing: false
+        },
+        isVideo360: true,
+        viewingDirection: {yaw: 0, roll: 0, pitch: 0}
+      },
+      startHideControlBarTimer: function () {
+        isStartHideControlBarTimer = true;
+      },
+      onTouched: function() { isTouched = true; },
```
pilievOoyala: althought it is small, please change it to 
onTouched: function() { 
  isTouched = true; 
},


It's much easier to step through this using the debugger this way.

tests/views/playingScreen-test.js
```diff
@@ -37,16 +38,62 @@ describe('PlayingScreen', function () {
...
+    );
+    DOM.setState({
+      isMouseDown: true,
+      XMouseStart: -10,
+      YMouseStart: -20
+    });
+
+    var screen = TestUtils.findRenderedDOMComponentWithClass(DOM, 'oo-state-screen-selectable');
+
+    // TestUtils.Simulate.mouseDown(screen[0]);
```
pilievOoyala: Can you take out the commented out code?

tests/views/playingScreen-test.js
```diff
@@ -37,16 +38,62 @@ describe('PlayingScreen', function () {
...
+    // expect(isTouched).toBe(true);
+    //
+    // TestUtils.Simulate.mouseUp(screen[0]);
+    // expect(isTouched).toBe(true);
+
+  });
 
-  it('creates a PlayingScreen and checks touchEnd', function () {
+  it('creates a PlayingScreen and checks touchEnd without video360', function () {
```
pilievOoyala: Excellent catch!

js/controller.js
```diff
@@ -259,6 +269,19 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
 
       this.accessibilityControls = new AccessibilityControls(this); //keyboard support
       this.state.screenToShow = CONSTANTS.SCREEN.INITIAL_SCREEN;
+
+      if (this.getVrParams && this.getVrParams()) {
```
aeng7: Can we check if typeof this.getVrParams is a function?

js/controller.js
```diff
@@ -259,6 +269,19 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+        this.state.isVideo360 = true;
+        this.mb.subscribe(OO.EVENTS.DIRECTION_CHANGED, 'customerUi', _.bind(this.setViewingDirection, this));
+      }
+    },
+
+    getVrParams: function(){
+      var playerParam = this.state.playerParam;
+      var bitWrapper = playerParam ? playerParam['bit-wrapper'] : null;
```
aeng7: We should define a generic way to relay this information from the video plugin to the skin. We do not want to add dependencies on bitmovin in the skin files.

js/controller.js
```diff
@@ -259,6 +269,19 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+      var isVr = !!bitWrapper && !!bitWrapper.source && !!bitWrapper.source.vr;
```
aeng7: We should define a generic way to pass this information from the video plugin to the skin (like via message bus). We do not want any dependencies on bitmovin in our skin.

js/views/playingScreen.js
```diff
@@ -156,14 +158,22 @@ var PlayingScreen = React.createClass({
...
+      className="oo-state-screen oo-playing-screen"
+      ref="PlayingScreen"
+      onMouseOver={this.showControlBar}
+      onMouseOut={this.hideControlBar}
+      onMouseMove={this.handlePlayerMouseMove}
+      onMouseLeave={this.handlePlayerMouseLeave}
+      onKeyUp={this.handleKeyDown}
```
aeng7: Why did we change this from onKeyDown to onKeyUp?

js/controller.js
```diff
@@ -643,11 +664,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
     },
 
     onBuffering: function(event) {
-      if (this.state.isInitialPlay == false && this.state.screenToShow == CONSTANTS.SCREEN.START_SCREEN) {
-        this.state.buffering = false;
-      } else {
-        this.state.buffering = true;
-      }
+      this.state.buffering = !(this.state.isInitialPlay == false && this.state.screenToShow == CONSTANTS.SCREEN.START_SCREEN);
```
pilievOoyala: Can you change this to ===?

aldigun:
You should open a PR to the submodule/skin.json. Submodule needs to be approved first.
On that note, in skin.json change. What is "stereo"?

js/constants/constants.js
```diff
@@ -40,9 +40,7 @@ module.exports = {
     OFF: "Off",
     DISCOVER: "Discover",
     UP_NEXT: "Up next",
-    VIDEO_QUALITY: "Video Quality",
```
pilievOoyala: Why is this being taken out?

pilievOoyala:
LGTM (looks good to me)

js/controller.js
```diff
@@ -260,6 +268,15 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
 
       this.accessibilityControls = new AccessibilityControls(this); //keyboard support
       this.state.screenToShow = CONSTANTS.SCREEN.INITIAL_SCREEN;
+
+      if (this.videoVr) {
+        this.mb.subscribe(OO.EVENTS.DIRECTION_CHANGED, 'customerUi', _.bind(this.setViewingDirection, this));
```
aeng7: OO.EVENTS.DIRECTION_CHANGED: I'm not sure where this event is defined, but can we rename it to include a mention of VR?

tests/views/playingScreen-test.js
```diff
@@ -38,17 +43,62 @@ describe('PlayingScreen', function () {
...
   });
 
-  it('creates a PlayingScreen and checks touchEnd', function () {
+  it('creates a PlayingScreen and checks mouseDown, mouseUp with video360', function() {
+    var isTouched = false;
+    var isStartHideControlBarTimer = false;
+    var mockController = {
+      state: {
+        isMobile: false,
+        accessibilityControlsEnabled: false,
```
aeng7: Do we have a defined behavior when accessibility controls are enabled? We should add some tests to test the defined behavior if we do have it defined.

If we do not have a defined behavior, what happens when accessibility controls are enabled? Do we have any conflicts with keyboard controls? For example, arrow keys might be used to navigate between elements when they are enabled.

js/controller.js
yvinokurov: This functionality together with your wishes is moved to feature / PLAYER-1422

aeng7:
LGTM, please look into accessibility interactions at a future date

PR #813, opened 2017-09-11
Feature/player 1422

IvanSakharovskii:
Merge only after PLAYER-1413

look around using drag & drop mouse events


js/components/controlBar.js
```diff
@@ -22,7 +22,13 @@ var ControlBar = React.createClass({
     this.isMobile = this.props.controller.state.isMobile;
     this.responsiveUIMultiple = this.getResponsiveUIMultiple(this.props.responsiveView);
     this.moreOptionsItems = null;
-    return {};
+    this.vr = this.props.controller
```
pilievOoyala: if this.props.controller is false, does this.vr become false or null?

pilievOoyala: I think this would be clearer in an if statement.

js/components/controlBar.js
```diff
@@ -22,7 +22,13 @@ var ControlBar = React.createClass({
...
+      && this.props.controller.getVrParams
+      && this.props.controller.getVrParams();
+    
+    return {
+      currentVolumeHead: 0
```
pilievOoyala: Should we actually be returning this since we are not explicitly setting the volume in the for this function?  is there a 'this' variable we can stick it into instead.

js/controller.js
```diff
@@ -216,6 +219,12 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       this.state.isSubscribed = true;
     },
 
+    vrSubscribes: function () {
```
pilievOoyala: this should be vrSubsribe. No 's' at the end.

js/controller.js
```diff
@@ -245,7 +254,9 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       // Setting the tabindex will let some screen readers recognize this element as a group
       // identified with the ARIA label above. We set it to -1 in order to prevent actual keyboard focus
       this.state.mainVideoInnerWrapper.attr('tabindex', '-1');
-      this.state.mainVideoInnerWrapper.append("<div class='oo-player-skin'></div>");
+
+      var $ooPlayerSkin = $('.oo-player-skin');
+      !$ooPlayerSkin.length && this.state.mainVideoInnerWrapper.append("<div class='oo-player-skin'></div>");
```
pilievOoyala: Same thing here, please make this into an if statement.

js/controller.js
```diff
@@ -259,6 +270,19 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
 
       this.accessibilityControls = new AccessibilityControls(this); //keyboard support
       this.state.screenToShow = CONSTANTS.SCREEN.INITIAL_SCREEN;
+
+      if (this.getVrParams && this.getVrParams()) {
```
pilievOoyala: i think if you use _.isFunction(this.getVrParams)  will be more what you want since, calling this.getVrParams() can still fail if it's string or number.

js/controller.js
```diff
@@ -259,6 +270,19 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+        this.mb.subscribe(OO.EVENTS.DIRECTION_CHANGED, 'customerUi', _.bind(this.setViewingDirection, this));
+      }
+    },
+
+    getVrParams: function(){
+      var playerParam = this.state.playerParam;
+      var bitWrapper = playerParam ? playerParam['bit-wrapper'] : null;
+      var isVr = !!bitWrapper && !!bitWrapper.source && !!bitWrapper.source.vr;
+
+      return isVr ? _.extend({}, bitWrapper.source.vr) : false;
```
pilievOoyala: if statement please.

js/controller.js
```diff
@@ -1222,6 +1269,11 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
     toggleMute: function(muted) {
       this.mb.publish(OO.EVENTS.CHANGE_VOLUME, (muted ? 0 : 1));
     },
+  
+    moveToDirection: function (rotate, direction) {
+      OO.log("moveToDirection is called");
```
pilievOoyala: change to "moveToDirection called"

js/mixins/responsiveManagerMixin.js
```diff
@@ -28,6 +29,10 @@ var ResponsiveManagerMixin = {
 
   generateResponsiveData: function() {
     var componentWidth = ReactDOM.findDOMNode(this).getBoundingClientRect().width;
```
pilievOoyala: I know this is in the original code too, but if ReactDOM.findDOMNode(this) fails, we'll through an error.  Can you store it in a local variable and make sure it's valid?

js/mixins/responsiveManagerMixin.js
```diff
@@ -56,6 +61,7 @@ var ResponsiveManagerMixin = {
     //set responsive data to state
     this.setState({
       componentWidth: componentWidth,
+      componentHeight: componentHeight > 0  ? componentHeight : 1,
```
pilievOoyala: Because in the debugger all of this is one step, can you move "componentHeight > 0  ? componentHeight : 1" to a local variable so we know what the value is. It's easier to debug.

js/views/playingScreen.js
```diff
@@ -115,16 +121,94 @@ var PlayingScreen = React.createClass({
...
+    }
+    
+    this.setState({
+      isMouseMove: false,
+    });
+  },
+  
+  getDirectionParams: function(pageX, pageY) {
+    var dx = pageX - this.state.XMouseStart
+      , dy = pageY - this.state.YMouseStart;
```
pilievOoyala: can you use var for all the new declarations?

js/views/playingScreen.js
```diff
@@ -156,14 +240,22 @@ var PlayingScreen = React.createClass({
...
     return (
-    <div className="oo-state-screen oo-playing-screen"
-         ref="PlayingScreen"
-         onMouseOver={this.showControlBar}
-         onMouseOut={this.hideControlBar}
-         onMouseMove={this.handlePlayerMouseMove}
-         onKeyDown={this.handleKeyDown}>
-
-      <div className="oo-state-screen-selectable" onMouseUp={this.handlePlayerMouseUp} onTouchEnd={this.handleTouchEnd}></div>
+    <div
```
pilievOoyala: indentation is wrong

tests/controller-test.js
```diff
@@ -212,7 +212,9 @@ OO = {
       createPluginElements: function() {},
       findMainVideoElement: function(a) {},
       loadConfigData: function(a, b, c, d) {},
-      cleanUpEventListeners: function(){}
+      cleanUpEventListeners: function(){},
```
pilievOoyala: Please make the spacing consistent with all the other function declarations here.

tests/views/playingScreen-test.js
```diff
@@ -106,6 +154,102 @@ describe('PlayingScreen', function () {
     expect(over).toBe(true);
   });
 
+  it('creates a PlayingScreen and checks mouseMove, mouseOver, mouseOut, keyUp with video360 fullscreen', function () {
+    var over = false
+      , out = false
```
pilievOoyala: use var for all these

tests/views/playingScreen-test.js
```diff
@@ -106,6 +154,102 @@ describe('PlayingScreen', function () {
...
+        isMobile: false,
+        isVideo360: true,
+        accessibilityControlsEnabled: false,
+        upNextInfo: {
+          showing: false
+        },
+        viewingDirection: {yaw: 0, roll: 0, pitch: 0}
+      },
+      startHideControlBarTimer: function() {moved = true},
+      togglePlayPause: function(){clicked = true},
```
pilievOoyala: space between ) and {

tests/views/playingScreen-test.js
```diff
@@ -106,6 +154,102 @@ describe('PlayingScreen', function () {
...
+    DOM.setState({
+      isMouseDown: true,
+      XMouseStart: -10,
+      YMouseStart: -20
+    });
+
+    var screen = TestUtils.scryRenderedDOMComponentsWithClass(DOM, 'oo-playing-screen');
+
+    var getDirectionParams = spyOn(DOM, 'getDirectionParams').andCallThrough()
+      , directionParams = getDirectionParams(0,0);
```
pilievOoyala: var

tests/views/playingScreen-test.js
```diff
@@ -188,7 +332,7 @@ describe('PlayingScreen', function () {
         }
       },
       startHideControlBarTimer: function() {moved = true},
-      togglePlayPause: function(){clicked = true},
+      // togglePlayPause: function(){clicked = true},
```
pilievOoyala: Just remove it if you need to comment it out.

js/components/controlBar.js
```diff
@@ -22,7 +22,13 @@ var ControlBar = React.createClass({
     this.isMobile = this.props.controller.state.isMobile;
     this.responsiveUIMultiple = this.getResponsiveUIMultiple(this.props.responsiveView);
     this.moreOptionsItems = null;
-    return {};
+    this.vr = this.props.controller
+      && this.props.controller.getVrParams
```
aeng7: Can we also add a check to see if getVrParams is a function, since that is our expectation?

js/controller.js
```diff
@@ -259,6 +270,19 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+      if (this.getVrParams && this.getVrParams()) {
+        this.state.isVideo360 = true;
+        this.mb.subscribe(OO.EVENTS.DIRECTION_CHANGED, 'customerUi', _.bind(this.setViewingDirection, this));
+      }
+    },
+
+    getVrParams: function(){
+      var playerParam = this.state.playerParam;
+      var bitWrapper = playerParam ? playerParam['bit-wrapper'] : null;
+      var isVr = !!bitWrapper && !!bitWrapper.source && !!bitWrapper.source.vr;
```
aeng7: We should define a generic way to pass this information from the video plugin to the skin (like via message bus). We do not want any dependencies on bitmovin in our skin.

js/controller.js
```diff
@@ -617,6 +641,26 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       }
     },
 
+    onTouching: function(params, isOnVideoClick) {
+      if (this.state.playerState == CONSTANTS.STATE.PLAYING) {
+        if (this.state.isVideo360 && isOnVideoClick) {
+          this.mb.publish(OO.EVENTS.TOUCHING, this.focusedElement, params);
```
aeng7: Where does params come from? It is not defined from within the scope of this function.

js/views/playingScreen.js
```diff
@@ -115,16 +121,94 @@ var PlayingScreen = React.createClass({
       this.showControlBar(event);
       this.props.controller.startHideControlBarTimer();
     }
-    else {
-      this.props.controller.togglePlayPause();
+    else if (!this.isVideo360) {
+      this.props.controller.togglePlayPause(event);
+    }
```
aeng7: Since we have a new flow now for detecting a player click to trigger togglePlayPause, can this flow be used in the non-vr 360 path? If so, we can remove this else if statement and simplify the code.

js/views/playingScreen.js
```diff
@@ -156,14 +240,22 @@ var PlayingScreen = React.createClass({
...
-
-      <div className="oo-state-screen-selectable" onMouseUp={this.handlePlayerMouseUp} onTouchEnd={this.handleTouchEnd}></div>
+    <div
+      className="oo-state-screen oo-playing-screen"
+      ref="PlayingScreen"
+      onMouseOver={this.showControlBar}
+      onMouseOut={this.hideControlBar}
+      onMouseMove={this.handlePlayerMouseMove}
+      onMouseLeave={this.handlePlayerMouseLeave}
+      onKeyUp={this.handleKeyDown}
```
aeng7: Why did we change onKeyDown to onKeyUp?

tests/views/playingScreen-test.js
```diff
@@ -106,6 +154,102 @@ describe('PlayingScreen', function () {
...
+    var screen = TestUtils.scryRenderedDOMComponentsWithClass(DOM, 'oo-playing-screen');
+
+    var getDirectionParams = spyOn(DOM, 'getDirectionParams').andCallThrough()
+      , directionParams = getDirectionParams(0,0);
+
+    TestUtils.Simulate.mouseMove(screen[0]);
+    expect(moved).toBe(true);
+    expect(isTouching).toBe(true);
+    expect(getDirectionParams).toHaveBeenCalled();
+    expect(directionParams).toEqual([10, 0, 53.33333333333333]);
```
aeng7: Can we add a comment explaining where these numbers came from? Thanks!

tests/views/playingScreen-test.js
```diff
@@ -106,6 +154,102 @@ describe('PlayingScreen', function () {
...
+    var closedCaptionOptions = {
+      cueText: "cue text"
+    };
+  
+    // Render pause screen into DOM
+    var DOM = TestUtils.renderIntoDocument(<PlayingScreen  controller = {mockController} closedCaptionOptions = {closedCaptionOptions}/>);
+    var screen = TestUtils.findRenderedDOMComponentWithClass(DOM, 'oo-state-screen-selectable');
+    
+    TestUtils.Simulate.click(screen);
+    expect(clicked).toBe(true);
```
aeng7: For toggles, can we use a number rather than a boolean for our unit tests? We want to make sure they are clicked the expected amount of times, rather than the fact that they were just clicked. Thanks!

js/views/playingScreen.js
```diff
@@ -19,9 +19,15 @@ var PlayingScreen = React.createClass({
   getInitialState: function() {
     this.isMobile = this.props.controller.state.isMobile;
     this.browserSupportsTouch = this.props.controller.state.browserSupportsTouch;
+    this.isVideo360 = this.props.controller.state.isVideo360;
```
jpfranco: I know that the older code is already doing this, but storing props in local state is an anti-pattern. Could we use the value from props directly?

js/controller.js
KseniiaSokolovaEpam: It is params which we pass to the function (yaw, roll,  pitch).

pilievOoyala:
LGTM

tests/controller-test.js
```diff
@@ -445,6 +445,8 @@ OO = {
 
     Html5Skin.toggleMute.call(controllerMock, true);
     Html5Skin.toggleMute.call(controllerMock, false);
+  
+    Html5Skin.moveToDirection.call(controllerMock);
```
jpfranco: This file was deprecated after recent updates. Please add regular unit tests in html5skin-test.js

js/constants/constants.js
```diff
@@ -7,7 +7,8 @@ module.exports = {
     PLAYING : "playing",
     PAUSE : "pause",
     END : "end",
-    ERROR : "error"
+    ERROR : "error",
+    TOUCH: "touch"
```
pilievOoyala: I think we may need this to be TOUCHING or maybe we can think of a better name. Because the point is that you are in a state where the camera is changing direction, right?

js/controller.js
```diff
@@ -635,6 +641,41 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+        this.mb.publish(OO.EVENTS.TOUCHED, this.focusedElement);
+      }
+    },
+  
+    setViewingDirection: function(event, yaw, roll, pitch) {
+      this.state.viewingDirection = {yaw: yaw, roll: roll, pitch: pitch};
+    },
+  
+    recreatingUI: function (event, elementId, params, settings) {
+      console.log('call recreatingUI args', arguments);
```
pilievOoyala: Is this leftover debug code, or is it useful to have all the time?  Maybe put this in a debugVR check.

js/controller.js
```diff
@@ -28,7 +28,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
   var Html5Skin = function (mb, id) {
     this.mb = mb;
     this.id = id;
-    this.videoVrSource = {};
+    this.videoVrSource = false;
```
aeng7: Would null work better here? (if videoVrSource is intended to be an object)

js/controller.js
```diff
@@ -268,11 +270,15 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+
+      if (this.videoVr) {
+        this.mb.subscribe(OO.EVENTS.DIRECTION_CHANGED, 'customerUi', _.bind(this.setViewingDirection, this));
+      }
     },
 
     setVideoVr: function(event, obj) {
       this.videoVr = true;
-      this.videoVrSource = obj.source || {}; //if we need video vr params
+      this.videoVrSource = obj.source || false; //if we need video vr params
```
aeng7: Would null work better here? (if videoVrSource is intended to be an object)

js/controller.js
```diff
@@ -635,6 +641,41 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+      if (params.skin && params.skin.config) {
+        $.getJSON(params.skin.config, function(data) {
+          this.state.customSkinJSON = data;
+          this.loadConfigData(this.state.playerParam, this.state.persistentSettings, data, this.state.skinMetaData);
+        }.bind(this));
+      } else {
+        this.loadConfigData(this.state.playerParam, this.state.persistentSettings, this.state.customSkinJSON, this.state.skinMetaData);
+      }
+  
+      this.accessibilityControls = new AccessibilityControls(this); //keyboard support
```
aeng7: @jpfranco , would this conflict with our existing accessibility controls?

js/views/playingScreen.js
```diff
@@ -19,11 +19,15 @@ var PlayingScreen = React.createClass({
...
+    this.videoVr = this.props.controller.videoVr;
 
     return {
       controlBarVisible: true,
-      timer: null
+      timer: null,
+      isMouseDown: false,
+      isMouseMove: false,
+      XMouseStart: 0,
+      YMouseStart: 0,
```
aeng7: Can we rename these 4 states to include "vr" since setting them requires videoVr to be true (such as in the handlePlayerMouseDown function)? Or otherwise, we can set them regardless of whether videoVr is true or not and adjust logic accordingly.

js/controller.js
KseniiaSokolovaEpam: +

js/controller.js

js/views/playingScreen.js

js/controller.js
```diff
@@ -169,6 +169,8 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       this.mb.subscribe(OO.EVENTS.ASSET_UPDATED, 'customerUi', _.bind(this.onAssetUpdated, this));
       this.mb.subscribe(OO.EVENTS.PLAYBACK_READY, 'customerUi', _.bind(this.onPlaybackReady, this));
       this.mb.subscribe(OO.EVENTS.VIDEO_VR, 'customerUi', _.bind(this.setVideoVr, this));
+      this.mb.subscribe(OO.EVENTS.VR_DIRECTION_CHANGED, 'customerUi', _.bind(this.setViewingDirection, this));
+      this.mb.subscribe(OO.EVENTS.RECREATING_UI, 'customerUi', _.bind(this.recreatingUI, this));
```
jpfranco: We should match the naming convention from other events, which is `onEventName`

js/controller.js
jpfranco: Yes, it looks like this would recreate the accessibility controls, which I don't think is necessary. Since we're not calling `AccessibilityControls.cleanUp()` this will also duplicate all existing event listeners. Unless there's a good reason for this I would suggest removing it.

js/controller.js
```diff
@@ -268,11 +270,15 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
 
       this.accessibilityControls = new AccessibilityControls(this); //keyboard support
       this.state.screenToShow = CONSTANTS.SCREEN.INITIAL_SCREEN;
+
+      if (this.videoVr) {
+        this.mb.subscribe(OO.EVENTS.VR_DIRECTION_CHANGED, 'customerUi', _.bind(this.setViewingDirection, this));
```
jpfranco: We don't subscribe to any other events in `onPlayerCreated`. Would it be better to handle this on `init()` or `subscribeBasicPlaybackEvents`?

js/controller.js
```diff
@@ -635,6 +641,40 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+    },
+  
+    recreatingUI: function (event, elementId, params, settings) {
+      if (!$('.oo-player-skin').length) {
+        this.state.mainVideoInnerWrapper.append("<div class='oo-player-skin'></div>")
+      }
+  
+      //load player with page level config param if exist
+      if (params.skin && params.skin.config) {
+        $.getJSON(params.skin.config, function(data) {
```
jpfranco: Is it necessary to request skin.json again? It looks like `this.state.customSkinJSON`is treated as read-only an can probably be re-used. We should confirm whether this is the case.

aeng7:
LGTM, please make sure to address the outstanding comments asap

js/controller.js

PR #822, opened 2017-09-14
[PLAYER-1598] [PLAYER-1698] Unit tests + Restricting tab navigation on fullscreen

jpfranco:
[PLAYER-1598]

- Added all the pending unit tests for scrubber and volume bar accessibility updates.

 [PLAYER-1698]

- Restricting tab navigation to control bar elements when player is in fullscreen mode. Tab focus will loop around on player elements only.

### DIFF
--- a/js/components/controlBar.js
+++ b/js/components/controlBar.js
@@ -224,15 +224,17 @@ var ControlBar = React.createClass({
   /**
    * Will handle the keydown event when the controlBar is active and it will restrict
    * tab navigation to elements that are within it when the player is in fullscreen mode.
+   * Note that this only handles the edge cases that are needed in order to loop the tab
+   * focus. Tabbing in between the elements is handled by the browser.
    * @private
    * @param {Object} evt Keydown event object.
    */
   handleControlBarKeyDown: function(evt) {
     if (
-      evt.key !== CONSTANTS.KEY_VALUES.TAB
-      || !this.props.controller.state.fullscreen
-      || !this.domNode
-      || !evt.target
+      evt.key !== CONSTANTS.KEY_VALUES.TAB ||
+      !this.props.controller.state.fullscreen ||
+      !this.domNode ||
+      !evt.target
     ) {
       return;
     }
@@ -268,6 +270,8 @@ var ControlBar = React.createClass({
           firstFocusableElement.focus();
         }
       }
+    } else {
+      OO.log('ControlBar: No focusable elements found');
     }
   },
 
--- a/js/components/controlBar.js
+++ b/js/components/controlBar.js
@@ -27,6 +27,7 @@ var ControlBar = React.createClass({
 
   componentDidMount: function () {
     window.addEventListener('orientationchange', this.closePopovers);
+    document.addEventListener('keydown', this.handleControlBarKeyDown);
     this.restoreFocusedControl();
   },
 
@@ -44,6 +45,7 @@ var ControlBar = React.createClass({
       this.props.controller.hideVolumeSliderBar();
     }
     window.removeEventListener('orientationchange', this.closePopovers);
+    document.removeEventListener('keydown', this.handleControlBarKeyDown);
   },
 
   /**
@@ -217,6 +219,56 @@ var ControlBar = React.createClass({
    */
   handleControlBarBlur: function(evt) {
     this.props.controller.state.focusedControl = null;
+  },
+
+  /**
+   * Will handle the keydown event when the controlBar is active and it will restrict
+   * tab navigation to elements that are within it when the player is in fullscreen mode.
+   * @private
+   * @param {Object} evt Keydown event object.
+   */
+  handleControlBarKeyDown: function(evt) {
+    if (
+      evt.key !== CONSTANTS.KEY_VALUES.TAB
+      || !this.props.controller.state.fullscreen
+      || !this.domNode
+      || !evt.target
+    ) {
+      return;
+    }
+    // Focusable elements on the control bar (this.domNode) are expected to have the
+    // data-focus-id attribute
+    var focusableElements = this.domNode.querySelectorAll('[data-focus-id]');
+
+    if (focusableElements.length) {
+      var firstFocusableElement = focusableElements[0];
+      var lastFocusableElement = focusableElements[focusableElements.length - 1];
+      // This indicates we're tabbing over the focusable control bar elements
+      if (evt.target.hasAttribute('data-focus-id')) {
+        if (evt.shiftKey) {
+          // Shift + tabbing on first element, focus on last
+          if (evt.target === firstFocusableElement) {
+            evt.preventDefault();
+            lastFocusableElement.focus();
+          }
+        } else {
+          // Tabbing on last element, focus on first
+          if (evt.target === lastFocusableElement) {
+            evt.preventDefault();
+            firstFocusableElement.focus();
+          }
+        }
+      // Keydown happened on a non-controlbar element
+      } else {
+        evt.preventDefault();
+
+        if (evt.shiftKey) {
+          lastFocusableElement.focus();
+        } else {
+          firstFocusableElement.focus();
+        }
+      }
+    }
   },
 
   populateControlBar: function () {
--- a/tests/components/accessibilityControls-test.js
+++ b/tests/components/accessibilityControls-test.js
@@ -1,5 +1,6 @@
 jest.dontMock('../../js/components/accessibilityControls');
 jest.dontMock('../../js/constants/constants');
+jest.dontMock('../../js/components/utils');
 
 var CONSTANTS = require('../../js/constants/constants');
 var AccessibilityControls = require('../../js/components/accessibilityControls');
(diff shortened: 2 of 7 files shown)


js/components/controlBar.js
```diff
@@ -219,6 +221,56 @@ var ControlBar = React.createClass({
...
+  /**
+   * Will handle the keydown event when the controlBar is active and it will restrict
+   * tab navigation to elements that are within it when the player is in fullscreen mode.
+   * @private
+   * @param {Object} evt Keydown event object.
+   */
+  handleControlBarKeyDown: function(evt) {
+    if (
+      evt.key !== CONSTANTS.KEY_VALUES.TAB
+      || !this.props.controller.state.fullscreen
```
pilievOoyala: Not super important, I think we usually put the && or || at the end of the line .

js/components/controlBar.js
```diff
@@ -219,6 +221,56 @@ var ControlBar = React.createClass({
...
 
+  /**
+   * Will handle the keydown event when the controlBar is active and it will restrict
+   * tab navigation to elements that are within it when the player is in fullscreen mode.
+   * @private
+   * @param {Object} evt Keydown event object.
+   */
+  handleControlBarKeyDown: function(evt) {
+    if (
+      evt.key !== CONSTANTS.KEY_VALUES.TAB
```
pilievOoyala: is evt guaranteed to not be null?  Someone may end up calling this outside of a callback.

js/components/controlBar.js
```diff
@@ -219,6 +221,56 @@ var ControlBar = React.createClass({
...
+      || !this.domNode
+      || !evt.target
+    ) {
+      return;
+    }
+    // Focusable elements on the control bar (this.domNode) are expected to have the
+    // data-focus-id attribute
+    var focusableElements = this.domNode.querySelectorAll('[data-focus-id]');
+
+    if (focusableElements.length) {
```
pilievOoyala: If focusableElements is empty, can we have a debug log or is it valid to not have any elements?  Want to make sure if someone is messing with css or something else and accidentally erases all these elements, they have a warning.

js/components/controlBar.js
pilievOoyala: Now that I look at the rest of the code, it's very dependent on evt. So it's probably ok not having the check.

js/components/controlBar.js
```diff
@@ -219,6 +221,56 @@ var ControlBar = React.createClass({
     this.props.controller.state.focusedControl = null;
   },
 
+  /**
+   * Will handle the keydown event when the controlBar is active and it will restrict
```
pilievOoyala: Can you add in that this only handles the edge case for wrapping at the beginning or the end.  The OS handles the rest.

js/components/controlBar.js
jpfranco: Ok, I can update this

js/components/controlBar.js
jpfranco: Not very likely since these are elements like the scrubber bar, play/pause button, but I can add log

js/components/controlBar.js
jpfranco: Yes this can only work as an event handler. I think it should just crash if called directly since it's not its intended use

jpfranco:
@pilievOoyala updated the PR

js/components/controlBar.js
```diff
@@ -219,6 +221,60 @@ var ControlBar = React.createClass({
...
+    ) {
+      return;
+    }
+    // Focusable elements on the control bar (this.domNode) are expected to have the
+    // data-focus-id attribute
+    var focusableElements = this.domNode.querySelectorAll('[data-focus-id]');
+
+    if (focusableElements.length) {
+      var firstFocusableElement = focusableElements[0];
+      var lastFocusableElement = focusableElements[focusableElements.length - 1];
```
aeng7: Is it possible for these first and last elements to be opposite in terms of tab order that the browser handles? Like the "last" focusable element precedes the first when just normally tabbing. 

Also, do we need any safeguards that the user isn't stuck tabbing between the first and last element only?

tests/components/accessibilityControls-test.js
```diff
@@ -133,6 +134,147 @@ describe('AccessibilityControls', function () {
...
+        expect(setVolumeCalled).toBe(false);
+      });
+
+      it('should constrain volume to supported values', function() {
+        var volume = 0;
+        mockCtrl.setVolume = function(vol) {
+          volume = vol;
+        };
+        mockCtrl.state.volumeState.volume = 0.5;
+        a11yCtrls.changeVolumeBy(500, true);
```
aeng7: Can we add a check to see if negative volumes are ignored as well?

tests/components/accessibilityControls-test.js
```diff
@@ -133,6 +134,147 @@ describe('AccessibilityControls', function () {
...
+        mockCtrl.skin.state.currentPlayhead = 30;
+        a11yCtrls.seekBy(120, false);
+        expect(newPlayhead).toBe(0);
+      });
+
+      it('should not seek when seeking is disabled', function() {
+        mockCtrl.state.screenToShow = CONSTANTS.SCREEN.AD_SCREEN;
+        mockCtrl.skin.state.duration = 60;
+        mockCtrl.skin.state.currentPlayhead = 0;
+        a11yCtrls.seekBy(5, true);
```
aeng7: Can we add a test to check that negative values are ignored?

tests/components/controlBar-test.js
```diff
@@ -1624,4 +1650,79 @@ describe('ControlBar', function () {
...
+
+    beforeEach(function() {
+      baseMockProps.skinConfig.buttons.desktopContent = [
+        { "name": "playPause", "location": "controlBar", "whenDoesNotFit": "keep", "minWidth": 45 },
+        { "name": "volume", "location": "controlBar", "whenDoesNotFit": "keep", "minWidth": 240 },
+        { "name": "fullscreen", "location": "controlBar", "whenDoesNotFit": "keep", "minWidth": 45 },
+      ];
+      // Mock addEventListener on document object since TestUtils.Simulate will not work in this case
+      eventMap = {};
+      document.addEventListener = jest.genMockFn().mockImpl(function(event, cb) {
```
aeng7: Do we need to restore this function in afterEach?

tests/components/controlBar-test.js
```diff
@@ -1624,4 +1650,79 @@ describe('ControlBar', function () {
...
+      expect(document.activeElement.getAttribute('data-focus-id')).toBe(focusableElements[focusableElements.length - 1].getAttribute('data-focus-id'));
+    });
+
+    it('should NOT constrain tab navigation to control bar elements when NOT in fullscreen mode', function() {
+      baseMockController.state.fullscreen = false;
+      // Tab on last focusable element, should NOT go back to the first
+      document.activeElement = null;
+      mockEvent.target = focusableElements[focusableElements.length - 1];
+      eventMap.keydown(mockEvent);
+      expect(document.activeElement).toBeNull();
```
aeng7: Is there a better way to test that we are not constraining tab navigation to control bar elements? Can we add non-player elements to our test DOM and check if those become the active element?

tests/components/scrubberBar-test.js
```diff
@@ -12,11 +13,51 @@ var skinConfig = require('../../config/skin.json');
...
+      controlBarVisible: baseMockController.state.controlBarVisible,
+      playerState: baseMockController.state.playerState,
+      isLiveStream: baseMockController.state.isLiveStream,
+      controller: baseMockController
+    });
+  };
+
+  // TODO
+  // Old unit tests should use the base mock controller and skinConfig
+  // instead of defining them manually each time
```
aeng7: Nice catch!

tests/components/scrubberBar-test.js
```diff
@@ -58,6 +99,80 @@ describe('ScrubberBar', function () {
...
+  it('should update the ARIA value in order to reflect the current playhead', function() {
+    baseMockController.state.currentPlayhead = 2;
+    baseMockController.state.duration = 60;
+    updateBaseMockProps();
+    var DOM = TestUtils.renderIntoDocument(<ScrubberBar {...baseMockProps}/>);
+    var scrubberBar = TestUtils.findRenderedDOMComponentWithClass(DOM, "oo-scrubber-bar");
+    expect(scrubberBar.getAttribute('aria-valuenow')).toBe('2.00');
+    expect(scrubberBar.getAttribute('aria-valuetext')).toBe('00:02 of 01:00');
+
+    baseMockController.state.currentPlayhead = 60;
```
aeng7: What happens when currentPlayhead is greater than duration with regards to the aria attributes? I know we have safeguards to prevent this from happening, but if it does happen, would anything break?

tests/components/controlBar-test.js
jpfranco: Will check, we probably do

tests/components/controlBar-test.js
jpfranco: I actually tried this first but I think jsdom only goes as far as implementing the `focus()` function on elements, but it will not react to tab key events. This behavior would have to be mocked, so I don't think it helps a lot

tests/components/scrubberBar-test.js
jpfranco: The screen reader might read out the wrong value

tests/components/accessibilityControls-test.js
jpfranco: Will do

tests/components/accessibilityControls-test.js
jpfranco: Hmm this actually inverts the direction of the seek, but works fine. I'm trying to think what would be the proper way to react to negative values. Any thoughts?

js/components/controlBar.js
jpfranco: I haven't found any official confirmation, but `querySelectorAll` uses depth-first pre-order traversal and it seems that the regular tab navigation flow (with tabindex=0) does this as well. At least according to this: https://goo.gl/8pF94w
If this is true then the element order will always be the same as the tab order. I did try this in all browsers though, and it works fine.

tests/components/accessibilityControls-test.js
aeng7: That sounds good to me. Can we add a test for this? Can we also check the same thing for changeVolumeBy if it functions similarly (negative inputs invert the direction)? Thanks!

tests/components/controlBar-test.js
aeng7: Have you seen any issues with manual testing? This sounds fine.

tests/components/scrubberBar-test.js
aeng7: That's fine.

js/components/controlBar.js
aeng7: Sounds good. Thanks for the thorough check!

tests/components/accessibilityControls-test.js
jpfranco: Sure, will add it

tests/components/controlBar-test.js
jpfranco: Nope, I tested this and it looks ok

tests/components/accessibilityControls-test.js
jpfranco: @aeng7 changeVolumeBy works a little differently. Since that parameter is a percentage it will just constrain it to 0 percent. I will add this case to the tests

jpfranco:
@pilievOoyala @aeng7 pushed one more update

aeng7:
LGTM

PR #825, opened 2017-09-18
[PLAYER-1761] Delaying rendering of loading spinner

jpfranco:
The skin will no longer render the loading spinner immediately after the video stalls. When a stalling event occurs the skin will wait for a specified amount of time and then it will render the spinner. If playback resumes before the time elapses the spinner will simply be cancelled.

### DIFF
--- a/js/constants/constants.js
+++ b/js/constants/constants.js
@@ -148,7 +148,8 @@ module.exports = {
 
   UI: {
     defaultScrubberBarHeight: 4,
-    DEFAULT_SCRUBBERBAR_LEFT_RIGHT_PADDING: 15
+    DEFAULT_SCRUBBERBAR_LEFT_RIGHT_PADDING: 15,
+    BUFFERING_SPINNER_DELAY: 1000
   },
 
   WATERMARK: {
--- a/js/constants/constants.js
+++ b/js/constants/constants.js
@@ -149,7 +149,7 @@ module.exports = {
   UI: {
     defaultScrubberBarHeight: 4,
     DEFAULT_SCRUBBERBAR_LEFT_RIGHT_PADDING: 15,
-    BUFFERING_SPINNER_DELAY: 1000
+    BUFFERING_SPINNER_DELAY: 1250
   },
 
   WATERMARK: {
--- a/js/constants/constants.js
+++ b/js/constants/constants.js
@@ -149,7 +149,7 @@ module.exports = {
   UI: {
     defaultScrubberBarHeight: 4,
     DEFAULT_SCRUBBERBAR_LEFT_RIGHT_PADDING: 15,
-    BUFFERING_SPINNER_DELAY: 1250
+    MAX_BUFFERING_SPINNER_DELAY: 60000 // Max allowed value of bufferingSpinnerDelay in milliseconds
   },
 
   WATERMARK: {
--- a/js/controller.js
+++ b/js/controller.js
@@ -608,6 +608,8 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       }
       this.skin.updatePlayhead(this.state.duration, this.state.duration, this.state.duration);
       this.state.playerState = CONSTANTS.STATE.END;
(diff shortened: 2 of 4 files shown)


js/controller.js
```diff
@@ -683,6 +737,8 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       this.state.pluginsClickElement.removeClass("oo-showing");
       // Restore anamorphic videos fix after ad playback if necessary
       this.trySetAnamorphicFixState(true);
+      // In case ad was skipped or errored while stalled
+      this.setBufferingState(false);
```
aeng7: Nice realization!

tests/controller-test.js
```diff
@@ -1,3 +1,38 @@
+/*
+ *
+ *  !!!IMPORTANT!!!
+ *
+ *  This file is deprecated. Please use html5skin-test.js for all new controller unit tests.
```
aeng7: The monster has been slain.

tests/html5skin-test.js
```diff
@@ -0,0 +1,196 @@
...
+  beforeEach(function() {
+    controller = new Html5Skin(OO.mb, 'id');
+    controller.state.pluginsElement = $('<div/>');
+    controller.state.pluginsClickElement = $('<div/>');
+    controller.state.mainVideoElement = {
+      classList: {
+        add: function() {},
+        remove: function() {}
+      },
+      getElementsByTagName: function() { return [controller.state.mainVideoElement] },
```
aeng7: How come we want this default behavior when calling getElementsByTagName?

js/constants/constants.js
```diff
@@ -148,7 +148,8 @@ module.exports = {
 
   UI: {
     defaultScrubberBarHeight: 4,
-    DEFAULT_SCRUBBERBAR_LEFT_RIGHT_PADDING: 15
+    DEFAULT_SCRUBBERBAR_LEFT_RIGHT_PADDING: 15,
+    BUFFERING_SPINNER_DELAY: 1250
```
aeng7: Can you add a comment specifying that this is in ms?

pilievOoyala:
One super edge case, what about if the live stream ends or some other video. We want to make sure the buffer stops.

jpfranco:
@pilievOoyala  I can probably add another check when `OO.EVENTS.PLAYED` is fired. Technically the video shouldn't end if it's stalled, but it shouldn't hurt to add it. I do handle other scenarios like ads cancelled, video error, etc.

js/constants/constants.js
jpfranco: Sure, although I think this is actually going to end up in the skin.json

tests/html5skin-test.js
jpfranco: The idea was to return a mock dom element like `controller.state.mainVideoElement`, but it does look a bit confusing. I'll see if I can return something else here

jpfranco:
@pilievOoyala @aeng7 updated the PR with your suggestions and also made the delay value configurable via skin.json

This is the related skin-config pr:
https://github.com/ooyala/skin-config/pull/178

js/controller.js
```diff
@@ -673,10 +675,12 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
      */
     startBufferingTimer: function() {
       this.stopBufferingTimer();
+      var bufferingSpinnerDelay = Utils.getPropertyValue(this.skin.props.skinConfig, 'general.bufferingSpinnerDelay');
```
aeng7: Should we still have a default in case it gets deleted/omitted from skin.json?

aeng7:
@jpfranco will this need another commit with the updated submodule?

aeng7:
LGTM

js/controller.js
jpfranco: This should have a value because the custom skin is merged with the default one. `constrainToRange` will give it a default value of 0 if this was undefined for some reason though.

jpfranco:
@aeng7 yes, we will need another pr for the skin update

PR #847, opened 2017-10-02
[feature/PLAYER-1865]: new font

yvinokurov:


jpfranco:
@yvinokurov can describe what is included in this updated font? @KseniiaSokolovaEpam @IvanSakharovskii please note that pull requests pointing to master branch need to be reviewed by the core team.

IvanSakharovskii:
Here is a new icons for stereoscopic and arrows for onScreen controls.

jpfranco:
@IvanSakharovskii thanks for clarifying. It looks like we used a different service for font generation. Have you been able to test whether the old icons still look the same when scaling, etc.?

IvanSakharovskii:
@jpfranco Yeah, we tested old and new icons and it works fine

aeng7:
LGTM

jpfranco:
@IvanSakharovskii it looks like the new icons are bigger. This is not as noticeable with our default style, but it might break the styles for customers who customize the player appearance. Is it possible to add the new icons without changing the size?

Before:
before

After:
after

PR #851, opened 2017-10-04
[PLAYER-1270] Perceived performance enhancements (part two)

jpfranco:
[Ticket](https://jira.corp.ooyala.com:8443/browse/PLAYER-1270)

This will mostly help improve the UX of transitions between videos for Discovery, Playlists and custom carousels.

The following fixes were implemented:

- The skin will now show a single state between videos (loading screen).
- The progress bar of new videos will now start at 0 (it used to flash a full progress bar for about a second and then reset to 0).
- The skin will no longer flash the replay button when transitioning between playlist videos.
- The skin will no longer flash the big play button for videos that have autoplay enabled.

**This PR needs to be merged after this one:**
https://git.corp.ooyala.com/projects/PBW/repos/mjolnir/pull-requests/290/overview

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -682,9 +682,8 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       if (this.state.afterOoyalaAd) {
         this.state.screenToShow = CONSTANTS.SCREEN.LOADING_SCREEN;
       } else {
-        // If we have played a video before and PLAYBACK_READY fires it means
-        // we're transitioning to a new video. For the time being this will autoplay
-        // the next video, so we only show the loading spinner in these cases.
+        // If the core tells us that it will autoplay then we just display the loading
+        // spinner, otherwise we need to render the big play button.
         if (params.willAutoplay) {
           this.state.screenToShow = CONSTANTS.SCREEN.LOADING_SCREEN;
         } else {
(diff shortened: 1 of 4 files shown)


js/controller.js
```diff
@@ -141,6 +141,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       "isPlaybackReadySubscribed": false,
       "isSkipAdClicked": false,
       "isInitialPlay": false,
+      "initialPlayHasOccurred": false,
```
aldigun: Once this was set once in initialPlay, this is no longer reset on setEmbedCode or assetChanged.

js/controller.js
```diff
@@ -355,7 +376,9 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
     onContentTreeFetched: function (event, contentTree) {
       this.state.contentTree = contentTree;
       this.state.playerState = CONSTANTS.STATE.START;
-      this.renderSkin({"contentTree": contentTree});
+      var duration = Utils.ensureNumber(contentTree.duration, 0) / 1000;
+      this.skin.updatePlayhead(null, duration);
```
aldigun: why the fist argument on updatePlayhead is set to null?

js/controller.js
jpfranco: This is on purpose, the variable indicates whether `initialPlay` has ever occurred. It's basically a way of knowing whether we're dealing with the first video on embed code changed.

js/controller.js
jpfranco: I modified `updatePlayhead` so that it uses the current values if they are not passed. The idea here is that we don't know the playhead at this point, so we only update the value that we have.

aldigun:
LGTM.

PR #854, opened 2017-10-05
Player 1788

aeng7:
Muted autoplay

-moved storing of oldVolume into the onVolumeChanged function from the handleMuteClick function


js/controller.js
```diff
@@ -1429,27 +1432,11 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
     },
 
     handleMuteClick: function() {
-      var newVolumeSettings = {};
       if (!this.state.volumeState.muted) {
-        //if we're muting, save the current volume so we can
-        //restore it when we un-mute
-        newVolumeSettings = {
-          oldVolume: this.state.volumeState.volume,
-          muted: !this.state.volumeState.muted
```
jpfranco: So this old handling of `muted` was redundant?

aeng7: Looks to be the case.

jpfranco:
LGTM

PR #861, opened 2017-10-06
[PBW-6804] Updating shrinkwrap version of deepmerge

jpfranco:
Without this, `npm install` will keep installing the old version of deepmerge


aldigun:
LGTM

PR #867, opened 2017-10-11
Merge branch feature/PLAYER-1350 into master - corrects for PR

KseniiaSokolovaEpam:


js/components/viewControlsVr.js
```diff
@@ -21,35 +22,34 @@ var ViewControlsVr = React.createClass({
...
       }
-
-
-      if (!this.props.controller.state.isPlayingAd) {
-        if (!this.isMobile) {
-          var desktopContent = this.props.skinConfig.buttons.desktopContent;
-          this.icon = _.find(desktopContent, function (el) {
-            return el.location === "mainView";
-          });
+      if (! (this.props.controller.state.isPlayingAd || this.isMobile)) {
```
IvanSakharovskii: remove space

js/components/viewControlsVr.js
```diff
@@ -21,35 +22,34 @@ var ViewControlsVr = React.createClass({
...
+        if (! (this.props.skinConfig && this.props.skinConfig.buttons && _.isArray(this.props.skinConfig.buttons.desktopContent))) {
```
IvanSakharovskii: same as above

js/components/viewControlsVr.js
```diff
@@ -21,35 +22,34 @@ var ViewControlsVr = React.createClass({
...
       }
-
-
-      if (!this.props.controller.state.isPlayingAd) {
-        if (!this.isMobile) {
-          var desktopContent = this.props.skinConfig.buttons.desktopContent;
-          this.icon = _.find(desktopContent, function (el) {
-            return el.location === "mainView";
-          });
+      if (!(this.props.controller.state.isPlayingAd || this.isMobile)) {
```
aldigun: I like this De Morgan law!

aldigun:
LGTM

scss/base/_variables.scss
```diff
@@ -131,6 +131,8 @@ $zindex-ad-top-bar:                       12600 !default;
 $zindex-popover:                          12600 !default;
 $SUPER_Z_INDEX:                           20000 !default; //in core, here for bird's eye view tracking
 $zindex-fullscreen:                  2147483647 !default; //in screenfull, here for bird's eye view tracking
+$zindex_1000:                              1000 !default;
+$zindex_1001:                              1001 !default;
```
pilievOoyala: While this is technically a constant, putting the value in the name doesn't seem useful. Also, all the other constants use '-' not underscore. Can you rename these to 
zindex-state-screen and zindex-vr-controls

pilievOoyala:
Everything looks good except the naming of the constants. Once thats in, I'll merge.

js/components/viewControlsVr.js
```diff
@@ -21,35 +22,34 @@ var ViewControlsVr = React.createClass({
...
+        if (!(this.props.skinConfig && this.props.skinConfig.buttons && _.isArray(this.props.skinConfig.buttons.desktopContent))) {
+          return;
         }
+        var desktopContent = this.props.skinConfig.buttons.desktopContent;
+        this.icon = _.find(desktopContent, function (el) {
```
jpfranco: I think that this project uses underscore as a dev dependency only, so including it might be increasing the file size somewhat. It's probably not a big deal, but the plugin receives an underscore reference during initialization, so there might be a way to reference that instead of requiring it. What do you guys think @pilievOoyala @aeng7 ?

js/skin.js
```diff
@@ -154,6 +163,8 @@ var Skin = React.createClass({
    */
 
   getDirectionParams: function(pageX, pageY) {
+    pageX = this.checkIsNumber(pageX) ? pageX : 0;
+    pageY = this.checkIsNumber(pageY) ? pageY : 0;
```
jpfranco: We have a `Utils.ensureNumber` function which does this (i.e. return a default value if the input is not a number). I recommend using this instead in order to avoid code duplication.

js/skin.js
```diff
@@ -182,6 +193,16 @@ var Skin = React.createClass({
...
 
+  /**
+   * @description check if the value is number
+   * @private
+   * @param value {number}
+   * @returns {boolean}
+   */
+  checkIsNumber: function(value) {
+    return ( !isNaN(parseFloat(value)) && isFinite(value) );
+  },
```
jpfranco: Same comment as above about `Utils.ensureNumber`

scss/base/_variables.scss
jpfranco: Can we use the element names like on the rest of the `zindex` vars?

scss/components/_state-screen.scss
```diff
@@ -30,7 +30,7 @@
     @extend .oo-expand;
     text-decoration:none;
     cursor: default;
-    z-index: 1000;
+    z-index: $zindex_1000;
```
jpfranco: I'm not sure if this has been fixed in a later version, but in my local environment this seems to block the click button of the Discovery Up Next panel. Please make sure to check whether this is the case.

screen shot 2017-10-12 at 12 28 32 pm

In order to see this you will need to add the Discovery plugin:
//player.ooyala.com/static/v4/candidate/latest/other-plugin/discovery_api.js

And use a video that has Discovery enabled, such as:

ec: l1MzhtNTE6eOf81reieCvXoynnA7TUdp
pbid: e4c115a4f4e74866817e1c7fa60dfc39
pcode: 9odGgyOhmvtSyMbOojUa85Ug4if_

js/components/viewControlsVr.js
KseniiaSokolovaEpam: :-)

js/skin.js

js/skin.js

scss/base/_variables.scss
KseniiaSokolovaEpam: /+

(ok, sure. But you may have a problem with logic of the names when you want to use the same values for z-index in other files.)

scss/components/_state-screen.scss
KseniiaSokolovaEpam: oh, thank you! We will create a new task with this issue.

js/components/viewControlsVr.js
aldigun: True, the less we need extra library, the better. We should utilize the provided underscore from plugin.

js/skin.js
```diff
@@ -152,8 +161,9 @@ var Skin = React.createClass({
    * @param pageY {number} y coordinate
    * @returns {[number, number, number]}
    */
-
   getDirectionParams: function(pageX, pageY) {
+    pageX = Utils.ensureNumber(pageX) ? pageX : 0;
```
jpfranco: The ternary operator is not necessary when using ` Utils.ensureNumber()`. The function takes a second parameter which is a default value to use when the input is not a number. This can be simplified like this: `Utils.ensureNumber(pageX, 0)`

js/components/viewControlsVr.js
yvinokurov: Connecting the library underscore is necessary for the passage of unit-tests. When executing the code, underscore ("_") is in the global scope, but when it runs tests it is not there.
We can write our solution for find in an array, but I think that this is not the best solution

pilievOoyala:
LGTM

js/components/viewControlsVr.js
jpfranco: Yes, we've been using it for unit tests, that part is fine. For the actual code it might be worthwhile to see if we can use the underscore reference from the global scope.

PR #870, opened 2017-10-16
[PLAYER-1679] Accessibility support for quality menu

jpfranco:
Added screen reader support and keyboard navigation for the video quality menu.

I will create a separate PR for the unit tests.


js/components/closeButton.js
```diff
@@ -1,14 +1,23 @@
...
-        <button className={this.props.cssClass}
-          onClick={this.props.closeAction}>
-          <Icon {...this.props} icon="dismiss"
-            className={this.props.className}/>
-        </button>
+      <button
+        className={this.props.cssClass}
+        onClick={this.props.closeAction}
+        onMouseUp={Utils.blurOnMouseUp}
+        data-focus-id={'close-' + Date.now()}
```
aeng7: Why do we need to add Date.now() to the data-focus-id?

jpfranco: This is in case we ever display more than one close button at a time, the id needs to be unique.

js/components/controlBar.js
```diff
@@ -150,12 +163,46 @@ var ControlBar = React.createClass({
...
+    if (this.props.controller.state.videoQualityOptions.showVideoQualityPopover === true) {
+      // Re-focus on quality button when closing the quality popover if the latter was
+      // originally opened with a key press.
+      if (params.restoreToggleButtonFocus &&
+          this.qualityMenuOpenedWithKeyboard &&
+          this.qualityBtnElement &&
+          typeof this.qualityBtnElement.focus === 'function') {
+        this.qualityBtnElement.focus();
+      }
+      this.qualityMenuOpenedWithKeyboard = false;
```
aeng7: What do you think of storing the focused element of the control bar when entering a submenu (like the quality options)? And we when close a submenu, restore the focused element, regardless of whether keyboard or mouse were used.

js/components/videoQualityPanel.js
```diff
@@ -24,53 +27,164 @@ var VideoQualityPanel = React.createClass({
...
+    var menuItemsList = [];
+    if (this.menuDomElement) {
+      menuItemsList = this.menuDomElement.querySelectorAll('[' + CONSTANTS.KEYBD_FOCUS_ID_ATTR + ']');
+    }
+    if (!menuItemsList.length) {
+      return;
+    }
+    // Since these elements aren't actually next to each other in the DOM, their position
+    // relative to one another is implied from their tab order, which should be the same as
+    // the one returned by querySelectorAll as long as tabindex is set to 0 (which should be the case).
```
aeng7: How possible would it be for this to cause un-intuitive UI? Like pressing left would move right, etc?

jpfranco: It's possible but you would have to arrange the elements in a weird way in the html or with css. This will navigate in the same order as it does when tabbing, so if the tab ordering is un-intuitive, the arrow keys will be awkward as well.

js/components/closeButton.js
aeng7: Just out of curiosity, what happens when the id's for the close button are not unique? Let's say the user changed their system clock between viewing two different close buttons, and now two close buttons have the same data-focus-id (however harder-than-winning-every-lottery-in-the-world this may be) ?

aeng7:
LGTM

js/components/closeButton.js
jpfranco: Nothing in this case since this is mostly used for control bar elements, but the idea is that you use it to store and find the element that needs to be refocused after the control bar is re-rendered, but if there are many elements with the same id you will likely get the wrong one when querying the DOM.

scss/components/_video-quality.scss
```diff
@@ -42,23 +44,32 @@
...
           outline: none;
-          opacity:1;
+          opacity: 1;
           @include vendor-prefixes(transition, opacity 0.3s ease);
 
           &:hover {
             opacity:0.6;
           }
 
+          &:focus {
```
pilievOoyala: Do we need spaces after the '&:' since you fixed then in the section above?

pilievOoyala:
LGTM. If you dont need to edit the spacing, @jpfranco then you can hit merge.

scss/components/_video-quality.scss
jpfranco: Nope, these are usually written without spaces

PR #871, opened 2017-10-16
[player-1897]

aeng7:
-skin will now listen to mute state events to change mute state rather than relying on volume values
-updated unit tests for mute states

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -115,7 +115,6 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       "volumeState": {
         "volume": 1,
         "muted": false,
-        "oldVolume": 1,
         "volumeSliderVisible": false
       },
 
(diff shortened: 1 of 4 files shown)


js/controller.js
```diff
@@ -464,19 +464,19 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
     },
 
     onVolumeChanged: function (event, newVolume) {
-      if (typeof this.state.volumeState.volume === "number") {
-        this.state.volumeState.oldVolume = this.state.volumeState.volume;
-      }
       if (newVolume <= 0) {
```
pilievOoyala: What happens if it's NaN or type of something else? What does this eval to?

aeng7: if newVolume is a type we aren't expecting, the state of the volume is going to be messed up since we are setting: `this.state.volumeState.volume = newVolume;` below. I believe this evaluates to false, though.

pilievOoyala: coo beans

pilievOoyala:
LGTM.

PR #873, opened 2017-10-18
[PLAYER-1270] Fix for uncaught exception when attempting to set playhead before skin

jpfranco:
Was missing some null checks.


pilievOoyala:
Super thumbs up!
LGTM

PR #876, opened 2017-10-23
Bugfix for support VR in playlists

IvanSakharovskii:


js/controller.js
```diff
@@ -307,6 +308,11 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       }
     },
 
+    onClearVideoType: function(event, params) {
```
jpfranco: I think this event might be redundant. Assuming that `VIDEO_VR` is fired after `SET_EMBED_CODE`, can we just use the `EMBED_CODE_CHANGED` handler to clear these values?

KseniiaSokolovaEpam: EMBED_CODE_CHANGED is called each time for each video in a playlist, but CLEAR_VIDEO_TYPE is published only when type of first video is not the same as type of second video.

jpfranco: Copying my comment from the mjolnir PR:

_I think clearing those values is trivial enough that it can be done on every `EMBED_CODE_CHANGED`. Adding a new event will not provide any significant performance gain but it will give us a new code path that might make things more complicated to debug. I also don't think it's a good idea to preserve state from a previous video, each new video should be treated as a separate entity._

If clearing the VR values early causes a visible transition, then I can see why we'd want to use a new event, but otherwise I think it might not be needed.

KseniiaSokolovaEpam: Coping from https://git.corp.ooyala.com/projects/PBW/repos/mjolnir/pull-requests/310/overview (top of the page):

Peter Iliev
We have a couple of things that need work here.
-Adding in the constant CLEAR_VIDEO_TYPE requires that this change is duplicated in html5-common. Then you have to update the subrepos for all repos that use html5-common. Otherwise, potentially the global variable could be overwritten.
-We would like you to remove all the the isPlaylist and playlistActive function params. There are two better options. Add a new message that the playlist plugin will send out, on the message bus, when it has received playlist data (maybe something like PLAYLIST_UPDATED and the message will have the playlist embedcodes). Then the video_controller can listen for this and set a flag that the plugins can check. Or you can wait for the HLS_VR sas param and we can refresh the UI every video and add in the VR buttons when necessary. But for the moment we would like to avoid adding in this param.

    Dmitrii Maslov
        CLEAR_VIDEO_TYPE was removed
        VIDEO_TYPE_CHANGED added to html5-common and is used instead of PlayListActive

pilievOoyala:
LGTM

PR #878, opened 2017-10-23
[PLAYER-2016] Adding type attribute to all buttons

jpfranco:
Buttons that don't have a `type` attribute will default to `type="submit"` in some browsers. This will cause issues if the player is contained within a form element.


pilievOoyala:
LGTM

PR #879, opened 2017-10-24
[PLAYER-1865] Updated fonts

IvanSakharovskii:


jpfranco:
@IvanSakharovskii looks good, thanks for the update.

PR #883, opened 2017-10-25
prevent initialPlay event from firing if initial play has already occ…

dallen42:
…urred.

This should prevent INITIAL_PLAY from being fired multiple times

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -1416,7 +1416,9 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
     togglePlayPause: function() {
       switch (this.state.playerState) {
         case CONSTANTS.STATE.START:
-          this.mb.publish(OO.EVENTS.INITIAL_PLAY, Date.now());
+          if (!this.state.isInitialPlay){
+            this.mb.publish(OO.EVENTS.INITIAL_PLAY, Date.now(), false);
+          }
           break;
         case CONSTANTS.STATE.END:
           if(Utils.isAndroid() || Utils.isIos()) {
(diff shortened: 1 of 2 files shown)


js/controller.js
```diff
@@ -1416,7 +1416,9 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
     togglePlayPause: function() {
       switch (this.state.playerState) {
         case CONSTANTS.STATE.START:
-          this.mb.publish(OO.EVENTS.INITIAL_PLAY, Date.now());
+          if (!this.state.isInitialPlay){
```
jpfranco: Since `this.state.isInitialPlay` is set on the event callback is it possible that a couple of `INITIAL_PLAY` events get fired before the value is updated? Probably not, but just wondering

jpfranco: @dallen42 if what I posted above is not an issue then the PR looks good. Would probably just add a unit test for this scenario.

dallen42: It might be possible, but I didn't observe it in practice. Without the if, I was able to generate multiple initialPlay events easily, but with the if I could not get any duplicates. Do you think we should set this.state.isInitialPlay here? My main concern with that would be if the callback somehow fails to get activated, in which case the other variables set there will not ever be set

jpfranco: @dallen42 oh ok, if you can't reproduce the issue by clicking quickly several times then it's probably fine. I agree that it's better not to set `this.state.isInitialPlay` until after the callback fires.

aeng7:
@dallen42 can we add unit tests? LGTM otherwise

js/controller.js
dallen42: @jpfranco added unit tests

aeng7:
LGTM

PR #884, opened 2017-10-25
Player 1788b

aeng7:
-skin muted autoplay changes
-unmute icon will appear on playing and ad screens
-we will now publish if the user initiated a mute state change
-fixed an issue where the UI could get out of sync with the video volume/mute state if we received a volume/mute state change event from a video other than the currently playing video
-fixed an issue where some variables were implicitly defined
-removed the exposeStaticApi global variable that was used for unit tests


pilievOoyala:
LGTM

js/components/volumeControls.js
```diff
@@ -8,6 +8,15 @@ var CONSTANTS = require('../constants/constants');
 
 var VolumeControls = React.createClass({
 
+  volumeChange: function(vol) {
+    var newVol = Utils.ensureNumber(vol);
```
jpfranco: I think this actually returns `null` when it's not a valid number, but you can pass a default value to return for these cases in the second parameter

js/components/volumeControls.js
```diff
@@ -184,8 +192,7 @@ VolumeControls.propTypes = {
       isMobile: React.PropTypes.bool.isRequired,
       volumeState: React.PropTypes.shape({
         volumeSliderVisible: React.PropTypes.bool.isRequired,
-        volume: React.PropTypes.number.isRequired,
-        muted: React.PropTypes.bool.isRequired
```
jpfranco: I think this `muted` prop is still being used above

js/controller.js
```diff
@@ -44,6 +44,8 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       "isLiveStream": false,
       "screenToShow": null,
       "playerState": null,
+      "isPlaying": false,
```
jpfranco: I think we already keep track of this with `this.state.playerState`

aeng7: `playerState` has a `CONSTANTS.STATE.PAUSE` state, as well as other states. I want to keep track of when we are actively showing a video, paused or playing. Is there something existing I can use? Should I rename this?

js/views/adScreen.js
```diff
@@ -187,6 +202,13 @@ var AdScreen = React.createClass({
           {playbackControlItems}
         </div>
 
+        {showUnmute ? <button className="oo-ad-screen oo-unmute"
+          onClick={this.unmuteClick}
```
jpfranco: Can we add `type="button"`? If we want this to be accessible this will also need `tabindex=0`

js/controller.js
aeng7: Actually I'll try to use `playerState.` It might work out.

js/views/playingScreen.js
```diff
@@ -229,6 +246,12 @@ var PlayingScreen = React.createClass({
           playerState={this.props.playerState}
           isLiveStream={this.props.isLiveStream} />
       </div>
+      {showUnmute ? <button className="oo-playing-screen oo-unmute"
```
jpfranco: Same comment about attributes like the one above

js/components/volumeControls.js
aeng7: Done.

js/components/volumeControls.js
aeng7: Done.

js/views/adScreen.js
aeng7: Done.

js/views/playingScreen.js
aeng7: Done.

js/controller.js
aeng7: Done.

js/views/playingScreen.js
```diff
@@ -229,6 +246,12 @@ var PlayingScreen = React.createClass({
...
+        onClick={this.unmuteClick}
+        aria-label={volumeAriaLabel}
+        >
+        <Icon {...this.props} icon={volumeIcon} ref="volumeIcon" />
+      </button> : null}
```
jpfranco: Not needed currently, but this button would probably work better as a separate component once David's updates are added.

aeng7: Yep, agreed.

js/views/adScreen.js
```diff
@@ -187,6 +202,15 @@ var AdScreen = React.createClass({
           {playbackControlItems}
         </div>
 
+        {showUnmute ? <button className="oo-ad-screen oo-unmute"
+          type="button"
+          tabindex="0"
```
jpfranco: Sorry, my bad. I think this actually needs to be `tabIndex` with capital i for React

js/views/playingScreen.js
```diff
@@ -229,6 +246,14 @@ var PlayingScreen = React.createClass({
           playerState={this.props.playerState}
           isLiveStream={this.props.isLiveStream} />
       </div>
+      {showUnmute ? <button className="oo-playing-screen oo-unmute"
+        type="button"
+        tabindex="0"
```
jpfranco: Needs capital I as well

js/views/adScreen.js
aeng7: Done

js/views/playingScreen.js
aeng7: Done

PR #888, opened 2017-10-30
[Feature/player-1350] Bugfix for video 360

IvanSakharovskii:


scss/components/_vr-controls.scss
```diff
@@ -1,21 +1,21 @@
 .oo-vr-icon {
   &--move {
     position: absolute;
-    height: 15px;
-    width: 15px;
+    height: calc(#{$vr-controls-full-size}/4);
```
pilievOoyala: Shouldn't we just adjust the size of the icon instead of this resizing?

pilievOoyala: Am I undestanding this correctly? When the camera is stationary, the icons for vr are bigger. When you are moving they shrink.  Is that correct?

Also, can you divide by a constant instead of 4. That way we can adjust it easily if we decide to change the size.  You can do it in scss/base/_variables.scss where you have vr-controls-full-size defined.

yvinokurov: fix

js/components/accessibilityControls.js
```diff
@@ -131,6 +131,10 @@ AccessibilityControls.prototype = {
    * @returns {boolean} true if moved
    */
   moveVrToDirection: function(e, charCode, isKeyDown, targetTagName) {
+    var keyDirectionMap = this.keyDirectionMap;
+    if (!(keyDirectionMap[charCode] || targetTagName !== "button")) {
```
aldigun: null check on keyDirectionMap. What if it is null, what if it is an empty array.

scss/responsive/_large.scss
```diff
@@ -151,21 +151,22 @@
...
-      line-height: normal;
+      font-size: $vr-controls-size-max;
     }
-    .oo-vr-icon--substrate {
-      height: $vr-controls-full-size * $large-multiple + 4px;
-      width: $vr-controls-full-size * $large-multiple + 4px;
+
+    .oo-vr-icon--move{
+      height: calc(#{$vr-controls-size-max}/4);
+      width: calc(#{$vr-controls-size-max}/4);
```
pilievOoyala: Can we replace this with vr-controls-move-max?

PR #890, opened 2017-10-30
Player 1788b

aeng7:
-animated unmute icon

### DIFF
--- a/js/components/unmuteIcon.js
+++ b/js/components/unmuteIcon.js
@@ -14,10 +14,6 @@ var UnmuteIcon = React.createClass({
     this.props.controller.handleMuteClick();
   },
 
-  //componentWillMount: function() {
-  //
-  //},
-
   componentDidMount: function() {
     if (this.state.expanded) {
       clearTimeout(this.timeout);
@@ -30,13 +26,13 @@ var UnmuteIcon = React.createClass({
       this.timeout = setTimeout(_.bind(function() {
         me.props.controller.state.volumeState.unmuteIconCollapsed = true;
         me.setState({expanded: false});
-      }, this), 5000);
+      }, this), 2000);
     }
   },
 
-  //componentWillUnmount: function() {
-  //
-  //},
+  componentWillUnmount: function() {
+    clearTimeout(this.timeout);
+  },
 
   render: function() {
     var volumeIcon, volumeAriaLabel;
@@ -49,13 +45,9 @@ var UnmuteIcon = React.createClass({
     }
 
     var myClass = ClassNames({
-      'oo-unmute': !this.state.expanded,
-      'oo-unmute-expanded': this.state.expanded
+      'oo-unmute': true,
+      'oo-expanded': this.state.expanded
     });
-
-    //<div className="oo-unmute-inner-wrapper">
-    //
-    //    </div>
 
     return (
       <button className={myClass}
--- a/js/components/unmuteIcon.js
+++ b/js/components/unmuteIcon.js
@@ -6,6 +6,7 @@ var React = require('react'),
 var UnmuteIcon = React.createClass({
   getInitialState: function () {
     return {
+      collapseTime: 2000,
       expanded: !this.props.controller.state.volumeState.unmuteIconCollapsed
     };
   },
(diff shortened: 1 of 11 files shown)


js/components/unmuteIcon.js
```diff
@@ -0,0 +1,66 @@
...
+        type="button"
+        tabIndex="0"
+        aria-label={volumeAriaLabel}
+        >
+
+        <div className="oo-unmute-icon-wrapper">
+            <Icon {...this.props} icon={volumeIcon} ref="volumeIcon"/>
+          </div>
+
+        {this.state.expanded ? <div className="oo-unmute-message">SELECT TO UNMUTE</div> : null}
```
jpfranco: Can we add this text to the `SKIN_TEXT` constants?

js/components/unmuteIcon.js
```diff
@@ -0,0 +1,66 @@
...
+
+      </button>
+    );
+  }
```
jpfranco: Not super important, but recommend adding proptypes

scss/base/_variables.scss
```diff
@@ -134,7 +134,7 @@ $zindex-fullscreen:                  2147483647 !default; //in screenfull, here
 $zindex-state-screen:                      1000 !default;
 $zindex-vr-controls:                       1001 !default;
 $zindex-up-next-panel:                     1002 !default;
-$zindex-unmute:                           13000 !default;
+$zindex-unmute:                      2147483647 !default;
```
jpfranco: Is the button inside a container with a lower z-index? If not then I think this large value might make it hard for customers to have elements that are on top of the player, like navigation bars, etc.

tests/components/unmuteIcon-test.js
```diff
@@ -0,0 +1,135 @@
...
+var ReactDOM = require('react-dom');
+var TestUtils = require('react-addons-test-utils');
+var UnmuteIcon = require('../../js/components/unmuteIcon');
+var CONSTANTS = require('../../js/constants/constants');
+var skinConfig = require('../../config/skin.json');
+var sinon = require('sinon');
+
+describe('UnmuteIcon', function () {
+  it('creates an expanded UnmuteIcon', function () {
+    var mockController = {
```
jpfranco: Also not super important, but this could probably be defined in `beforeEach` in order to avoid defining it every time.

scss/base/_variables.scss
aeng7: It's inside the playing screen and ad screen containers, which don't appear to have defined z indices. I've had some IMA ads have click priority over the mute icon, so we're in a tough spot to decide what this z-index should be. I can change this to 19999 to be right under our "SUPER_Z_INDEX" for the time being.

jpfranco: Oh ok I see, sounds good.

tests/components/unmuteIcon-test.js
aeng7: Done

js/components/unmuteIcon.js
aeng7: Done

js/components/unmuteIcon.js
aeng7: Done

pilievOoyala:
LGTM

js/components/unmuteIcon.js
```diff
@@ -0,0 +1,77 @@
+var React = require('react'),
+    Icon = require('./icon'),
+    ClassNames = require('classnames'),
+    CONSTANTS = require('../constants/constants');
+
+var UnmuteIcon = React.createClass({
+  getInitialState: function () {
+    return {
+      collapseTime: 2000,
+      expanded: !this.props.controller.state.volumeState.unmuteIconCollapsed
```
jpfranco: Sorry, one last thing I forgot to mention. Usually it's not recommended to assign values from props to local state. We can probably avoid the local state by using `unmuteIconCollapsed` directly when rendering. We can probably change this later though

aeng7: I'll make this change now.

aeng7: Done

PR #893, opened 2017-10-31
[PLAYER-1992] Passing isFullScreen parameter when firing SET_CLOSED_CAPTIONS_LANGUAGE event

jpfranco:
This should be merged together with the following PR:
https://git.corp.ooyala.com/projects/PBW/repos/mjolnir/pull-requests/326/overview

The skin triggers `SET_CLOSED_CAPTIONS_LANGUAGE` every time the video resumes playback, so this was causing closed captions to disappear on fullscreen iOS when pausing and then resuming. The core uses the `isFullScreen` parameter to determine whether to show the native closed captions on iOS or not, so this state was being lost.


pilievOoyala:
LGTM

PR #894, opened 2017-11-01
[player-2063]

aeng7:
-fixed an issue where the volume icon and aria labels were not showing the volumeOff icon when volume was set to 0

### DIFF
--- a/tests/components/controlBar-test.js
+++ b/tests/components/controlBar-test.js
@@ -372,26 +372,12 @@ describe('ControlBar', function () {
   });
 
   it('should display unmute volume icon when volume is set to a non-zero value', function() {
-    var mockController = {
-      state: {
-        isMobile: false,
-        volumeState: {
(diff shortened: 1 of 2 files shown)


tests/components/controlBar-test.js
```diff
@@ -371,6 +371,108 @@ describe('ControlBar', function () {
     expect(newVolume).toBeGreaterThan(-1);
   });
 
+  it('should display unmute volume icon when volume is set to a non-zero value', function() {
+    var mockController = {
```
jpfranco: Can we use `baseMockController` here so that we don't repeat this each time?

aeng7: Yep, I'll make this change

aeng7: Done

pilievOoyala:
LGTM

PR #900, opened 2017-11-02
[PLAYER-2120] Fix for quality menu getting keyboard focus sometimes when using mouse

jpfranco:
Also took the opportunity to fix the gray shadow on the un-mute button.

### DIFF
--- a/js/components/controlBar.js
+++ b/js/components/controlBar.js
@@ -187,6 +187,9 @@ var ControlBar = React.createClass({
   },
 
   toggleQualityPopover: function () {
+    if (this.props.controller.state.videoQualityOptions.showVideoQualityPopover) {
+      this.qualityMenuOpenedWithKeyboard = false;
+    }
     this.props.controller.toggleVideoQualityPopOver();
   },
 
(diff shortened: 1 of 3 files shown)


aeng7:
@jpfranco can we add a unit test for this? LGTM otherwise

jpfranco:
@aeng7 sure, added.

js/components/controlBar.js
```diff
@@ -202,7 +205,6 @@ var ControlBar = React.createClass({
           typeof this.qualityBtnElement.focus === 'function') {
         this.qualityBtnElement.focus();
       }
-      this.qualityMenuOpenedWithKeyboard = false;
```
pilievOoyala: Shouldn't we keep this here ?

aeng7:
LGTM

js/components/controlBar.js
jpfranco: Nope, this was moved to `toggleQualityPopover` which is called below.

pilievOoyala:
LGTM

PR #901, opened 2017-11-03
[player-2086]

aeng7:
-fixed an issue where muting/unmuting via skin was not notifying all plugins to do so
-improved unit tests on this


pilievOoyala:
LGTM

PR #902, opened 2017-11-06
[player-2144]

aeng7:
-skin will set currentVideoId earlier on vc play rather than vc playing. This will allow initial volume to pass through

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -1349,6 +1349,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
 
     unsubscribeBasicPlaybackEvents: function() {
       this.mb.unsubscribe(OO.EVENTS.INITIAL_PLAY, 'customerUi');
+      this.mb.unsubscribe(OO.EVENTS.VC_PLAY, 'customerUi');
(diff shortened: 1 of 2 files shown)


js/controller.js
```diff
@@ -188,6 +188,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       if(!this.state.isSubscribed) {
         this.mb.subscribe(OO.EVENTS.SEND_QUALITY_CHANGE, 'customerUi', _.bind(this.receiveVideoQualityChangeEvent, this));
         this.mb.subscribe(OO.EVENTS.INITIAL_PLAY, 'customerUi', _.bind(this.onInitialPlay, this));
+        this.mb.subscribe(OO.EVENTS.VC_PLAY, 'customerUi', _.bind(this.onVcPlay, this));
```
jpfranco: Does this need to be added to `unsubscribeBasicPlaybackEvents`?

aeng7: Yep, adding in

aeng7: Done

pilievOoyala:
LGTM

PR #904, opened 2017-11-07
fix unit tests for PLAYER-2006

dallen42:


dallen42:
My unit test for PLAYER-2006 was failing in master, this should fix it so all unit tests are passing w/ latest skin-conifg subrepo

PR #905, opened 2017-11-08
update countdown text to have singular nouns if amount is one

dallen42:


js/components/utils.js
```diff
@@ -418,14 +418,20 @@ var Utils = {
   * @returns {String} The countdown time string
   */
   getStartCountdown: function(countdownTimestamp) {
+    var dayString = "day";
+    var hourString = "hour";
+    var minuteString = "minute";
     try {
       if (countdownTimestamp < 0) return "";
       var days = Math.floor(countdownTimestamp / (24 * 60 * 60 * 1000));
+      if (days != 1) dayString += "s";
```
pilievOoyala: Can you put these on separate lines and include the {}?

pilievOoyala: It's easier to debug in case there is something wrong with the 'then' part

pilievOoyala:
LGTM except for the formatting.

aeng7:
@dallen42 , we'll merge this to get it in the build. Can you add some unit tests and address Peter's comment in a future PR? Thanks!

js/components/utils.js
```diff
@@ -418,14 +418,20 @@ var Utils = {
   * @returns {String} The countdown time string
   */
   getStartCountdown: function(countdownTimestamp) {
+    var dayString = "day";
+    var hourString = "hour";
+    var minuteString = "minute";
     try {
       if (countdownTimestamp < 0) return "";
       var days = Math.floor(countdownTimestamp / (24 * 60 * 60 * 1000));
```
rcvrgs: for code maintainability, it would be better if you store the divisor day component in a variable and reuse it wherever you need it. Like 
```
var dayDivisor =  24 * 60 * 60 * 1000;
var days = Math.floor(countdownTimestamp / dayDivisor);
countdownTimestamp -= days * dayDivisor;
```

Same applies for the hours and minutes if necessary.

PR #906, opened 2017-11-10
[Feature/player-2161] Update Node JS

yvinokurov:


package.json
```diff
@@ -29,33 +29,33 @@
     "screenfull": "3.0.0"
   },
   "devDependencies": {
-    "browserify": "13.0.0",
-    "bulk-require": "^1.0.0",
+    "browserify": "14.5.0",
+    "bulk-require": "^1.0.1",
```
pilievOoyala: I dont remember what the different symbols like '^' mean but feel free to change it to just the version number if it will be safer. So we don't run the risk of a build not working because it installs a version that is slightly different.

jpfranco:
This was implemented in https://github.com/ooyala/html5-skin/pull/1085, closing.

PR #909, opened 2017-11-14
[PBW-6952] Fix for skin container element not being created when multiple players exist

jpfranco:


pilievOoyala:
LGTM

js/controller.js
```diff
@@ -288,7 +288,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       // identified with the ARIA label above. We set it to -1 in order to prevent actual keyboard focus
       this.state.mainVideoInnerWrapper.attr('tabindex', '-1');
 
-      if (!$('.oo-player-skin').length) {
+      if (!this.state.mainVideoInnerWrapper.children('.oo-player-skin').length) {
```
pilievOoyala: Do we need to check that we got anything or does the .children method always return an array? (Asking in case it doesn't find anything)

aeng7: Are there any other instances in our skin where they do the same thing (use jQuery to find an element)?

aeng7:
LGTM

js/controller.js
jpfranco: @pilievOoyala yes this will return an empty array when no children are found.

@aeng7  we don't use it specifically to find an element, but jquery is used in several places in the controller. `this.state.mainVideoInnerWrapper` is a jquery element so I thought was better not to get the dom node and then search

PR #911, opened 2017-11-15
Feature/player 1350

IvanSakharovskii:


pilievOoyala:
Please double check why we are removing from the language files. I think someone had an outdated version when they pushed changes.

js/components/thumbnail.js
```diff
@@ -7,12 +7,107 @@ var React = require('react'),
...
+  componentWillReceiveProps: function(nextProps) {
+    if (this.props.vrViewingDirection !== nextProps.vrViewingDirection) {
+      var yaw = nextProps.vrViewingDirection.yaw;
+      var pitch = nextProps.vrViewingDirection.pitch;
+      this.setCurrentViewVr(yaw, pitch);
+    }
+  },
   shouldComponentUpdate: function(nextProps) {
-    return (nextProps.hoverPosition != this.props.hoverPosition);
+    var hoverPosition = nextProps.hoverPosition != this.props.hoverPosition;
```
pilievOoyala: Can you rename these variable to updateHoverPositon, updateFullscreen, updateVrViewDIrection. It's a little bit clearer.

js/components/thumbnail.js
```diff
@@ -7,12 +7,107 @@ var React = require('react'),
     Utils = require('./utils');
 
 var Thumbnail = React.createClass({
+  getInitialState: function() {
+    this.positionY = 0;
```
aldigun: Where are these numbers coming from?

js/components/thumbnail.js
```diff
@@ -7,12 +7,107 @@ var React = require('react'),
...
+    pitch = pitch >= 360 ? 0 : pitch;
+
+    var positionY = -(((imageHeight - thumbnailHeight) / 2) - pitch);
+    var bottomCoordinate = -(imageHeight - thumbnailHeight);
+    if (positionY > 0) {
+      positionY = 0;
+    } else if (positionY < bottomCoordinate) {
+      positionY = bottomCoordinate;
+    }
+    var positionX = -(imageWidth - thumbnailWidth/2 - imageWidth*yaw/360);
```
pilievOoyala: Can you fix the spacing on imageWidth*yaw/360 ?

js/components/thumbnail.js
```diff
@@ -7,12 +7,107 @@ var React = require('react'),
...
+
+  /**
+   * @description return current coefficient of the yaw if yaw > 360 or yaw < -360 degrees
+   * @param {Number} yaw - angle in degrees
+   * @private
+   * @returns {number} coefficient showing how many times to take 360 degrees
+   */
+  getCurrentYawVr: function(yaw) {
+    var k = yaw <= -360 ? -1 : 1;
+    var ratio = (k*yaw)/360;
```
pilievOoyala: fix the spacing.

js/components/thumbnail.js
```diff
@@ -7,12 +7,107 @@ var React = require('react'),
...
+    ratio = ~~ratio;
+    return (yaw - k*ratio*360);
```
pilievOoyala: can you fix the spacing? And stick the result in a variable. Then return the variable. It's easier to debug the math if there is a problem.

js/components/thumbnail.js
```diff
@@ -7,12 +7,107 @@ var React = require('react'),
...
+    }
+    if (thumbnailHeight) {
+      this.thumbnailHeight = thumbnailHeight;
+    }
+  },
+
+  setImageSizes: function() {
+    var thumbnail = Utils.findThumbnail(this.props.thumbnails, this.props.hoverTime, this.props.duration, this.props.videoVr);
+    this.imageWidth = thumbnail.imageWidth;
+    this.imageHeight = thumbnail.imageHeight;
```
aldigun: (line 64-65) I think thumbnail size should not be dictated by the actual image's size. We do not want the thumbnail to blow up or shrink.

js/components/utils.js
```diff
@@ -579,12 +579,20 @@ var Utils = {
...
+  * @param {Boolean} isVideoVr - if video is vr
   * @returns {Object} object that contains URL and index of requested thumbnail
   */
-  findThumbnail: function(thumbnails, hoverTime, duration) {
+  findThumbnail: function(thumbnails, hoverTime, duration, isVideoVr) {
     var timeSlices = thumbnails.data.available_time_slices;
     var width = thumbnails.data.available_widths[0]; //choosing the lowest size
 
+    if (isVideoVr) {
+      // it is necessary to take bigger image for showing part of the image
```
aldigun: Why bigger image?
- Player will have considerable performance downgrade due to downloading bigger size image
- I do not want the thumbnail size to blow up (or shrink)
- Has this been discussed with Steve? I disagree with the usage of bigger images.

js/components/thumbnail.js
```diff
@@ -7,12 +7,107 @@ var React = require('react'),
     Utils = require('./utils');
 
 var Thumbnail = React.createClass({
+  getInitialState: function() {
+    this.positionY = 0;
+    this.positionX = -320;
+    this.imageWidth = 0;
```
aldigun: What is the difference between imageWidth and thumbnailWidth?

js/components/thumbnail.js
KseniiaSokolovaEpam: \+

js/components/thumbnail.js
KseniiaSokolovaEpam: It is default value for ideal situation.

js/components/thumbnail.js
KseniiaSokolovaEpam: \+

js/components/thumbnail.js
KseniiaSokolovaEpam: \+

js/components/thumbnail.js
KseniiaSokolovaEpam: \+

js/components/thumbnail.js
KseniiaSokolovaEpam: Yes, we should not for 2d, but for 360 we have to do it.

js/components/thumbnail.js
KseniiaSokolovaEpam: thumbnailWidth is width of thumbnail,
imageWidth is width of image for thumbnail, so if this image is image for 360, i.e. sphire, the width can be bigger than thumbnailWidth

js/components/accessibilityControls.js
```diff
@@ -147,11 +145,30 @@ AccessibilityControls.prototype = {
     }
     this.vrRotationAllowed = !isKeyDown; //prevent repeat of keyDown
     this.controller.moveVrToDirection(false, keyDirectionMap[charCode]); //stop rotation if isKeyDown === false or prevent prev rotation if press a button (isKeyDown === true)
+
+    var inPrevKeyPressedArrIndex = -1;
```
aeng7: It seems like this index and the for loop before is only used if isKeyDown is false in the if statement directly below. If this is the case, can we move this block of code into the if statement?

js/components/thumbnail.js
```diff
@@ -4,15 +4,126 @@
  * @module Thumbnail
  */
 var React = require('react'),
-    Utils = require('./utils');
+    Utils = require('./utils'),
+    ReactDOM = require('react-dom');
 
 var Thumbnail = React.createClass({
+  getInitialState: function() {
```
aeng7: Why were the initial values for the initial state chosen as such? Like, why does positionX start at -320, thumbnail width start at 80, etc?

js/components/thumbnail.js
```diff
@@ -4,15 +4,126 @@
...
+      var yaw = nextProps.vrViewingDirection.yaw;
+      var pitch = nextProps.vrViewingDirection.pitch;
+      this.setCurrentViewVr(yaw, pitch);
+    }
+  },
   shouldComponentUpdate: function(nextProps) {
-    return (nextProps.hoverPosition != this.props.hoverPosition);
+    var updateHoverPositon = nextProps.hoverPosition != this.props.hoverPosition;
+    var updateFullscreen  = nextProps.fullscreen != this.props.fullscreen && this.props.videoVr;
+    var updateVrViewDIrection = nextProps.vrViewingDirection != this.props.vrViewingDirection;
```
aeng7: The I in direction is capitalized. It should be lower case.

js/components/thumbnail.js
```diff
@@ -4,15 +4,126 @@
...
+          imageWidth = 380;
+          imageHeight = thumbnail.imageHeight * 380 / thumbnail.imageWidth;
+        }
+        this.imageWidth = imageWidth;
+        this.imageHeight = imageHeight;
+      }
+    }
+  },
+
+  /**x
```
aeng7: Extra 'x'

js/components/accessibilityControls.js
```diff
@@ -147,11 +145,30 @@ AccessibilityControls.prototype = {
...
+      if (this.prevKeyPressedArr[i] === charCode) {
+        inPrevKeyPressedArrIndex = i;
+        break;
+      }
+    }
     if (isKeyDown === true) {
-      this.controller.moveVrToDirection(true, keyDirectionMap[charCode]);
-      return true;
+      this.prevKeyPressedArr.push(charCode);
+    } else { // if button is up remove a code of the button from this.prevKeyPressedArr
```
pilievOoyala: Just a minor fix, can you change the wording to this:
// if button is up, remove it from this.prevKeyPressedArr

js/components/accessibilityControls.js
```diff
@@ -147,11 +145,30 @@ AccessibilityControls.prototype = {
...
+        this.prevKeyPressedArr.splice(inPrevKeyPressedArrIndex, 1);
+      }
+    }
+    if (this.prevKeyPressedArr.length) {
+      isKeyDown = true;
+      charCode = this.prevKeyPressedArr[this.prevKeyPressedArr.length-1];
     }
-    return false;
+    //rotate if a button is pressed, stop rotate if other case
+    this.controller.moveVrToDirection(isKeyDown, keyDirectionMap[charCode]);
```
pilievOoyala: So we are only rotating based on what the last key pressed is?  You can't move in two directions at once? Otherwise why keep an array of what is pressed. You could just keep track of what the last key pressed is.

js/components/thumbnail.js
```diff
@@ -4,15 +4,126 @@
...
+  /**
+   * @description return current coefficient of the yaw if yaw > 360 or yaw < -360 degrees
+   * @param {Number} yaw - angle in degrees
+   * @private
+   * @returns {number} coefficient showing how many times to take 360 degrees
+   */
+  getCurrentYawVr: function(yaw) {
+    var k = yaw <= -360 ? -1 : 1;
+    var ratio = k * yaw / 360;
+    ratio = ~~ratio;
```
aeng7: How come we want to floor the ratio?

js/components/utils.js
```diff
@@ -579,11 +579,18 @@ var Utils = {
...
   * @returns {Object} object that contains URL and index of requested thumbnail
   */
-  findThumbnail: function(thumbnails, hoverTime, duration) {
+  findThumbnail: function(thumbnails, hoverTime, duration, isVideoVr) {
     var timeSlices = thumbnails.data.available_time_slices;
     var width = thumbnails.data.available_widths[0]; //choosing the lowest size
+    if (isVideoVr && width < 380) {
+      // it is necessary to take bigger image for showing part of the image
+      // so choose not the lowest size but bigger one, the best width is 380
+      var index = thumbnails.data.available_widths.length >= 5 ? 4 : thumbnails.data.available_widths.length - 1;
```
aeng7: How did we choose the value 5 for:

thumbnails.data.available_widths.length >= 5

tests/components/thumbnail-test.js
```diff
@@ -142,4 +142,27 @@ describe('Thumbnail', function () {
       expect(thumbnailTime.length).toBe(1);
     }
   });
+
+  it('tests functions for vr preview', function () {
```
aeng7: Does this unit test test all the changes we've made above? It seems like there should be more unit tests. Please add some more coverage if needed. Thanks!

js/components/thumbnail.js
pilievOoyala: This truncates the value to remove any decimal, correct?  Why do we need to do that?

js/components/thumbnail.js
```diff
@@ -4,15 +4,126 @@
...
+    }
+  },
+
+  setImageSizes: function() {
+    var thumbnail = Utils.findThumbnail(this.props.thumbnails, this.props.hoverTime, this.props.duration, this.props.videoVr);
+    if (thumbnail !== null && typeof thumbnail === 'object') {
+      var imageWidth = thumbnail.imageWidth;
+      var imageHeight = thumbnail.imageHeight;
+      if (imageWidth && imageHeight) {
+        if (imageWidth > 380) {
```
pilievOoyala: 380 seems to be a magic number you use in another function as well. Please make this a constant.

js/components/utils.js
```diff
@@ -579,11 +579,18 @@ var Utils = {
...
   * @param {Number} hoverTime - time value to find thumbnail for
   * @param {Number} duration - duration of the video
+  * @param {Boolean} isVideoVr - if video is vr
   * @returns {Object} object that contains URL and index of requested thumbnail
   */
-  findThumbnail: function(thumbnails, hoverTime, duration) {
+  findThumbnail: function(thumbnails, hoverTime, duration, isVideoVr) {
     var timeSlices = thumbnails.data.available_time_slices;
     var width = thumbnails.data.available_widths[0]; //choosing the lowest size
+    if (isVideoVr && width < 380) {
```
pilievOoyala: this is the other place 380 is used.

js/components/accessibilityControls.js
KseniiaSokolovaEpam: \+

js/components/thumbnail.js
KseniiaSokolovaEpam: \+

js/components/thumbnail.js
KseniiaSokolovaEpam: \+

js/components/accessibilityControls.js
KseniiaSokolovaEpam: \+

js/components/accessibilityControls.js
KseniiaSokolovaEpam: Yes, but situations when a user pressed several buttons in the one time are happened, and when the user unpressed each of the buttons we need to keep rotation in current direction. For example, an order of pressing is D & A & W, so firstly we get right rotation, after that we get left rotation, an in the end we get top rotation. If the user unpressed A and after that unpressed W, so we need to rotate to right, but right rotation is not last. So we need to save last 0-4 rotations for the situation.

js/components/thumbnail.js
KseniiaSokolovaEpam: bitmovin can  return a value like 764, but we need only value from 0 to 360. So, in example 764 is equal 44 : 764 - 2*360, in this case 2 is ratio. And ratio is the largest integer divisor.

KseniiaSokolovaEpam: bitmovin can  return a value like 764, but we need only value from 0 to 360. So, in example 764 is equal 44 : 764 - 2*360, in this case 2 is ratio. And ratio is the largest integer divisor.

js/components/thumbnail.js
KseniiaSokolovaEpam: \+

js/components/utils.js
KseniiaSokolovaEpam: \+

js/components/utils.js
KseniiaSokolovaEpam: We have 2 sizes for thumbnails width: 80px for small screens and 119px for other. For 360 image we need to get only part of the server image  and show this part. So we need to find the best proportions between thumbnails sizes and images sizes. Usually thumbnailWidth * 4 is enought for show a most viewing part of the image.

tests/components/thumbnail-test.js
KseniiaSokolovaEpam: Ok, thank you, but in next PR a code for 360 thumbnail will be changed. So I add more tests in next PR.

js/components/thumbnail.js
KseniiaSokolovaEpam: I got values for small thumbnails (width is 80px) (this variables (what you are asking about) are used only for 360 thumbnails), and if we get ideal sizes of images and thumbnail, we have -320 px for x-axis offset. This variables are changed after counting for real size of thumbnails and bg images.

js/components/accessibilityControls.js
pilievOoyala: Sounds good. Thank you for the explanation.

js/components/utils.js
```diff
@@ -579,11 +579,18 @@ var Utils = {
...
+  * @param {Boolean} isVideoVr - if video is vr
   * @returns {Object} object that contains URL and index of requested thumbnail
   */
-  findThumbnail: function(thumbnails, hoverTime, duration) {
+  findThumbnail: function(thumbnails, hoverTime, duration, isVideoVr) {
     var timeSlices = thumbnails.data.available_time_slices;
     var width = thumbnails.data.available_widths[0]; //choosing the lowest size
+    if (isVideoVr && width < CONSTANTS.THUMBNAIL.MAX_VR_THUMBNAIL_BG_WIDTH) {
+      // it is necessary to take bigger image for showing part of the image
+      // so choose not the lowest size but bigger one, the best width is 380
```
pilievOoyala: Should we change this comment because we have a constant instead of the number 380?

pilievOoyala:
LGTM

js/components/utils.js
```diff
@@ -579,11 +579,18 @@ var Utils = {
...
+      var index = thumbnails.data.available_widths.length >= 5 ? 4 : thumbnails.data.available_widths.length - 1;
```
aldigun: Where are the numbers 5 and 4 coming from?
Perhaps you need to go through the width array, then use the closes to MAX_VR_THUMBNAIL_BG_WIDTH?

pilievOoyala: Should 5 and 4 be constants. It's not clear what they mean.

PR #914, opened 2017-11-16
[PLAYER-1678] CC menu accessibility (PR #1)

jpfranco:
Included in this PR:
- Refactored popover display logic in order to reduce duplicate code
- Added AccessibleButton component and replaced several of the existing buttons with it
- Added AccessibleMenu component
- Enabled accessibility for the CC popover (small window)

Still pending:
- All of the other CC menus


js/components/accessibleButton.js
```diff
@@ -0,0 +1,92 @@
...
+  role: React.PropTypes.string,
+  onClick: React.PropTypes.func
+};
+
+AccessibleButton.defaultProps = {
+  focusId: Math.random().toString(36).substr(2, 10),
+  ariaChecked: null,
+  ariaHasPopup: null,
+  ariaExpanded: null,
+  role: null,
```
aeng7: What is role used for?

js/components/controlBar.js
```diff
@@ -146,77 +146,80 @@ var ControlBar = React.createClass({
...
+      this.togglePopover(CONSTANTS.MENU_OPTIONS.CLOSED_CAPTIONS);
+      this.closeQualityPopover();
     }
   },
 
-  toggleQualityPopover: function () {
-    if (this.props.controller.state.videoQualityOptions.showVideoQualityPopover) {
-      this.qualityMenuOpenedWithKeyboard = false;
+  configureMenuAutofocus: function(menu) {
+    var menuOptions = this.props.controller.state[menu] || {};
```
aeng7: Should we just return if this.props.controller.state[menu] is falsy? I don't see anything below impact anything outside the scope of this function.

js/components/accessibleButton.js
jpfranco: For screen readers to know whether the element is being used as button, checkbox, menu item, etc. Buttons have implicit roles of "button", but in this case we also use them as menu items.

js/components/controlBar.js
jpfranco: Yes I actually changed it in order to save a few lines of code. I can add the return if you prefer it.

js/components/controlBar.js
```diff
@@ -506,25 +498,21 @@ var ControlBar = React.createClass({
...
-            data-focus-id="quality"
-            tabIndex="0"
-            aria-label={CONSTANTS.ARIA_LABELS.VIDEO_QUALITY}
-            aria-haspopup="true"
-            aria-expanded={this.props.controller.state.videoQualityOptions.showVideoQualityPopover ? true : null}>
+            className={qualityClass}
+            focusId="quality"
+            ariaLabel={CONSTANTS.ARIA_LABELS.VIDEO_QUALITY}
+            ariaHasPopup="true"
+            ariaExpanded={this.props.controller.state.videoQualityOptions.showPopover ? true : null}
```
aeng7: How come we are using null instead of false here?

jpfranco: For some of these attribute it's better not to render them unless they are true or have a value. React doesn't render attributes that are set to `null`

js/components/controlBar.js
```diff
@@ -506,25 +498,21 @@ var ControlBar = React.createClass({
...
-            onMouseUp={Utils.blurOnMouseUp}
-            onClick={this.handleQualityClick}
-            onKeyDown={this.handleQualityKeyDown}
-            data-focus-id="quality"
-            tabIndex="0"
-            aria-label={CONSTANTS.ARIA_LABELS.VIDEO_QUALITY}
-            aria-haspopup="true"
-            aria-expanded={this.props.controller.state.videoQualityOptions.showVideoQualityPopover ? true : null}>
+            className={qualityClass}
+            focusId="quality"
```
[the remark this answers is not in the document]
jpfranco: Yes, will update them

js/components/higher-order/accessibleMenu.js
```diff
@@ -0,0 +1,116 @@
...
+        default:
+          break;
+      }
+    },
+
+    /**
+     * Finds the previous or next sibling of the given menu item and gives it focus.
+     * @private
+     * @param {Element} menuItem The menuItem element whose sibling we want to focus on.
+     * @param {Boolean} useNextSibling Choses the sibling next to menuItem when true and the previous one when false.
```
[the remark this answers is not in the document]
jpfranco: Will update

js/components/higher-order/accessibleMenu.js
```diff
@@ -0,0 +1,116 @@
...
+     */
+    focusOnMenuItemSibling: function(menuItem, useNextSibling) {
+      var menuItemsList = [];
+      if (this.menuDomElement) {
+        menuItemsList = this.menuDomElement.querySelectorAll('[' + CONSTANTS.KEYBD_FOCUS_ID_ATTR + ']');
```
aeng7: Since this query's off of the menu dom element for this specific accessible menu, is it possible to focus on another accessible menu's elements? If not, is this expected behavior?

js/components/higher-order/accessibleMenu.js
```diff
@@ -0,0 +1,116 @@
...
+      }
+      if (!menuItemsList.length) {
+        return;
+      }
+      // Since these elements aren't actually next to each other in the DOM, their position
+      // relative to one another is implied from their tab order, which should be the same as
+      // the one returned by querySelectorAll as long as tabindex is set to 0 (which should be the case).
+      var siblingIndex = this.getMenuItemSiblingIndex(menuItemsList, menuItem, useNextSibling);
+      var menuItem = menuItemsList[siblingIndex];
```
aeng7: Since it is possible to access the property called '-1' of the array, should we do an explicit check for the siblingIndex to be >= 0? The getMenuItemSiblingIndex function currently returns -1 if menuItem is absent from the list.

js/components/higher-order/accessibleMenu.js
jpfranco: So far yes, this is expected because the component assumes that the whole view will be a single menu. There is currently no use case for having multiple menus active or treating different sections of a view as separate menus.

js/components/higher-order/accessibleMenu.js
jpfranco: Will add it, but we do check `menuItem` below, so without the check it's a bit cleaner

tests/components/closed-caption/onOffSwitch-test.js
```diff
@@ -0,0 +1,61 @@
...
+      },
+      skinConfig: {
+        general: {
+          accentColor: '#fff'
+        }
+      },
+      localizableStrings: [],
+      controller: {
+        toggleClosedCaptionEnabled: function() {
+          console.log(">>>huh?");
```
aeng7: Do we need this console log?

tests/components/videoQualityPanel-test.js
```diff
@@ -1,6 +1,8 @@
 jest.dontMock('../../js/components/videoQualityPanel')
     .dontMock('../../js/components/utils')
     .dontMock('../../js/components/icon')
+    .dontMock('../../js/components/higher-order/accessibleMenu')
+    .dontMock('../../js/components//accessibleButton')
```
aeng7: extra /

js/components/higher-order/accessibleMenu.js
aeng7: Got it.

js/components/higher-order/accessibleMenu.js
aeng7: Actually we can leave it as is. Checking menuItem and menuItem.focus is enough

js/components/controlBar.js
aeng7: Sounds good, we can leave it as is.

aeng7:
LGTM outside of a few minor comments

tests/components/closed-caption/onOffSwitch-test.js
jpfranco: lol no, forgot to remove this

PR #921, opened 2017-11-22
Player 2011

aeng7:
Quality panel resolution changes


js/components/videoQualityPanel.js
```diff
@@ -84,27 +94,70 @@ var VideoQualityPanel = React.createClass({
       if (availableBitrates[i].id == 'auto') {
         this.addAutoButton(bitrateButtons);
       } else {
+        label = null;
```
jpfranco: Just a suggestion, but I think this section could probably be broken into two functions (something like getAvailableBitrate, getLabels). This might make it easier to read since the render function is a bit large now

aeng7: Sure, I'll get this in the next PR since I need to update for multiple resolutions anyways once I get feedback from Steve.

pilievOoyala:
LGTM

PR #927, opened 2017-11-27
Feature/player 1350

IvanSakharovskii:


js/components/viewControlsVr.js
```diff
@@ -18,15 +18,42 @@ var ViewControlsVr = React.createClass({
       e.cancelBubble = true; // IE
       this.props.controller.state.accessibilityControlsEnabled = true;
     }
+
     this.props.controller.moveVrToDirection(isRotated, direction);
   },
 
+  /**
+   * @method ViewControlsVr#_setIconSymbol
```
pilievOoyala: Using the word 'set' would mean you should be able to set the value to something that I pass in.  This should be named _setupIconSymbol or _initIconSymbol

js/components/viewControlsVr.js
```diff
@@ -18,15 +18,42 @@ var ViewControlsVr = React.createClass({
...
+    this.icon = _.find(desktopContent, function (el) {
+      return el.location === "mainView";
+    });
+  },
+
+  /**
+   * @method ViewControlsVr#_setSubstrateSymbol
+   * @private
+   */
+  _setSubstrateSymbol: function () {
```
pilievOoyala: Same thing here, change 'set' to something else since I can't set it to a value.  Also what does substrate mean here?  I see that we've used it a couple of times in the VR code but it's confusing what it means.  Is this the VR UI surface?

js/controller.js
```diff
@@ -32,6 +32,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
     this.videoVrSource = null;
     this.videoVr = false;
     this.captionDirection = '';
+    this.isNewVideo = true;
```
pilievOoyala: This is only ever set to true.  Do we need this? 

In general we keep track if the video is a replay or not. Not whether it's a new video. To keep it consistent with all the other codes in all the repos I suggest changing this.

js/views/playingScreen.js
```diff
@@ -177,6 +203,27 @@ var PlayingScreen = React.createClass({
     this.props.controller.handleMuteClick();
   },
 
+  /**
+   *
+   * @param vrKey - key for durection in config
```
pilievOoyala: durection -> Is this direction or duration?

pilievOoyala: Also, please rename the variable to to vrDirection or vrDuration.  Although it's used as a key, whoever is using this function shouldn't know that.

js/components/viewControlsVr.js
yvinokurov: fix

js/components/viewControlsVr.js
yvinokurov: fix

js/controller.js
pilievOoyala: change to isNewVrVideo

js/views/playingScreen.js

pilievOoyala:
LGTM

PR #928, opened 2017-11-28
[player-2011]

aeng7:
-removed “quality level” aria label for bitrate only video quality selections
-added tiered resolutions for resolution only video quality selections
-moved button creation outside of the render function into a separate function


pilievOoyala:
LGTM

PR #934, opened 2017-11-30
[player-2348] New option "showAdControlsOverAds" to display the contr…

pilievOoyala:
…ols on top of ads.  May prevent being able to click on interactive ads.  Only works if you enable the control bar is css for the ad screen.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -1138,10 +1138,6 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       this.renderSkin();
       this.createPluginElements();
 
-      if (params.showAdControlsOverAds === true) {
-        this.state.pluginsElement.css("bottom", 0);
-        this.state.pluginsElementClickElement.css("bottom", 0);
-      }
       if (typeof this.state.config.closedCaptionOptions === 'object' &&
         this.state.config.closedCaptionOptions.language !== undefined) {
         this.setCaptionDirection(this.state.config.closedCaptionOptions.language);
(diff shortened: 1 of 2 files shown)


pilievOoyala:
Please don't merge just yet. I just want your initial opinion of this.

aeng7:
Does the control bar remain static on screen still? We may need more changes to mimic the regular content user experience.

js/controller.js
```diff
@@ -1137,6 +1137,11 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       this.state.configLoaded = true;
       this.renderSkin();
       this.createPluginElements();
+
+      if (params.showAdControlsOverAds === true) {
```
jpfranco: Looks ok but wouldn't it be better to use the same mechanism that is used for VAST in order to show the controls?

pilievOoyala:
@aeng7 THe control bar works just like it does with the main video.
@jpfranco  I'll look into that but from what I saw, the control isn't showing their either.

aeng7:
Can we add some unit tests? LGTM otherwise.

js/controller.js
```diff
@@ -1145,10 +1146,26 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+      }
+
       var fullClass = (this.state.config.adScreen.showControlBar ? "" : " oo-full");
       $("#" + this.state.elementId + " .oo-player-skin").append("<div class='oo-player-skin-plugins"+fullClass+"'></div><div class='oo-player-skin-plugins-click-layer"+fullClass+"'></div>");
       this.state.pluginsElement = $("#" + this.state.elementId + " .oo-player-skin-plugins");
       this.state.pluginsClickElement = $("#" + this.state.elementId + " .oo-player-skin-plugins-click-layer");
+
+      //if playerControlsOverAds is true, then we need to set the size of the
+      //elements to be the full size of the player and not end where the control bar begins.
+      if (this.state.playerParam && this.state.playerParam.playerControlsOverAds) {
```
jpfranco: Does this need to be reset at some point in case there's a new video or settings that don't require this?

pilievOoyala: I was considering this a player setting and not a video setting. So it wouldn't change between videos

pilievOoyala: Do we support changing the skin between videos?

jpfranco: Oh ok, I don't think we support that, so no need to reset then.

tests/html5skin-test.js
```diff
@@ -448,4 +448,54 @@ describe('Controller', function() {
 
   });
 
+  describe("Show player controls over ads", function() {
```
aeng7: Can we add tests to check the css of the plugins element and plugins click element depending on the value of playerControlsOverAds?

tests/html5skin-test.js
```diff
@@ -15,6 +15,20 @@ var CONSTANTS = require('../js/constants/constants');
 
 var Html5Skin;
 
+var elementId = 'adrfgyi';
+var videoId = 'ag5dfdtooon2cncj714i';
```
aeng7: toooon

aeng7:
LGTM, although in the future we should see if we can run the unit tests without creating a full fledged Ooyala video element

PR #940, opened 2017-12-01
[bugfix/PBW-6987] Player 4.19.3 - watermark logo not clickable anymore

IvanSakharovskii:


scss/base/_variables.scss
```diff
@@ -133,6 +133,7 @@ $SUPER_Z_INDEX:                           20000 !default; //in core, here for bi
 $zindex-fullscreen:                  2147483647 !default; //in screenfull, here for bird's eye view tracking
 $zindex-state-screen:                      1000 !default;
 $zindex-vr-controls:                       1001 !default;
+$zindex-watermark:                        12601 !default;
```
jpfranco: How was this value chosen? Would 1001, be enough since the watermark is currently being blocked by `.oo-state-screen-selectable` (which has a z-index of 1000)?

scss/components/_watermark.scss
```diff
@@ -4,6 +4,7 @@
   max-width: 100%;
   pointer-events: auto;
   outline: none !important;
+  z-index: $zindex-watermark;
```
jpfranco: Looks good overall, but please make sure to check the following:

- Are there any other elements affected by the higher z-index of `.oo-state-screen-selectable` that we haven't found yet? This was also affecting the up next panel, so there could be others remaining.
- Does the new z-index for watermark place it on top of other elements that it shouldn't be on top of? Not sure if there's a case where this is true, but we should verify.

scss/base/_variables.scss
KseniiaSokolovaEpam: \+

pilievOoyala:
LGTM

jpfranco:
Reverting for now as discussed with @IvanSakharovskii

PR #941, opened 2017-12-01
Player 590

dallen42:


js/components/countDownClock.js
```diff
@@ -156,7 +156,15 @@ var CountDownClock = React.createClass({
   startDiscoveryVideo: function() {
     var eventData = {
           "clickedVideo" : this.props.discoveryData.relatedVideos[0],
-          "custom" : this.props.discoveryData.custom
+          "custom" : {
+            "source": CONSTANTS.SCREEN.UP_NEXT_SCREEN,
+            "asset" : { "id" : this.props.discoveryData.relatedVideos[0].embed_code, "idType" : CONSTANTS.DISCOVERY.ID_TYPE},
+            "autoplay": true,
+            "assetPosition" : 1, 
+            "pageSize" : 1,
```
aeng7: What are pageSize and assetPosition? Should these always be 1?

js/components/countDownClock.js
```diff
@@ -165,7 +173,15 @@ var CountDownClock = React.createClass({
     OO.log("startUpNext");
     var eventData = {
       "clickedVideo" : this.props.upNextInfo.upNextData,
-      "custom" : {"source": CONSTANTS.SCREEN.UP_NEXT_SCREEN}
+      "custom": {
```
aeng7: If this is the same as the one above in startDiscoveryVideo, can we move this into a separate function to avoid duplicate code?

aeng7:
Can we add some unit tests?

js/controller.js
```diff
@@ -1575,28 +1578,55 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+                        "uiTag" : uiTag,
+                        "assetInfo": assetData
+                     }
+        };
+        this.mb.publish(OO.EVENTS.DISCOVERY_API.SEND_DISPLAY_EVENT, eventData);
+      }
+    },
+
+    getDiscoveryContext: function(discoveryAsset) {
+      if (discoveryAsset == null) return {};
```
pilievOoyala: can you put this on two separate lines?

js/components/discoveryPanel.js
```diff
@@ -221,7 +241,10 @@ DiscoveryPanel.defaultProps = {
     relatedVideos: []
   },
   controller: {
-    sendDiscoveryClickEvent: function(a,b){}
+    sendDiscoveryClickEvent: function(a,b){},
+    sendDiscoveryDisplayEvent: function(a,b,c,d,e){}
```
pilievOoyala: Maybe I'm getting confused but here this looks like a temp/stub function. In the unit test changes below(tests/controller-test.js) you are sending actual values for each param.  Where are those defined?

js/components/countDownClock.js
dallen42: @aeng7 refactored this into a few functions now in Utils

js/controller.js
dallen42: @pilievOoyala moved to Utils and fixed

dallen42:
@aeng7 added some unit tests for the new Utils, and move decode64 to skin to make testing easier... let me know if you think we need any more coverage here

js/components/discoveryPanel.js
dallen42: @pilievOoyala I think this stub is needed for unit tests in discoveryPanel-test.js. Because of the changes for analytics.js sendDiscoveryDisplayEvent is now fired from within the UI element and not the controller. In fact, that unit test in controller-test.js was wrong, it should be testing the old function sendDiscoveryDisplayEvent function that was fired in controller.js (renamed sendDiscoveryDisplayEventOld) that we need to keep around in order to still have a path for sending discovery events to reporter.js (until it is removed completely). I updated the controller-test.js unit test fix that now

pilievOoyala:
@dallen42 Is this still valid or can we close it?

dallen42:
Closing this as it is old and requires a lot of code merge now

PR #949, opened 2017-12-08
Feature/player 1350

IvanSakharovskii:


js/skin.js
```diff
@@ -73,17 +73,40 @@ var Skin = React.createClass({
     });
   },
 
+  /**
+   * @public
+   * @description - returns the correct coordinates of events depending on the platform
+   * @param e - event
+   * @returns {object} - coordinates x, y
+   */
+  getCoords: function(e) {
```
jpfranco: I think this should be moved to `Utils.js`. It's not directly related to `skin.js` logic

js/controller.js
```diff
@@ -1293,10 +1293,16 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       }
       // no support
       else {
-        if(this.state.isFullWindow) {
-          this.exitFullWindow();
+
+        if(OO.isIos && this.videoVr){
+          // toggle fullscreen with bitmovin api
+          this.mb.publish(OO.EVENTS.TOGGLE_FULLSCREEN_VR, this.focusedElement);
```
jpfranco: Can we add more details about why the usual fullscreen mechanism doesn't work for VR?

js/controller.js
```diff
@@ -1293,10 +1293,16 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       }
       // no support
       else {
-        if(this.state.isFullWindow) {
-          this.exitFullWindow();
+
+        if(OO.isIos && this.videoVr){
```
jpfranco: Do we need a unit test for this condition?

js/views/pauseScreen.js
```diff
@@ -53,19 +53,22 @@ var PauseScreen = React.createClass({
...
 
   handlePlayerMouseDown: function(e) {
+    e.persist();
     this.props.controller.state.accessibilityControlsEnabled = true;
     this.props.controller.state.isClickedOutside = false;
     this.props.handleVrPlayerMouseDown(e);
   },
   handlePlayerMouseMove: function(e) {
     e.preventDefault();
+    e.persist();
```
jpfranco: Please make sure that these `preventDefault`'s, etc. are not applied to all videos if they are only needed for VR360. Not sure if this is the case here, but we should check.

pilievOoyala:
Look fine aside from JP's comments.

js/skin.js
yvinokurov: fix

js/controller.js
yvinokurov: Ok. Fix

js/controller.js
yvinokurov: this.state.isVideoFullScreenSupported == false;
this.state.isFullScreenSupported == false; 
The rest also did not work. fullscreen bitmovin works correctly

js/components/controlBar.js
```diff
@@ -101,13 +101,70 @@ var ControlBar = React.createClass({
...
+        this.unlockScreenOrientation();
+      }
+    }
+  },
+
+  /**
+   * @description set landscape orientation if it is possible
+   * @private
+   */
+  setLandscapeScreenOrientation: function() {
```
pilievOoyala: if this is private can we put a "_" in the front?

js/components/controlBar.js
```diff
@@ -101,13 +101,70 @@ var ControlBar = React.createClass({
...
+      }
     }
-    if(this.props.controller && typeof this.props.controller.toggleStereoVr === "function") {
+  },
+
+  /**
+   * @description set possibility to use all orientations
+   * @private
+   */
+  unlockScreenOrientation: function() {
```
pilievOoyala: here too

js/controller.js
pilievOoyala: Can you add this in as a comment?

PR #950, opened 2017-12-08
[PLAYER-1678] Added remaining unit tests

jpfranco:
Apart from the unit tests, this PR also includes:

- Fixed a few React warnings
- Switched the remaining control bar buttons to `AccessibleButton`'s.

### DIFF
--- a/tests/components/colorSelector-test.js
+++ b/tests/components/colorSelector-test.js
@@ -49,7 +49,7 @@ describe('Tabs', function() {
   it('should render a ColorSelector', function() {
     renderComponent();
     expect(component).toBeTruthy();
-    expect(buttons.length).toBe(6);
+    expect(buttons.length).toBe(colors.length);
   });
 
   it('should set menu role and ARIA label on main element', function() {
--- a/tests/components/colorSelector-test.js
+++ b/tests/components/colorSelector-test.js
@@ -1,0 +1,87 @@
+jest
+.dontMock('../../js/components/colorSelector')
+.dontMock('../../js/components/accessibleButton')
+.dontMock('../../js/components/higher-order/accessibleMenu')
+.dontMock('../../js/components/utils')
+.dontMock('../../js/constants/constants')
+.dontMock('classnames');
+
+var React = require('react');
+var ReactDOM = require('react-dom');
+var TestUtils = require('react-addons-test-utils');
+var ColorSelector = require('../../js/components/colorSelector');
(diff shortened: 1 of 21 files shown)


pilievOoyala:
LGTM

tests/components/colorSelector-test.js
```diff
@@ -0,0 +1,87 @@
...
+          accentColor: 'blue'
+        }
+      }
+    };
+  });
+
+  it('should render a ColorSelector', function() {
+    renderComponent();
+    expect(component).toBeTruthy();
+    expect(buttons.length).toBe(6);
```
aeng7: Should we compare this against colors.length rather than 6?

tests/components/slider-test.js
```diff
@@ -0,0 +1,154 @@
...
+    expect(element.getAttribute('aria-label')).toBe(props.ariaLabel);
+    expect(element.getAttribute('aria-valuemin')).toBe('0');
+    expect(element.getAttribute('aria-valuemax')).toBe('100');
+    expect(element.getAttribute('aria-valuenow')).toBe('50');
+    expect(element.getAttribute('aria-valuetext')).toBe('50% ' + props.settingName);
+  });
+
+  it('should set a default focus id if none is provided', function() {
+    renderComponent();
+    expect(element.getAttribute(CONSTANTS.KEYBD_FOCUS_ID_ATTR).length).toBe(10);
```
aeng7: Where does the number 10 come from?

tests/components/slider-test.js
```diff
@@ -0,0 +1,154 @@
...
+  });
+
+  it('should update ARIA values when value changes', function() {
```
aeng7: Should we add some tests for negative numbers and numbers greater than 1?

tests/components/colorSelector-test.js
jpfranco: Updating

tests/components/slider-test.js
jpfranco: This is just the length of the generated strings. Since they're random I can't check the actual value, but this is a bit better than checking that it's truthy.

aeng7:
LGTM outside a few minor comments

tests/components/slider-test.js
aeng7: Sounds good

aeng7:
Feel free to merge after updating

PR #957, opened 2017-12-14
Feature/player 1350

IvanSakharovskii:


scss/base/_variables.scss
```diff
@@ -132,10 +132,6 @@ $zindex-ad-top-bar:                       12600 !default;
 $zindex-popover:                          12600 !default;
 $SUPER_Z_INDEX:                           20000 !default; //in core, here for bird's eye view tracking
 $zindex-fullscreen:                  2147483647 !default; //in screenfull, here for bird's eye view tracking
-$zindex-state-screen:                      1000 !default;
-$zindex-vr-controls:                       1001 !default;
-$zindex-watermark:                         1001 !default;
-$zindex-up-next-panel:                     1002 !default;
```
jpfranco: Just to confirm, these are the values that had been added for VR360 changes so this would take us back to the original state, correct?

jpfranco:
Changes look good, would just recommend adding unit tests.

scss/base/_variables.scss
pilievOoyala: The git blame show Kseniia is the last to touch it.  
be8c8c626 (Kseniia Sokolova  2017-10-13 16:27:39 +0300 135) $zindex-state-screen:                      1000 !default;
be8c8c626 (Kseniia Sokolova  2017-10-13 16:27:39 +0300 136) $zindex-vr-controls:                       1001 !default;
ff4b81c55 (Kseniia Sokolova  2017-12-04 17:39:47 +0300 137) $zindex-watermark:                         1001 !default;
7c9ace1b5 (Kseniia Sokolova  2017-10-16 15:44:23 +0300 138) $zindex-up-next-panel:                     1002 !default;

pilievOoyala: And before that they were renamed to their current form. So I'm confident they were added for vr.

pilievOoyala:
Going to merge but @IvanSakharovskii please make sure to add unit tests in a separate check in.  Thanks!

js/controller.js
```diff
@@ -295,8 +295,8 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       if (!this.state.mainVideoInnerWrapper.children('.oo-player-skin').length) {
         this.state.mainVideoInnerWrapper.append("<div class='oo-player-skin'></div>")
       }
-
-      this.state.mainVideoInnerWrapper.attr('style', '');
+      //https://stackoverflow.com/questions/42206645/konvajs-unable-to-preventdefault-inside-passive-event-listener-due-to-target-be
+      this.state.mainVideoInnerWrapper.attr('style', 'touch-action: none');
```
jpfranco: Sorry, I didn't notice this line originally. Even though this is not inherited, blocking touches on the main wrapper might have unexpected consequences. We should at least limit this to VR360, but we should look for a different approach if possible.

jpfranco: I'm also noticing a bug after pulling these changes. Previously, tapping on the video would should the controls and then you would tap again if you wanted to pause. After these changes, tapping once pauses the video immediately. Could this line be related?

jpfranco:
Reverting for now, as discussed with @IvanSakharovskii

PR #965, opened 2017-12-18
Feature/player 1350 without revert

yvinokurov:


js/controller.js
```diff
@@ -312,8 +312,21 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
 
+    /**
+     * Set style "touch-action: none" only for video 360 on mobile devices
+     * see details: https://stackoverflow.com/questions/42206645/konvajs-unable-to-preventdefault-inside-passive-event-listener-due-to-target-be
+     */
+    setInlineStyles: function () {
+      if(this.videoVr && this.state.isMobile) {
+        this.state.mainVideoInnerWrapper.attr('style', 'touch-action: none');
+      } else {
+        this.state.mainVideoInnerWrapper.attr('style', '');
```
IvanSakharovskii: can we do something like
`if (this.videoVr) {
     if (this.state.isMobile) {
        this.state.mainVideoInnerWrapper.attr('style', 'touch-action: none');
     } else {
         this.state.mainVideoInnerWrapper.attr('style', '');
     }
  }
`

yvinokurov: fix

PR #966, opened 2017-12-18
Feature/player 1350

IvanSakharovskii:


pilievOoyala:
LGTM.  @jpfranco Can you review this too?

jpfranco:
@IvanSakharovskii we normally require 2 approvals before merging to master. Please do not merge until review is complete.

js/controller.js
```diff
@@ -296,7 +297,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         this.state.mainVideoInnerWrapper.append("<div class='oo-player-skin'></div>")
       }
 
-      this.state.mainVideoInnerWrapper.attr('style', '');
```
jpfranco: We should keep this line where `style` is cleared. This could've been added for a reason.

js/controller.js
```diff
@@ -312,16 +313,41 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+
     onSetVideoVr: function(event, params) {
       this.videoVr = true;
+      this.setInlineStyles();
       if (params) {
         this.videoVrSource = params.source || null; //if we need video vr params
       }
     },
 
+    handleVrMobileOrientation: function(e) {
```
jpfranco: Can we add some comments describing what this function does?

js/components/accessibilityControls.js
```diff
@@ -27,6 +27,7 @@ AccessibilityControls.prototype = {
   cleanUp : function() {
     document.removeEventListener("keydown", this.keyEventDown);
     document.removeEventListener("keyup", this.keyEventUp);
+    document.removeEventListener("deviceorientation", this.controller.handleVrMobileOrientation);
```
jpfranco: We shouldn't be removing an event listener that was set up in a different file. We should either add the event listener here or move the removal to the original file.

js/controller.js
```diff
@@ -353,6 +379,11 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         this.enableFullScreen();
         this.updateAspectRatio();
       }
+      if (this.videoVr) {
+        if (window.DeviceOrientationEvent) {
+          window.addEventListener('deviceorientation', this.handleVrMobileOrientation.bind(this), false);
```
jpfranco: Note that `bind` is returning a new function here, so calling `removeEventListener` with `this.handleVrMobileOrientation` will not remove this listener.

jpfranco:
@IvanSakharovskii please include a description and unit tests for all of your pull requests.

js/controller.js
yvinokurov: This line was added by me in one of the past PRs. In mainVideoInnerWrapper, style = "width: inherit; height: inherit; position: relative; padding-top: 50%;". Inline-style is difficult to override so I changed style = ''. Now the style changes are only for video360. For the flat video, everything remains unchanged

js/components/accessibilityControls.js
KseniiaSokolovaEpam: \+

js/controller.js

js/controller.js

PR #990, opened 2018-01-19
[player-2427]

aeng7:
-fixed an issue where if content aspect ratio is published during an ad, the ad aspect ratio will change to that of the reported content aspect ratio


pilievOoyala:
LGTM

PR #992, opened 2018-01-23
don't hide horizontal scrollbar

dallen42:


dallen42:
For PBW-7063 - please review but do not merge yet (may need UI/UX opinion on this first)

scss/components/_tabs.scss
```diff
@@ -52,11 +52,6 @@
     display: inline-block;
   }
 
-  //hide horizontal scrollbars
-  .scrollbar-container.horizontal {
```
jpfranco: I think this is for the tabs on the CC menu, do we want to show the scrollbar here as well? The tabs have left and right arrows currently.

dallen42: @jpfranco yeah I thought that was affecting the quality page too but it wasn't actually, reverted it

jpfranco:
Forgot that this wasn't ready to merge. Please wait for confirmation before merging.

PR #1000, opened 2018-01-30
Feature/player 2630

yvinokurov:
creating additional tests, refactoring existing tests


tests/components/controlBar-test.js
```diff
@@ -167,7 +167,7 @@ describe('ControlBar', function () {
   it('not render stereo button if content not vr', function () {
     var mockController = {
       state: {
-        isMobile: false,
+        isMobile: true,
```
IvanSakharovskii: Can we add separate tests for desktop and mobile?

yvinokurov: This is a test fix, the test was incorrect. Under any circumstances, there is no stereo button on the desktop.

PR #1001, opened 2018-01-30
[bugfix/PLAYER-2783] Clicking on any icon in the control bar pauses t…

KseniiaSokolovaEpam:
…he video/ anywhere outside the player pauses the video

### DIFF
--- a/js/skin.js
+++ b/js/skin.js
@@ -207,6 +207,17 @@ var Skin = React.createClass({
     return vrViewingDirectionValue
   },
 
+
+  /**
+   *
+   * @param comparableClass {string} - className for comparing
+   * @param originClass {string} - className for comparing
+   * @returns {boolean} returns true if classes are identical
+   */
+  checkClassName: function(comparableClass, originClass) {
+    return !!(comparableClass === originClass);
+  },
+
   render: function() {
     var screen;
 
@@ -267,6 +278,7 @@ var Skin = React.createClass({
               handleVrPlayerMouseUp={this.handleVrPlayerMouseUp}
               handleVrPlayerClick={this.handleVrPlayerClick}
               handleVrPlayerFocus={this.handleVrPlayerFocus}
+              checkClassName={this.checkClassName}
               isVrMouseMove={this.state.isVrMouseMove}
               contentTree={this.state.contentTree}
               currentPlayhead={this.state.currentPlayhead}
@@ -309,6 +321,7 @@ var Skin = React.createClass({
               handleVrPlayerMouseMove={this.handleVrPlayerMouseMove}
               handleVrPlayerMouseUp={this.handleVrPlayerMouseUp}
               handleVrPlayerClick={this.handleVrPlayerClick}
+              checkClassName={this.checkClassName}
               isVrMouseMove={this.state.isVrMouseMove}
               contentTree={this.state.contentTree}
               currentPlayhead={this.state.currentPlayhead}
--- a/js/skin.js
+++ b/js/skin.js
@@ -207,17 +207,6 @@ var Skin = React.createClass({
     return vrViewingDirectionValue
   },
 
-
-  /**
-   *
-   * @param comparableClass {string} - className for comparing
-   * @param originClass {string} - className for comparing
-   * @returns {boolean} returns true if classes are identical
-   */
-  areClassNamesEquals: function(comparableClass, originClass) {
-    return !!(comparableClass === originClass);
-  },
-
   render: function() {
     var screen;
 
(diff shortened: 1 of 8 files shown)


js/skin.js
```diff
@@ -207,6 +207,17 @@ var Skin = React.createClass({
...
   },
 
+
+  /**
+   *
+   * @param comparableClass {string} - className for comparing
+   * @param originClass {string} - className for comparing
+   * @returns {boolean} returns true if classes are identical
+   */
+  areClassNamesEquals: function(comparableClass, originClass) {
```
jpfranco: I think this is too generic to be in `skin.js`, this would probably be better inside `Utils`

js/skin.js
```diff
@@ -267,6 +278,7 @@ var Skin = React.createClass({
               handleVrPlayerMouseUp={this.handleVrPlayerMouseUp}
               handleVrPlayerClick={this.handleVrPlayerClick}
               handleVrPlayerFocus={this.handleVrPlayerFocus}
+              areClassNamesEquals={this.areClassNamesEquals}
```
jpfranco: Why do we need to pass this function to components? It would be cleaner to have components import these from Utils.

js/views/pauseScreen.js
```diff
@@ -64,7 +65,8 @@ var PauseScreen = React.createClass({
   },
 
   handleTouchEnd: function(e) {
-    if (e.target.className === "oo-state-screen-selectable") {
+    var isSelectableScreen = this.props.areClassNamesEquals(e.target.className, this.selectableScreenClassName);
```
jpfranco: Is it necessary to listen to these events at the document level even if we only execute the handler for a specific element? Wouldn't it be better to attach the event handler to that element only?

jpfranco:
Is this update addressing the issue where the accessibility controls are triggered when using the keyboard outside the player? I'm still seeing this issue on this branch when the video is playing.

js/views/playingScreen.js
```diff
@@ -24,6 +24,7 @@ var PlayingScreen = React.createClass({
   getInitialState: function() {
     this.isMobile = this.props.controller.state.isMobile;
     this.browserSupportsTouch = this.props.controller.state.browserSupportsTouch;
+    this.selectableScreenClassName = 'oo-state-screen-selectable';
```
jpfranco: As I mentioned before, I think it would be preferable to set the event handlers on the element that requires them instead of the document, but if the events are needed at the document level then we should probably store this class in a constant since it's used in several places.

KseniiaSokolovaEpam: For vr 360 we need to listen this event on document.

js/skin.js

js/skin.js

js/views/pauseScreen.js
KseniiaSokolovaEpam: For vr 360 we need to listen this event on document. And if add listener in two places it will be dublicate of the code.

jpfranco: I see. I would personally prefer to have two separate handlers in order to avoid checking for a class (which could potentially change), but leave it up to the second reviewer @aeng7 @pilievOoyala

js/views/playingScreen.js
```diff
@@ -187,9 +188,11 @@ var PlayingScreen = React.createClass({
...
       e.stopPropagation(); // W3C
       e.cancelBubble = true; // IE
-      if (!this.props.controller.videoVr && !this.props.controller.state.accessibilityControlsEnabled) {
-        this.props.controller.togglePlayPause();
+      var isSelectableScreen = Utils.areArgumentsEqual(e.target.className, CONSTANTS.CLASS_NAMES.SELECTABLE_SCREEN);
+      if (!this.props.controller.videoVr && isSelectableScreen) {
+        this.props.controller.togglePlayPause();//if clicked on selectableSceen
       }
+      //the order of the loop and this.props.controller.state is not important
       this.props.controller.state.accessibilityControlsEnabled = true;
```
pilievOoyala: Can we take this line out? Because it's turning accessbility on regardless of it was enabled before this.

pilievOoyala:
LGTM. @aeng7 @dallen42 Can you also review?

pilievOoyala:
LGTM now.

js/views/pauseScreen.js
KseniiaSokolovaEpam: yes, i separated it already in todays pr.

KseniiaSokolovaEpam:
@jpfranco 
>> Is this update addressing the issue where the accessibility controls are triggered when using the keyboard outside the player? I'm still seeing this issue on this branch when the video is in playing mode (although it seems to be fixed when paused).

Where did you check ? I checked with the last build (you need to use sandbox/PLAYER-2783 for skin) https://goo.gl/n13e3A and everything is ok here.

jpfranco:
@KseniiaSokolovaEpam I tested this locally with the `bugfix/PLAYER-2783` branch, before the latest updates. The issue was still manifesting at that point.

PR #1002, opened 2018-01-30
Added missing unit tests 

IvanSakharovskii:


tests/components/viewControlsVr-test.js
```diff
@@ -124,4 +122,99 @@ describe('viewControlsVr', function () {
...
+    };
+    var DOM = TestUtils.renderIntoDocument( <ViewControlsVr {...mockProps}/> );
+    var iconSubstrate = TestUtils.scryRenderedDOMComponentsWithClass(DOM, 'oo-vr-icon--substrate');
+    var iconSymbol = TestUtils.scryRenderedDOMComponentsWithClass(DOM, 'oo-vr-icon--icon-symbol');
+
+    expect(iconSubstrate.length).toBe(1);
+    expect(iconSymbol.length).toBe(1);
+  });
+
+  it('should be render viewControlsVr only desktop', function () {
```
pilievOoyala: In another checkin can you check that is does render on desktop.  Not just check that it doesn't render on mobile.

pilievOoyala: So please check what happens when isMobile is false.

yvinokurov: This check exists. Look at the tests "check condition: if video support vr360 then viewControlsVr exist" "on the ViewControlsVr should be two icons: one icon of the background and one icon of the symbol" It checks not only the presence of the control, but also checks for the presence of internal elements.

PR #1003, opened 2018-02-01
[player-2812]

aeng7:
-we will now show the ad control bar if the ad gets paused so that the user has an indication to unpause the ad

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -753,6 +753,10 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         this.renderSkin();
       }
       else if (videoId == OO.VIDEO.ADS){
+        //If we pause during an ad (such as for clickthroughs or when autoplay fails)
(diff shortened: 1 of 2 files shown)


js/controller.js
```diff
@@ -753,6 +753,10 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         this.renderSkin();
       }
       else if (videoId == OO.VIDEO.ADS){
+        //If we pause during an ad (such as for clickthroughs or when autoplay fails)
+        //we'll show the control bar so that the user has an indication that the video
+        //must be unpaused to resume
+        this.state.config.adScreen.showControlBar = true;
```
pilievOoyala: Do we need to worry about the autohiding, or will it show again if click the screen after it hides?

aeng7: It autohides by itself. It will appear from then on when the user hovers over the video.

pilievOoyala:
LGTM

PR #1004, opened 2018-02-09
[Bugfix/player 2900] If geo-blocking on ios, then we need to exit full-screen mode.

yvinokurov:
fixed bug https://jira.corp.ooyala.com/browse/PLAYER-2900
If geo-blocking on ios, then you need to exit full-screen mode.
test added


pilievOoyala:
This is not the right place for this code.  This code needs to live inside the VTC.  The skin is optional, so if a customer uses a completely custom skin, then they will not have this fix.

jpfranco:
We implemented a workaround for this issue on the VTC. You can take a look at the changes here: https://git.corp.ooyala.com/projects/PBW/repos/mjolnir/pull-requests/420/overview

yvinokurov:
Thank you Peter, ! I did not take it into account. I will be more attentive.

PR #1006, opened 2018-02-15
 [bugfix/PLAYER-2220]: videoVr should not blur

IvanSakharovskii:
Video 360 should not be blured on pause (videoVr should not blur). This is because even for paused video user is able to look around and this is useless for blured video.
If we have Video360 then we don't set class to blur the player.
Unit tests added.


pilievOoyala:
LGTM

PR #1008, opened 2018-02-17
[player-2502]

aeng7:
-pause screen will no longer blur video and show an opaque overlay when there is no title and no description

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -2064,7 +2064,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
      */
     addBlur: function() {
       // [PLAYER-2220]: videoVr should not blur. This prevents a circular review on a pause.
-      if (!this.videoVr) {
+      if (!this.videoVr && this.state.mainVideoElement && this.state.mainVideoElement.classList) {
         this.state.mainVideoElement.classList.add('oo-blur');
       }
     },
--- a/js/controller.js
+++ b/js/controller.js
@@ -738,6 +738,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
           OO.log("Should display DISCOVERY_SCREEN on pause");
           this.sendDiscoveryDisplayEvent("pauseScreen");
           this.state.screenToShow = CONSTANTS.SCREEN.DISCOVERY_SCREEN;
+          this.addBlur();
         } else if (this.skin.props.skinConfig.pauseScreen.screenToShowOnPause === "social") {
           // Remove this comment once pause screen implemented
         } else {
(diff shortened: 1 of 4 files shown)


js/controller.js
```diff
@@ -749,11 +750,6 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
           this.state.screenToShow = CONSTANTS.SCREEN.PAUSE_SCREEN;
         }
         this.state.playerState = CONSTANTS.STATE.PAUSE;
-
-        // [PLAYER-2220]: videoVr should not blur. This prevents a circular review on a pause.
-        if (!this.videoVr) {
-          this.state.mainVideoElement.classList.add('oo-blur');
```
jpfranco: It looks like this used to be executed even if the condition on the if statement where `addBlur()` is called was false. Will this change any behavior?

js/controller.js
```diff
@@ -2061,6 +2057,18 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
 
+    /**
+     * Adds a blur effect to the video
+     * @protected
+     * @method Html5Skin#addBlur
+     */
+    addBlur: function() {
+      // [PLAYER-2220]: videoVr should not blur. This prevents a circular review on a pause.
+      if (!this.videoVr) {
+        this.state.mainVideoElement.classList.add('oo-blur');
```
jpfranco: Do we need null checks for `mainVideoElement` since this can be called at a different point in the code?

js/views/pauseScreen.js
```diff
@@ -216,11 +219,19 @@ var PauseScreen = React.createClass({
         controlBarVisible={this.state.controlBarVisible}
       /> : null;
 
+    var containsTitle = this.props.skinConfig.pauseScreen.showTitle && !!title;
+    var containsDescription = this.props.skinConfig.pauseScreen.showDescription && !!description;
+    var containsText = containsTitle || containsDescription;
+
+    if (containsText) {
+      this.props.controller.addBlur();
```
jpfranco: Does this need to be executed on each render or would it be better to execute it once on `componentDidMount`?

js/controller.js
aeng7: Added

js/views/pauseScreen.js
aeng7: Added a new state ``containsText`` which will control the blur and the fade underlay. Moved the ``addBlur`` call into ``componentDidMount``.

js/controller.js
jpfranco: Ok, sounds good

pilievOoyala:
LGTM

PR #1009, opened 2018-02-19
Feature/player-2956 Create functions for sending the info from plugins to core

KseniiaSokolovaEpam:
Listeners for MULTI_AUDIO_FETCHED event are added, function onMultiAudioFetched is empty now. No unit tests while the function is empty.


pilievOoyala:
LGTM. But do we need to add an unsubscribe to one of these two functions too:
    unsubscribeFromMessageBus or  unsubscribeBasicPlaybackEvents

aeng7:
LGTM

PR #1010, opened 2018-02-19
Feature/player 2956

KseniiaSokolovaEpam:
Listeners for MULTI_AUDIO_FETCHED event are added, function onMultiAudioFetched is empty now. No unit tests while the function is empty.


js/controller.js
```diff
@@ -855,6 +856,16 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
     },
 
+    /**
+     *
+     * @param event {String} name of a event
+     * @param multiAudio {Object} - audio which fetched for the current video
+     * @param multiAudio.multiAudio {Array} - list of objects with data for each audio
+     */
+    onMultiAudioFetched: function(event, multiAudio) {
+    //  TODO: code for a video with multiaudio should be here
```
yvinokurov: What is it?

PR #1012, opened 2018-02-22
[PLAYER-2947] dev-epam to master. 

KseniiaSokolovaEpam:
Add UI component if multiAudio is available. 
Unit tests for functions added, 
for UI will be added after UI implemented. 
Unit tests were run without errors.

## **Important!**
_When you merge config to Master (in this PR https://github.com/ooyala/skin-config/pull/190 ) 
then you will need to set HEAD of config in html5-skin to Master._

### DIFF
--- a/js/components/controlBar.js
+++ b/js/components/controlBar.js
@@ -13,6 +13,7 @@ var React = require('react'),
   VolumeControls = require('./volumeControls'),
   VideoQualityPanel = require('./videoQualityPanel'),
   ClosedCaptionPopover = require('./closed-caption/closedCaptionPopover'),
+  MultiAudioPopover = require('./multi-audio/multiAudioPopover'),
   Logo = require('./logo'),
   Icon = require('./icon'),
   Tooltip = require('./tooltip');
@@ -230,6 +231,13 @@ var ControlBar = React.createClass({
       this.togglePopover(CONSTANTS.MENU_OPTIONS.CLOSED_CAPTIONS);
       this.closeQualityPopover();
     }
+  },
+
+  handleMultiAudioClick: function () {
+    this.configureMenuAutofocus(CONSTANTS.MENU_OPTIONS.MULTI_AUDIO);
+
+    this.togglePopover(CONSTANTS.MENU_OPTIONS.MULTI_AUDIO);
+    this.closePopovers();
   },
 
   configureMenuAutofocus: function(menu) {
@@ -478,6 +486,12 @@ var ControlBar = React.createClass({
       "oo-selected": this.props.controller.state.closedCaptionOptions.showPopover
     });
 
+    var multiAudioClass = ClassNames({
+      "oo-multiaudio": true,
+      "oo-control-bar-item": true,
+      "oo-selected": this.props.controller.state.multiAudioOptions.showPopover
+    });
+
     var selectedStyle = {};
     selectedStyle["color"] = this.props.skinConfig.general.accentColor ? this.props.skinConfig.general.accentColor : null;
 
@@ -623,6 +637,36 @@ var ControlBar = React.createClass({
         )
       }).bind(this),
 
+      "multiAudio": (function (alignment) {
+        return (
+          <div className="oo-multiaudio-container" key="multiAudio">
+            <AccessibleButton
+              ref={function(e) { this.toggleButtons[CONSTANTS.MENU_OPTIONS.MULTI_AUDIO] = e }.bind(this)}
+              style={selectedStyle}
+              className={multiAudioClass}
+              focusId={CONSTANTS.FOCUS_IDS.MULTI_AUDIO}
+              ariaLabel={CONSTANTS.ARIA_LABELS.MULTI_AUDIO}
+              ariaHasPopup={true}
+              ariaExpanded={this.props.controller.state.multiAudioOptions.showPopover ? true : null}
+              onClick={this.handleMultiAudioClick}
+            >
+              <Icon {...this.props} icon="multiAudio" style={dynamicStyles.iconCharacter}
+                    onMouseOver={this.highlight} onMouseOut={this.removeHighlight} />
+              <Tooltip enabled={isTooltipEnabled} text={Utils.getLocalizedString(this.props.language, CONSTANTS.SKIN_TEXT.MULTI_AUDIO, this.props.localizableStrings)} responsivenessMultiplier={this.responsiveUIMultiple} bottom={this.responsiveUIMultiple * this.props.skinConfig.controlBar.height} alignment={alignment} />
+            </AccessibleButton>
+            {this.props.controller.state.multiAudioOptions.showPopover &&
+            <Popover
+              popoverClassName="oo-popover oo-popover-pull-right"
+              autoFocus={this.props.controller.state.multiAudioOptions.autoFocus}
+              closeActionEnabled={this.props.controller.state.accessibilityControlsEnabled}
+              closeAction={this.closePopover.bind(this, CONSTANTS.MENU_OPTIONS.MULTI_AUDIO)}>
+              <MultiAudioPopover {...this.props} togglePopoverAction={this.closePopover.bind(this, CONSTANTS.MENU_OPTIONS.MULTI_AUDIO)} />
+            </Popover>
+            }
+          </div>
+        )
+      }).bind(this),
+
       "share": (function (alignment) {
         return <a className="oo-share oo-control-bar-item"
           onClick={this.handleShareClick} key="share" aria-hidden="true">
@@ -750,6 +794,11 @@ var ControlBar = React.createClass({
       if (defaultItems[k].name === "live" &&
         (typeof this.props.isLiveStream === 'undefined' ||
           !(this.props.isLiveStream))) {
+        continue;
+      }
+
+      if (defaultItems[k].name === "multiAudio" &&
+        !this.props.controller.state.multiAudio) {
         continue;
       }
 
--- a/js/components/multi-audio/multiAudioPopover.js
+++ b/js/components/multi-audio/multiAudioPopover.js
@@ -1,0 +1,62 @@
+var React = require('react'),
+  CONSTANTS = require('../../constants/constants'),
+  AccessibleMenu = require('../higher-order/accessibleMenu'),
+  CloseButton = require('../closeButton');
+
+var MultiAudioPopover = React.createClass({
+  /**
+   * close the popover
+   */
+  handleClose: function() {
+    this.props.togglePopoverAction({
+      restoreToggleButtonFocus: true
+    });
+  },
+
+  /**
+   * Call setCurrentAudio from controller
+   * @param {String} id - id of the audio track to activate
+   */
+  handleMultiAudioClick: function(id) {
+    //change audio only if try to set not current audio track
+    if (this.props.controller && this.props.controller.state.currentAudioId !== id) {
+      if (typeof this.props.controller.setCurrentAudio === 'function') {
+        this.props.controller.setCurrentAudio(id);
+      }
+    }
+  },
+
+  render: function() {
+    var multiAudio = [];
+    if (this.props.controller && this.props.controller.state && !!this.props.controller.state.multiAudio) {
+      multiAudio = this.props.controller.state.multiAudio.multiAudio;
+    }
+    return (
+      <ul className="oo-popover-horizontal" role="menu">
+        <li role="presentation">
+          <CloseButton
+            {...this.props}
+            role={CONSTANTS.ARIA_ROLES.MENU_ITEM}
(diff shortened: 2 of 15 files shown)


js/components/controlBar.js
```diff
@@ -232,6 +233,13 @@ var ControlBar = React.createClass({
     }
   },
 
+  handleMultiAudioClick: function () {
+    this.configureMenuAutofocus(CONSTANTS.MENU_OPTIONS.MULTI_AUDIO);
+
+    this.togglePopover(CONSTANTS.MENU_OPTIONS.MULTI_AUDIO);
+    this.closePopovers();
```
jpfranco: I think we should probably update `this. closePopovers()` in order to include the newly added one. When calling `this. closePopovers()`  I would expect it would close all (multi audio included). We could possibly add a parameter to that function in order allow closing all except the one that is specified, that way we can avoid manually closing each other popover when a new one is opened.

js/components/controlBar.js
```diff
@@ -623,6 +637,36 @@ var ControlBar = React.createClass({
...
+              <Icon {...this.props} icon="multiAudio" style={dynamicStyles.iconCharacter}
+                    onMouseOver={this.highlight} onMouseOut={this.removeHighlight} />
+              <Tooltip enabled={isTooltipEnabled} text={Utils.getLocalizedString(this.props.language, CONSTANTS.SKIN_TEXT.MULTI_AUDIO, this.props.localizableStrings)} responsivenessMultiplier={this.responsiveUIMultiple} bottom={this.responsiveUIMultiple * this.props.skinConfig.controlBar.height} alignment={alignment} />
+            </AccessibleButton>
+            {this.props.controller.state.multiAudioOptions.showPopover &&
+            <Popover
+              popoverClassName="oo-popover oo-popover-pull-right"
+              autoFocus={this.props.controller.state.multiAudioOptions.autoFocus}
+              closeActionEnabled={this.props.controller.state.accessibilityControlsEnabled}
+              closeAction={this.closePopover.bind(this, CONSTANTS.MENU_OPTIONS.MULTI_AUDIO)}>
```
jpfranco: I think these binds are cleaner than creating a new function specifically for multi audio, and we can keep them, but in general we should avoid these whenever possible because they will be creating a new function on each render.

js/components/multi-audio/multiAudioPopover.js
```diff
@@ -0,0 +1,62 @@
...
+  CONSTANTS = require('../../constants/constants'),
+  AccessibleMenu = require('../higher-order/accessibleMenu'),
+  CloseButton = require('../closeButton');
+
+var MultiAudioPopover = React.createClass({
+  /**
+   * close the popover
+   */
+  handleClose: function() {
+    this.props.togglePopoverAction({
```
jpfranco: Do we need a null check here?

js/components/multi-audio/multiAudioPopover.js
```diff
@@ -0,0 +1,62 @@
...
+
+  render: function() {
+    var multiAudio = [];
+    if (this.props.controller && this.props.controller.state && !!this.props.controller.state.multiAudio) {
+      multiAudio = this.props.controller.state.multiAudio.multiAudio;
+    }
+    return (
+      <ul className="oo-popover-horizontal" role="menu">
+        <li role="presentation">
+          <CloseButton
```
jpfranco: If this is meant to be accessible then it might make sense to this button to the end of the list, otherwise users navigating with the keyboard will focus on the close button before reaching the audio options.

js/components/multi-audio/multiAudioPopover.js
```diff
@@ -0,0 +1,62 @@
...
+            {...this.props}
+            role={CONSTANTS.ARIA_ROLES.MENU_ITEM}
+            closeAction={this.handleClose} />
+        </li>
+        {multiAudio.map(function(audio){
+          return (
+            <li>
+              <span
```
jpfranco: Should this be an `AccessibleButton`?

js/components/multi-audio/multiAudioPopover.js
```diff
@@ -0,0 +1,62 @@
...
+            </li>
+          );
+        }, this)}
+      </ul>
+    );
+  }
+});
+
+// Extend with AccessibleMenu features
+MultiAudioPopover = AccessibleMenu(MultiAudioPopover);
```
jpfranco: The `AccessibleMenu` functionality only works when used in conjunction with `AccessibleButton` components, so we'd need to update the `oo-multiaudio-elements` above. Please take a look at the `AccessibleMenu` implementation

js/components/multi-audio/multiAudioPopover.js
```diff
@@ -0,0 +1,62 @@
...
+
+module.exports = MultiAudioPopover;
```
jpfranco: Can we add `PropTypes`?

js/components/multi-audio/multiAudioPopover.js
```diff
@@ -0,0 +1,62 @@
...
+      if (typeof this.props.controller.setCurrentAudio === 'function') {
+        this.props.controller.setCurrentAudio(id);
+      }
+    }
+  },
+
+  render: function() {
+    var multiAudio = [];
+    if (this.props.controller && this.props.controller.state && !!this.props.controller.state.multiAudio) {
+      multiAudio = this.props.controller.state.multiAudio.multiAudio;
```
jpfranco: Should `multiAudio.multiAudio` be something like `multiAudio.tracks`? It's not clear what the second `multiAudio` refers to.

js/controller.js
```diff
@@ -859,7 +867,24 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
-    //  TODO: code for a video with multiaudio should be here
+      this.state.multiAudio = multiAudio;
+    },
+
+    /**
+     * @fires OO.EVENTS.SET_CURRENT_AUDIO
+     * @param {String} id - the id of the audio track to activate
+     */
+    setCurrentAudio: function(id) {
+      this.mb.publish(OO.EVENTS.SET_CURRENT_AUDIO, id);
```
jpfranco: Should the event be fired after the current audio is set?

tests/components/controlBar-test.js
```diff
@@ -934,14 +952,92 @@ describe('ControlBar', function () {
     expect(ccButtons.length).toBe(1);
   });
 
+  it('shows/hides multiaudio button if multiaudio available', function() {
+    var mockController = {
```
jpfranco: Please use `baseMockController` and `baseMockProps` instead of defining a new object on each test.

tests/components/controlBar-test.js
```diff
@@ -934,14 +952,92 @@ describe('ControlBar', function () {
...
+        },
+        closedCaptionOptions: {},
+        multiAudioOptions: {},
+        videoQualityOptions: {
+          availableBitrates: null
+        }
+      }
+    };
+
+    var oneButtonSkinConfig = Utils.clone(skinConfig);
```
jpfranco: Please note that this creates a shallow copy, so if you modify any nested properties you will still be modifying the original object. This limitation is already addressed by `baseMockProps`, so that's another reason to switch to that.

js/components/controlBar.js
```diff
@@ -753,6 +797,11 @@ var ControlBar = React.createClass({
         continue;
       }
 
+      if (defaultItems[k].name === "multiAudio" &&
+        !this.props.controller.state.multiAudio) {
```
aeng7: Is this state defined yet? Or will this be defined in a future commit?

js/components/controlBar.js
```diff
@@ -478,6 +486,12 @@ var ControlBar = React.createClass({
       "oo-selected": this.props.controller.state.closedCaptionOptions.showPopover
     });
 
+    var multiAudioClass = ClassNames({
+      "oo-multiaudio": true,
+      "oo-control-bar-item": true,
+      "oo-selected": this.props.controller.state.multiAudioOptions.showPopover
```
aeng7: Is this state defined yet? Or will this be defined in a future commit? I don't see ``this.props.controller.state.multiAudioOptions`` defined on the master branch, so this will throw a null pointer exception if this is merged as is.

js/components/multi-audio/multiAudioPopover.js
```diff
@@ -0,0 +1,62 @@
...
+    });
+  },
+
+  /**
+   * Call setCurrentAudio from controller
+   * @param {String} id - id of the audio track to activate
+   */
+  handleMultiAudioClick: function(id) {
+    //change audio only if try to set not current audio track
+    if (this.props.controller && this.props.controller.state.currentAudioId !== id) {
```
aeng7: Should we move this check into the ``setCurrentAudio`` function in the controller? I feel like it would be better for the controller to check its own state and have ``multiAudioPopover`` be less dependent on the controllers state.

js/components/multi-audio/multiAudioPopover.js
aeng7: Is it possible to pass this in to this component as prop? This would make this component less dependent on the controller.

js/components/controlBar.js
KseniiaSokolovaEpam: @jpfranco yes, I wanted to do it, but now we do not have approved UI, and maybe the design will be changed (maybe there will be no popovers or be only one).

js/components/multi-audio/multiAudioPopover.js
KseniiaSokolovaEpam: no

js/components/multi-audio/multiAudioPopover.js
KseniiaSokolovaEpam: @jpfranco There is only fish for development (not real interface. The interface will be after approving).

js/components/multi-audio/multiAudioPopover.js
KseniiaSokolovaEpam: @jpfranco There is only "fish" for development (not real interface. The interface will be after approving).

js/components/multi-audio/multiAudioPopover.js
KseniiaSokolovaEpam: @jpfranco Ok, i will remember it.

js/components/multi-audio/multiAudioPopover.js
KseniiaSokolovaEpam: \+

js/components/multi-audio/multiAudioPopover.js
KseniiaSokolovaEpam: \+

js/controller.js
KseniiaSokolovaEpam: This event are fired after, for example, user has clicked to audio button and chose new audio.

tests/components/controlBar-test.js
KseniiaSokolovaEpam: \+

tests/components/controlBar-test.js
KseniiaSokolovaEpam: \+

js/components/multi-audio/multiAudioPopover.js
KseniiaSokolovaEpam: \+

js/components/multi-audio/multiAudioPopover.js
KseniiaSokolovaEpam: @aeng7 But we still will use functions from controller in this component. Do you really think that it is necessary? Thank you.

js/components/controlBar.js
jpfranco: @KseniiaSokolovaEpam I think there's a good chance that we'll forget to add this later if the UI does end up using a popover. It might be better to include this earlier, but I leave it up to you

js/components/multi-audio/multiAudioPopover.js
jpfranco: Why not? Someone else could use this component later on and not pass this function. This parameter is also not listed as required in the proptypes

js/components/multi-audio/multiAudioPopover.js
jpfranco: Thanks

js/components/multi-audio/multiAudioPopover.js
aeng7: Having these defined in the propTypes is fine as well. You're correct in that we depend on a lot of functions in the controller. We should still try to isolate things when possible.

js/components/multi-audio/multiAudioPopover.js
KseniiaSokolovaEpam: \+

KseniiaSokolovaEpam:
I switched "config" to master branch.

PR #1013, opened 2018-02-22
[PLAYER-2102] Preventing control bar from flashing during video transitions

jpfranco:
[Ticket](https://jira.corp.ooyala.com/browse/PLAYER-2102)

This will basically just prevent the control bar from flashing briefly when a new Discovery or Playlist video is first shown. These changes also prevent the start screen from flashing during video transitions. This last issue had already been fixed, but was re-introduced recently.

I also refactored some old unit tests in order to avoid re-declaring a large mock object on each separate test.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -886,8 +886,10 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       if (this.state.afterOoyalaAd) {
         this.state.screenToShow = CONSTANTS.SCREEN.LOADING_SCREEN;
       } else {
-        // If the core tells us that it will autoplay then we just display the loading
-        // spinner, otherwise we need to render the big play button.
+        // If the core tells us that it will autoplay then we display the loading
+        // spinner (with the thumbnail and description if it's the first video, or just a
+        // black background when transitioning to a new one), otherwise we need to render
+        // the big play button.
         if (params.willAutoplay) {
           // Show just the loading spinner when transitioning to another Discovery or
           // Playlist video, otherwise show the spinner and thumbnail.
--- a/js/controller.js
+++ b/js/controller.js
@@ -1622,6 +1622,10 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
           break;
         case CONSTANTS.STATE.PLAYING:
           this.isNewVrVideo = false;
+          // In some cases the controlBarVisible var might get out of sync if not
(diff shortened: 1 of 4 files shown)


js/controller.js
```diff
@@ -889,7 +889,13 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
         // If the core tells us that it will autoplay then we just display the loading
         // spinner, otherwise we need to render the big play button.
```
aeng7: Do we need to update this comment?

tests/views/playingScreen-test.js
```diff
@@ -9,39 +9,46 @@ var PlayingScreen = require('../../js/views/playingScreen');
...
-
-    var closedCaptionOptions = {
+    closedCaptionOptions = {
       cueText: "cue text"
     };
+  });
+
+  it('creates a PlayingScreen and checks mouseMove, mouseUp without video360', function () {
+    var isMoved = false
+      , isTouched = false;
```
aeng7: Should we be checking this variable in this unit test? It looks to be unused right now.

js/controller.js
jpfranco: Yes, will update

tests/views/playingScreen-test.js
jpfranco: Not sure, these are mostly old unit tests which I refactored in order to use `mockController` instead of declaring one each time, but I I'll check and remove if necessary.

aeng7:
Thanks for removing the mock definitions from each individual test

aeng7:
LGTM outside of a few minor comments

tests/views/playingScreen-test.js
jpfranco: @aeng7 so I think this unit test is wrong. It looks like it might be testing that these events are not fired for regular videos, however, this is not tested properly because the event listeners are being added to the `document`, so they are not triggered by `TestUtils.Simulate`. If I fix this in order to use DOM events the unit test fails because the events are actually being triggered even without 360 video. I'm not sure if it would be better to just remove this.

aeng7: @jpfranco should we have EPAM look at this unit test?

jpfranco: @aeng7 yeah I think that would probably be better. I might be misunderstanding what this does.

PR #1014, opened 2018-02-27
Pbw 7142

pilievOoyala:


js/components/utils.js
```diff
@@ -496,9 +496,13 @@ var Utils = {
         currentProp = props[i];
         currentObject = value = currentObject[currentProp];
       }
-      return value || defaltValue;
+
+      if (value === undefined) {
```
aeng7: Value could be null (for example, if currentObject[currentProp] is null), and we would return null instead of the defaultValue. Is this what we wanted?

jpfranco: I think that since `null` is deliberately set, it could be considered to be the property value and thus shouldn't be replaced with the default, although in practice it might cause some confusion. Either way works for me as long as we specify it in the jsdoc, but I agree that we should use `typeof value === 'undefined'`

pilievOoyala: I'm going to use hasOwnProperty instead, since you could set it to anything and what we really care about is that the key exists in the object.

pilievOoyala: I'm going with your solution because the hasOwnProperty has issues with variables from prototypes.

tests/components/utils-test.js
```diff
@@ -381,6 +381,19 @@ describe('Utils', function () {
...
+    var falsyVal = Utils.getPropertyValue({ property: { nestedProp: 0 } }, 'property.nestedProp', 1);
+    expect(falsyVal === 0).toEqual(true);
+
+    falsyVal = Utils.getPropertyValue({ property: { nestedProp: null } }, 'property.nestedProp', 1);
+    expect(falsyVal === null).toEqual(true);
+
+    var stringVal = Utils.getPropertyValue({ property: { nestedProp: { test:'test' } } }, 'property.nestedProp.test', 1);
+    expect(stringVal === 'test').toEqual(true);
+
+    functionVal = Utils.getPropertyValue({ property: { nestedProp: { test:'test' } } }, 'property.nestedProp.toString', 1);
```
aeng7: Should we test a newly defined function in addition to toString, which is already part of an object's prototype? We can also try calling this function returned by Utils.getPropertyValue to see if it works properly.

js/components/utils.js
```diff
@@ -481,10 +481,10 @@ var Utils = {
    * @function getPropertyValue
    * @param {Object} object - The object we want to extract the property form
    * @param {String} propertyPath - A path that points to a nested property in the object with a form like 'prop.nestedProp1.nestedProp2'
-   * @param {Object} defaltValue - (Optional) A default value to return when the property is undefined
+   * @param {Object} defaultValue - (Optional) A default value to return when the property is undefined
```
jpfranco: Since @aeng7 had a comment, can we also add a note here in order to clarify that `null` will be returned instead of `defaultValue`?

tests/components/utils-test.js
pilievOoyala: Sure. I'll add it.

js/components/utils.js
pilievOoyala: Will do.

aeng7:
LGTM

PR #1015, opened 2018-03-01
Bugfix/player 3003

yvinokurov:
This PR is the solution of the problem "https://jira.corp.ooyala.com/browse/PLAYER-3003".
Description of the problem.
Native fullscreen on IOS on the forefront of the outputs of the video. If video360 is played, the video stream is projected onto the canvas element.
Therefore, the user watches the black screen.

Solution.
We used the solution from the bitmovin. In this solution, the canvas element is on the forefront. 
If you press the "Fullscreen" button, the event "TOGGLE_FULLSCREEN_VR" will be sent for video360 on IOS.
Also, for a better understanding, some of the logic from the "toggleFullscreen" method has been put into separate methods. The logic did not change.

We added unit tests.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -1430,8 +1430,8 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
      * @private
      */
     toggleIOSNativeFullscreen: function () {
-      if(this.state.fullscreen) {
-        this.state.mainVideoInnerWrapper.removeClass('oo-fullscreen');
+      if (this.state.fullscreen) {
+        this.state.mainVideoElement.webkitExitFullscreen();
       } else {
         this.state.mainVideoElement.webkitEnterFullscreen();
       }
--- a/js/controller.js
+++ b/js/controller.js
@@ -1412,13 +1412,13 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       }
       // partial support, video element only (iOS) and not vr
       else if (this.state.isVideoFullScreenSupported && !this.videoVr) {
-        this.fullscreenIosNative();
+        this.enterIOSNativeFullscreen();
       } else { // no support
         if (this.videoVr) {
           // if videoVr we don't use native fullscreen (IOS)
           this.mb.publish(OO.EVENTS.TOGGLE_FULLSCREEN_VR, this.focusedElement);
         }
-        this.noSupportFullscreen();
+        this.enterPseudoFullscreenMode();
       }
 
       this.state.fullscreen = !this.state.fullscreen;
@@ -1429,7 +1429,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
      * Toggle fullscreen for video element only (iOS)
      * @private
      */
-    fullscreenIosNative: function () {
+    enterIOSNativeFullscreen: function () {
       if(this.state.fullscreen) {
         this.state.mainVideoInnerWrapper.removeClass('oo-fullscreen');
       } else {
@@ -1441,7 +1441,7 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
      * Toggle fullscreen if native fullscreen is not supported (pseudo fullscreen)
      * @private
      */
-    noSupportFullscreen: function () {
+    enterPseudoFullscreenMode: function () {
       if(this.state.isFullWindow) {
         this.exitFullWindow();
       } else {
--- a/js/controller.js
+++ b/js/controller.js
@@ -1412,13 +1412,13 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
       }
       // partial support, video element only (iOS) and not vr
       else if (this.state.isVideoFullScreenSupported && !this.videoVr) {
-        this.enterIOSNativeFullscreen();
+        this.toggleIOSNativeFullscreen();
       } else { // no support
         if (this.videoVr) {
           // if videoVr we don't use native fullscreen (IOS)
           this.mb.publish(OO.EVENTS.TOGGLE_FULLSCREEN_VR, this.focusedElement);
         }
-        this.enterPseudoFullscreenMode();
+        this.togglePseudoFullscreenMode();
       }
 
       this.state.fullscreen = !this.state.fullscreen;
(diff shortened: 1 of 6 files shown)


js/controller.js
```diff
@@ -1410,24 +1410,45 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+      } else {
+        this.state.mainVideoElement.webkitEnterFullscreen();
+      }
+    },
+
+    /**
+     * Toggle fullscreen if native fullscreen is not supported (pseudo fullscreen)
+     * @private
+     */
+    noSupportFullscreen: function () {
```
IvanSakharovskii: rename function `enterPseudoFullscreenMode`

js/controller.js
```diff
@@ -1410,24 +1410,45 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+
       this.state.fullscreen = !this.state.fullscreen;
       this.renderSkin();
     },
 
+    /**
+     * Toggle fullscreen for video element only (iOS)
+     * @private
+     */
+    fullscreenIosNative: function () {
```
IvanSakharovskii: rename function `enterIOSNativeFullscreen`

js/controller.js
yvinokurov: fixed

js/controller.js
yvinokurov: fixed

js/controller.js
```diff
@@ -1410,24 +1410,45 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+      } else {
+        this.state.mainVideoElement.webkitEnterFullscreen();
+      }
+    },
+
+    /**
+     * Toggle fullscreen if native fullscreen is not supported (pseudo fullscreen)
+     * @private
+     */
+    enterPseudoFullscreenMode: function () {
```
jpfranco: The word `enter` here is misleading, this should be `toggle`

js/controller.js
```diff
@@ -1410,24 +1410,45 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
       this.renderSkin();
     },
 
+    /**
+     * Toggle fullscreen for video element only (iOS)
+     * @private
+     */
+    enterIOSNativeFullscreen: function () {
+      if(this.state.fullscreen) {
+        this.state.mainVideoInnerWrapper.removeClass('oo-fullscreen');
```
jpfranco: Why are we now removing the class instead of calling `webkitExitFullscreen()`?

scss/skin/_default.scss
```diff
@@ -47,12 +47,14 @@
...
-    position:fixed;
-    overflow: hidden;
-    top:0;
-    left:0;
-    width:100%;
-    height:100%;
+    width: 100%!important;
+    height: 100%!important;
+    top: 0px;
+    bottom: 0px;
```
jpfranco: Should bottom be 100%? Having both top and bottom at 0 would create a flat line if height wasn't 100%/

scss/skin/_default.scss
```diff
@@ -47,12 +47,14 @@
 
   // "full window" style, if fullscreen is not supported natively
   &.oo-fullscreen {
-    position:fixed;
-    overflow: hidden;
-    top:0;
-    left:0;
-    width:100%;
-    height:100%;
+    width: 100%!important;
```
jpfranco: Please check to see if `!important` is really necessary in these styles. If not, please remove.

tests/html5skin-test.js
```diff
@@ -679,5 +679,17 @@ describe('Controller', function() {
       expect(controller.state.videoQualityOptions.availableBitrates).toBe(qualities.bitrates);
     });
 
+    describe('Toggle fullscreen', function () {
```
jpfranco: This `describe` is for tests related to video qualities. This test should be moved to a different section

js/controller.js
```diff
@@ -1410,24 +1410,45 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
+
       this.state.fullscreen = !this.state.fullscreen;
       this.renderSkin();
     },
 
+    /**
+     * Toggle fullscreen for video element only (iOS)
+     * @private
+     */
+    enterIOSNativeFullscreen: function () {
```
jpfranco: This should be `toggle` instead of `enter`

js/controller.js
yvinokurov: fixed

js/controller.js
yvinokurov: fixed

js/controller.js
yvinokurov: fixed

tests/html5skin-test.js
yvinokurov: fixed

scss/skin/_default.scss
yvinokurov: we refactored css

scss/skin/_default.scss
yvinokurov: we refactored css

jpfranco:
Need another review when you get a chance @aeng7 @pilievOoyala @dallen42

js/controller.js
```diff
@@ -1410,24 +1410,45 @@ OO.plugin("Html5Skin", function (OO, _, $, W) {
...
         Fullscreen.toggle(this.state.mainVideoInnerWrapper.get(0));
       }
-      // partial support, video element only (iOS)
-      else if (this.state.isVideoFullScreenSupported) {
-        if (this.state.fullscreen) {
-          this.state.mainVideoElement.webkitExitFullscreen();
-        } else {
-          this.state.mainVideoElement.webkitEnterFullscreen();
-        }
+      // partial support, video element only (iOS) and not vr
```
aeng7: Should we have any is iOS checks for any of these iOS only situations?

aeng7:
LGTM, other than a small comment. I'll merge now but please check the comment.

js/controller.js
yvinokurov: I think that this is not necessary. This code based on the properties of the system. It worked earlier, before this PR. In this PR, we added check for video360, because standard fullscreen does not work correctly with video360.

PR #1020, opened 2018-03-02
Feature/player 3068

KseniiaSokolovaEpam:
Add flag ("showMultiAudioIcon") that disables multi-audio functionality to avoid confusions fromusers while this part is not fully ready.

controller.state.showMultiAudioIcon = true; was added for tests in feature/PLAYER-2946


aeng7:
Can we add a unit test for this? Or is this best saved for when more of the implementation is completed?

aeng7:
LGTM

PR #1021, opened 2018-03-02
Feature/player 2946 to master

KseniiaSokolovaEpam:
Task - https://jira.corp.ooyala.com/browse/PLAYER-2946
"Change current languge when user changes it, choosen language should be saved in a Storage"

Set track from Storage if the track id is saved in Storage.
Enabled track is colored in blue. (without UI, only for preview).

Unit tests added.
Unit tests were run without errors.

### DIFF
--- a/js/components/multi-audio/multiAudioPopover.js
+++ b/js/components/multi-audio/multiAudioPopover.js
@@ -26,7 +26,10 @@ var MultiAudioPopover = React.createClass({
 
   render: function() {
     var multiAudio = [];
-    if (this.props.controller && this.props.controller.state && !!this.props.controller.state.multiAudio) {
+    if (this.props.controller &&
+      this.props.controller.state &&
+      !!this.props.controller.state.multiAudio &&
+      Array.isArray(this.props.controller.state.multiAudio.tracks) ) {
       multiAudio = this.props.controller.state.multiAudio.tracks;
     }
     return (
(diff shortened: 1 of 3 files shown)


js/components/multi-audio/multiAudioPopover.js
```diff
@@ -38,11 +41,17 @@ var MultiAudioPopover = React.createClass({
             closeAction={this.handleClose} />
         </li>
         {multiAudio.map(function(audio){
+          var color = "white";
+          if (audio.enabled) {
+            color = "blue";
```
aeng7: Should we have this color be configurable?

aeng7:
LGTM other than a small comment.

js/components/multi-audio/multiAudioPopover.js
DmitriiMaslov: @aeng7 this will be implemented in another pull requests. For such customization we have the following ticket - https://jira.corp.ooyala.com/browse/PLAYER-543

js/components/multi-audio/multiAudioPopover.js
```diff
@@ -38,11 +41,17 @@ var MultiAudioPopover = React.createClass({
...
+          }
+
           return (
             <li>
               <span
                 className="oo-multiaudio-element"
                 onClick={this.handleMultiAudioClick.bind(this, audio.id)}
+                style={{color:color}}
```
jpfranco: I know that this is temporary, but I think this should be a class instead of a style. The customer might want to customize the appearance of selected items. I think we should also add this scenario to the unit tests of the multi audio popover component.

yvinokurov: Now, user doesn’t see a functionality of multiaudio. It was added for checking and testing by the developers. In tasks “https://jira.corp.ooyala.com/browse/PLAYER-542” and “https://jira.corp.ooyala.com/browse/PLAYER-522" the interface will be completely redesigned and css classes will be added. Now, it makes no sense to refactor and add unit tests.

PR #1038, opened 2018-03-27
Add ESLint and ESLint run command in package.json

yaroslavgribov:
# Description

This PR introduces a new feature - ESLint(Linting utility for js and jsx). It's not in any case a single source of truth, it's just and attempt to start a discussion. Some of the rules are taken from the styleguide, some are just recommended and community-wide adopted best practices. We could add more rules as we progress.

## Summary

ESLint is an awesome tool that could allow us enforce code style rules, reduce the number of code style comments in PR and help us write better code overall. It has integrations with pretty much every editor and is really easy to setup and use.

## Purpose 

The purpose of this PR is to introduce the feature, essentially it does not affect any aspect of the development at this point, but it might be used later at the build stage, CI, etc.

## Installing

ESLint can be installed both globally and locally but it is recommended to install it locally (same as with gulp at this point). Just run `npm install` and it should work.

## Running

`npm run eslint` checks *only staged* changes on `.js` files in the current repository. This way it's easier to adopt it, since we don't have to fix every conflict on every file, we could just start with Litter-Pickup refactoring. 
It does not check code style rules violations on every file in the project, it just checks the most recent changes. It's considered to be a common practice to run it as a pre-commit hook, so I went forward and created `scripts/git/pre-commit`. We can link the file later or copy it manually to reside in `.git/hooks` folder, as the `.git` folder is not tracked by VCS.

If you want to see it in action - please try to make changes to some js files and run `npm run eslint`.

## More read:

- https://eslint.org/docs/about/ - official docs
- https://medium.com/the-node-js-collection/why-and-how-to-use-eslint-in-your-project-742d0bc61ed7 - a nice article by Node.js foundation

- https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks - info on git-hooks.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -19,7 +19,7 @@
         "no-unreachable": ["error"],
         "no-unused-vars": ["warn"],
         "valid-typeof": ["warn"],
-        "space-before-function-paren": ["warn"],
+        "space-before-function-paren": ["never"],
         "space-before-blocks": ["warn"],
         "spaced-comment": ["warn"],
         "brace-style": ["warn"],
--- a/.eslintrc
+++ b/.eslintrc
@@ -23,7 +23,7 @@
         "spaced-comment": ["warn"],
         "brace-style": ["warn"],
         "one-var-declaration-per-line": ["warn"],
-        "no-magic-numbers": ["warn"],
+        "no-magic-numbers": ["warn", { "ignore": [0, -1, 1] }],
         "valid-jsdoc": ["warn"],
         
         "max-len": ["error", 110],
--- a/.eslintrc
+++ b/.eslintrc
@@ -19,7 +19,6 @@
         "no-unreachable": ["error"],
         "no-unused-vars": ["warn"],
         "valid-typeof": ["warn"],
-        "space-before-function-paren": ["never"],
         "space-before-blocks": ["warn"],
         "spaced-comment": ["warn"],
         "brace-style": ["warn"],
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,45 @@
+{
+    "env": {
+        "browser": true,
+        "commonjs": true,
+        "es6": false,
+        "node": false
+    },
+    "parserOptions": {
+        "ecmaFeatures": {
+            "jsx": true
+        },
+        "ecmaVersion": 5,
+        "sourceType": "module"
+    },
+    "plugins": [
+        "react"
+    ],
+    "rules": {
+        "no-unreachable": ["error"],
+        "no-unused-vars": ["warn"],
+        "valid-typeof": ["warn"],
+        "space-before-function-paren": ["warn"],
+        "space-before-blocks": ["warn"],
+        "spaced-comment": ["warn"],
+        "brace-style": ["warn"],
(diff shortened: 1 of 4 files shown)


jpfranco:
I like the approach of running against the most recent changes, this will be very useful for adoption. I think there might be an issue with the command though, I get errors if I modify multiple files:

```
C02S21K6G8WP:html5-skin jfranco$ npm run eslint

> html5-skin@0.0.1 eslint /Users/jfranco/repos/html5-skin
> LIST=`git diff --name-only | grep -f *.js | grep -v json;`; if [ $LIST ]; then eslint $LIST; fi

sh: line 0: [: js/components/accessibleButton.js: unary operator expected
C02S21K6G8WP:html5-skin jfranco$
```

With a single file it's working fine.

yaroslavgribov:
Thanks @jpfranco I couldn't find a easy way to check a list of files using this approach, so I added a bash script to check for changes. I also updated the comment above with the explanation, please take a look.

pilievOoyala:
Looks great. I went through the list and I like all the rules. I think we should add 
"space-before-function-paren": ["error", "never"],
"space-in-parens": ["error", "never"],
"space-before-blocks": ["error", "always"],
"no-multiple-empty-lines":["max": 2],
 "require-jsdoc": ["error", {
        "require": {
            "FunctionDeclaration": true,
            "MethodDefinition": true,
            "ClassDeclaration": true,
            "ArrowFunctionExpression": true,
            "FunctionExpression": true
        }
    }]

I'm not sure about the jsdoc one, what happens for anonymous functions that get passed into this like jQuery's find? But we can try it out. These should agree with our coding style but @jpfranco  @aeng7  you can correct me if I'm wrong.

I would also like to add the https://eslint.org/docs/rules/yoda rule.  Only because it's a star wars reference :)

yaroslavgribov:
Those are really good rules as well! The latest version of eslint that's supported on our node is 2.13.1, and it doesn't support `"no-multiple-empty-lines":["max", 2]`. So I added `"no-multiple-empty-lines":["error"]` that seems to do the same thing. Also `"FunctionDeclaration": true,`
`"ArrowFunctionExpression": true` is not supported as per [CHANGELOG](https://github.com/eslint/eslint/blob/master/CHANGELOG.md).

Also added yoda 🦄.

pilievOoyala:
LGTM. Just need one more approval.  @jpfranco @aeng7  Whenever you have time.

.eslintrc
```diff
@@ -0,0 +1,50 @@
...
+        "valid-typeof": ["warn"],
+        "space-before-blocks": ["warn"],
+        "keyword-spacing": ["warn"],
+        "spaced-comment": ["warn"],
+        "brace-style": ["warn"],
+        "one-var-declaration-per-line": ["warn"],
+        "no-magic-numbers": ["warn", { "ignore": [0, -1, 1] }],
+        "valid-jsdoc": ["warn"],
+        "camelcase": ["warn"],
+
```
aeng7: Is there a reason for these extra empty lines? Just out of curiosity. It doesn't immediately look like any groupings of any sort.

PR #1040, opened 2018-03-30
[feat/PLAYER-3234] Hide CC column in new Multi-Audio menu

KseniiaSokolovaEpam:
We hide CC column in new Multi-Audio menu for the first release only. 
(When a new design is existed and CC settings are moved to menu as well then CC column will be shown again.)
Old unit tests are passed.
There is no nessaccary to create new tests.

### DIFF
--- a/js/components/closed-caption-multi-audio-menu/closedCaptionMultiAudioMenu.js
+++ b/js/components/closed-caption-multi-audio-menu/closedCaptionMultiAudioMenu.js
@@ -80,6 +80,7 @@ var ClosedCaptionMultiAudioMenu = React.createClass({
       );
       closedCaptionsCol = (
         <Tab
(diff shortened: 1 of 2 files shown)


pilievOoyala:
LGTM. @jpfranco  Can you review as well.

js/components/closed-caption-multi-audio-menu/closedCaptionMultiAudioMenu.js
```diff
@@ -80,6 +80,7 @@ var ClosedCaptionMultiAudioMenu = React.createClass({
       );
       closedCaptionsCol = (
         <Tab
+          tabClassName="oo-hidden"
```
jpfranco: I think it's fine to just call this `className`

PR #1046, opened 2018-04-06
[feature/nvmrc] Add nvmrc similar to core

yaroslavgribov:
Since we have a need to update nodejs, I think it might be a good idea to have .nvmrc shared across all repositories. We already use one in core.

Added `engines` field in `package.json`.
[More on engines](https://docs.npmjs.com/files/package.json#engines)


.nvmrc
```diff
@@ -0,0 +1 @@
+v0.10.40
```
jpfranco: I recommend switching to v0.10.46, which is what our Jenkins build jobs use, but LGTM otherwise.

yaroslavgribov: Oh I didn't know we could use 0.10.46, I just copied one from https://git.corp.ooyala.com/projects/PBW/repos/mjolnir/browse/.nvmrc. 
Are there any major differences between the two?

yaroslavgribov:
@jpfranco @pilievOoyala @aeng7 @aldigun 
Could someone please take a look at this one?

.nvmrc
jpfranco: Probably not, but I think it's better for our dev environment to use the same version as our build environment in order to avoid package version differences when doing npm install, etc.

pilievOoyala:
LGTM

PR #1049, opened 2018-04-09
Codestyle - ESLint --fix

yaroslavgribov:
Well this is a big one.

Fixes all problems that can be auto-fixed with `eslint --fix` so we don't have to run it.
I still had to run it with `--no-verify` option, because not all problems can be auto-fixed.
However it reduced 
image
to
image
in `js/` folder
and 
image
to 
image
in `tests/` folder.

The alternative approach is to fix it by hand every time someone touches the files.

- [x] All unit tests passed

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,6 +2,7 @@
   "env": {
     "browser": true,
     "commonjs": true,
+    "jest": true,
     "es6": false,
     "node": false
   },
@@ -56,6 +57,8 @@
     ],
     "yoda": ["error"],
 
-    "react/jsx-uses-vars": ["error"]
+    "react/jsx-uses-vars": ["error"],
+    "react/jsx-uses-react": ["error"],
+    "react/react-in-jsx-scope": ["error"]
   }
 }
--- a/js/components/accessibilityControls.js
+++ b/js/components/accessibilityControls.js
@@ -1,7 +1,7 @@
 var CONSTANTS = require('./../constants/constants');
 var Utils = require('./utils');
 
-var AccessibilityControls = function (controller) {
+var AccessibilityControls = function(controller) {
   this.controller = controller;
   this.vrRotationAllowed = true; // flag for checking repeat of keyDown
   this.keyDirectionMap = {};
@@ -13,20 +13,20 @@ var AccessibilityControls = function (controller) {
     seekRate: 1,
     lastKeyDownTime: 0,
   };
-  this.prevKeyPressedArr = []; //list of codes of pressed buttons
+  this.prevKeyPressedArr = []; // list of codes of pressed buttons
   this.keyEventDown = this.keyEventDown.bind(this);
   this.keyEventUp = this.keyEventUp.bind(this);
   this.moveVrToDirection = this.moveVrToDirection.bind(this);
   this.getTargetTagName = this.getTargetTagName.bind(this);
 
-  document.addEventListener("keydown", this.keyEventDown);
-  document.addEventListener("keyup", this.keyEventUp);
+  document.addEventListener('keydown', this.keyEventDown);
+  document.addEventListener('keyup', this.keyEventUp);
 };
 
 AccessibilityControls.prototype = {
   cleanUp : function() {
-    document.removeEventListener("keydown", this.keyEventDown);
-    document.removeEventListener("keyup", this.keyEventUp);
+    document.removeEventListener('keydown', this.keyEventDown);
+    document.removeEventListener('keyup', this.keyEventUp);
   },
 
   keyEventDown: function(e) {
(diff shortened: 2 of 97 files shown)


.eslintrc
```diff
@@ -2,9 +2,13 @@
   "env": {
     "browser": true,
     "commonjs": true,
+    "jest": true,
     "es6": false,
     "node": false
   },
+  "globals": {
+      "OO": true
```
yaroslavgribov: This line adds "OO" as accepted global variable.

.eslintrc
```diff
@@ -16,7 +20,7 @@
   "rules": {
     "brace-style": ["warn"],
     "camelcase": ["warn"],
-    "id-length": ["warn"],
+    "id-length": ["warn", { "exceptions": ["_", "$"]}],
```
yaroslavgribov: Adds _ and $ as accepted short variable names.

js/components/accessibilityControls.js
```diff
@@ -37,38 +37,38 @@ AccessibilityControls.prototype = {
     var targetTagName = this.getTargetTagName(e);
     var charCode = e.which || e.keyCode;
     if (this.controller.videoVr) {
-      this.moveVrToDirection(e, charCode, true, targetTagName); //start rotate 360
+      this.moveVrToDirection(e, charCode, true, targetTagName); // start rotate 360
     }
 
     switch (charCode) {
-      case CONSTANTS.KEYCODES.SPACE_KEY:
+    case CONSTANTS.KEYCODES.SPACE_KEY:
```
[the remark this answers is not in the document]
pilievOoyala: I agree.  @yaroslavgribov can you add this rule:

 indent: ["error", 2, { "SwitchCase": 1 }]

I got it from here https://eslint.org/docs/rules/indent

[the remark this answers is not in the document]
yaroslavgribov: Good point, thanks!
I think this is the default behaviour for `"indent": ["error", 2]` that we agreed on, but it is also extensible https://eslint.org/docs/rules/indent#options , so we could have 
> Indent of 2 spaces with SwitchCase set to 1 will indent case clauses with 2 spaces with respect to switch statements. 
using `"indent": ["error", 2, { "SwitchCase": 1 }]`

js/components/closed-caption/captionOpacityTab.js
```diff
@@ -1,9 +1,9 @@
 var React = require('react'),
-    ClassNames = require('classnames'),
-    Utils = require('../utils'),
-    CONSTANTS = require('../../constants/constants'),
-    SelectionContainer = require('./selectionContainer'),
-    Slider = require('../slider');
+  ClassNames = require('classnames'),
```
yaroslavgribov: I think this is extensible as well with `indent: ["error", 2, { "VariableDeclarator": { "var": 2 } } ]`, should we add this change?

aeng7: Yep, I think this will help readability as well

js/components/viewControlsVr.js
```diff
@@ -72,13 +72,13 @@ var ViewControlsVr = React.createClass({
...
 
-  render: function () {
+  render: function() {
     var isShowing = !!(this.icon && this.icon.name);
 
     return !isShowing ? null :
-      (<div className={classnames("oo-vr-icon-container view-controls", {"oo-vr-icon-container--hidden": !this.props.controlBarVisible})}>
-        <Icon {...this.props} icon={this.backgroundIcon} className={classnames("oo-vr-icon--substrate")}/>
-        <Icon {...this.props} icon={this.icon.name} className={classnames("oo-vr-icon--icon-symbol")}/>
+      (<div className={classnames('oo-vr-icon-container view-controls', {'oo-vr-icon-container--hidden': !this.props.controlBarVisible})}>
```
yaroslavgribov: It feels like we had more double quotes than single quotes, should we stick with doubles?

yaroslavgribov: image
image

yaroslavgribov: image
image

aeng7: That's barely more. I feel like we should use whatever looks more pleasant/convenient. I'm fine with either one.

js/components/adOverlay.js
```diff
@@ -1,10 +1,10 @@
-/********************************************************************
+/** ******************************************************************
   AD OVERLAY
 *********************************************************************/
 var React = require('react'),
-    ClassNames = require('classnames'),
-    CloseButton = require('./closeButton'),
-    CONSTANTS = require('../constants/constants');
+  ClassNames = require('classnames'),
```
pilievOoyala: is there rule for enforcing that you must use 'var' for new variables? @aeng7  what do you think?

yaroslavgribov: @pilievOoyala  yeah there's https://eslint.org/docs/rules/one-var if I understand you correctly

js/components/viewControlsVr.js
pilievOoyala: Let's stick with single. I think we have been using those more recently.

js/components/adPanel.js
```diff
@@ -8,11 +8,11 @@
  * @constructor
  */
 var React = require('react'),
-  CONSTANTS = require('../constants/constants'),
-  Spinner = require('./spinner'),
-  ClassNames = require('classnames'),
-  Utils = require('./utils'),
-  Icon = require('../components/icon');
+    CONSTANTS = require('../constants/constants'),
```
pilievOoyala: I don't know if my other comment got erased.  

@aeng7  What do you think if we enforce having to write 'var' for all new variable declarations instead of the this comma separated style?

It would use this rule I think: 
one-var: ["error", "never"]
From: https://eslint.org/docs/rules/one-var

yaroslavgribov: I personally like to split the declarations and it could be better to split them if we plan on moving to const.

yaroslavgribov: Although it seems it can't be autofixed https://github.com/eslint/eslint/issues/9072

pilievOoyala: Let's not put that in and we'll just update the standards for now.  Maybe one of us can write a script to fix it up.

pilievOoyala: Actually, can you put that rule in as a warning?

pilievOoyala:
You can merge it after that last rule update.  Thanks again @yaroslavgribov

yaroslavgribov:
Full ESLint run on `js/` 
image
`tests/` 
image

PR #1051, opened 2018-04-11
Bugfix/player 3289  Language menu is closed by clicking on Language icon after switching to window mode 

KseniiaSokolovaEpam:
Bug fix for
[https://jira.corp.ooyala.com:8443/browse/PLAYER-3295] and
[https://jira.corp.ooyala.com/browse/PLAYER-3289]
When we click on the fullscreen icon in fullscreen mode popovers are closing;
When we change a language in a multiaudio popover (large screen), the popover is closing;
It was checked it in the test.

\+ path to const was fixed (for quality popover)
\+ some changes for eslint


js/skin.js
```diff
@@ -155,6 +160,7 @@ var Skin = React.createClass({
   /**
    * @public
    * @description set isVrMouseMove to false for click event
+   * @return {undefined}
```
pilievOoyala: We should take this out since it doesn't return anything.

js/skin.js
```diff
@@ -165,17 +171,19 @@ var Skin = React.createClass({
   /**
    * @public
    * @description set accessibilityControlsEnabled to true for focus event
+   * @return {undefined}
```
pilievOoyala: We should take this out since it doesn't return anything.

pilievOoyala:
LGTM. Except for the two minor jsdoc things.

js/skin.js
```diff
@@ -122,6 +125,7 @@ var Skin = React.createClass({
   /**
    * @public
    * @description the function is called when we stop the rotation
+   * @return {undefined}
```
aeng7: Another undefined jsdoc return

aeng7:
LGTM other than Peter's comments

js/skin.js

js/skin.js

js/skin.js

PR #1052, opened 2018-04-11
Bugfix/player 3308

yaroslavgribov:
Fixes issue where user selected default track was set after SET_CURRENT_AUDIO rather than actual user input.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -898,7 +898,8 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
      */
     setCurrentAudio: function(currentTrack) {
       // user selected track should be saved to localstorage
-      OO.setItem(OO.CONSTANTS.SELECTED_AUDIO, currentTrack);
+      var currentAudioTrack = JSON.stringify(currentTrack);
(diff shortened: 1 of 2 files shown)


pilievOoyala:
LGTM

js/controller.js
```diff
@@ -897,6 +897,10 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
      * @param currentTrack {{id: String, label: String, lang: String}} - current active audio track
      */
     setCurrentAudio: function(currentTrack) {
+      // user selected track should be saved to localstorage
+      var currentAudioTrack = JSON.stringify(currentTrack);
+      OO.setItem(OO.CONSTANTS.SELECTED_AUDIO, currentAudioTrack);
```
aeng7: Where is the stored value used? Just out of curiosity

aeng7:
LGTM other than a small comment

js/controller.js
yaroslavgribov: @aeng7 Thanks for the review. 
We use it later in the core to restore the default track based on user previous selection.

PR #1056, opened 2018-04-13
Bugfix/player-3314 to master Language selection closes the CC window

KseniiaSokolovaEpam:
Fix the issue that  window CC is closed after change a language.

CLOSED_CAPTION_SCREEN is used instead of CLOSEDCAPTION_SCREEN in the skin plugin.

No tests are needed.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -639,7 +639,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
         } else if (this.state.playerState === CONSTANTS.STATE.PLAYING) {
           this.state.screenToShow = CONSTANTS.SCREEN.PLAYING_SCREEN;
         } else if (this.state.playerState === CONSTANTS.STATE.PAUSE &&
-          this.state.screenToShow !== CONSTANTS.SCREEN.CLOSEDCAPTION_SCREEN) {
+          this.state.screenToShow !== CONSTANTS.SCREEN.CLOSED_CAPTION_SCREEN) {
           this.state.screenToShow = CONSTANTS.SCREEN.PAUSE_SCREEN;
         }
       }
--- a/js/controller.js
+++ b/js/controller.js
@@ -1384,7 +1384,6 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       } else {
         this.state.closedCaptionOptions.cueText = null;
       }
-      this.renderSkin();
     },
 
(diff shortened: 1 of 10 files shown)


js/controller.js
```diff
@@ -1388,7 +1388,6 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       } else {
         this.state.closedCaptionOptions.cueText = null;
       }
-      this.renderSkin();
```
jpfranco: This change worries me a bit. This event is necessary in order to display closed captions and it looks like this change would prevent captions from being displayed in some cases. Consider that there are other events calling `this.renderSkin()`, which might give the impression that this particular call to `this.renderSkin()` is not needed. However, if this event fires and no other action causes the skin to be rendered, then it looks like the closed captions would not be displayed. Have we tested whether closed captions are affected by this change?

I also don't fully understand why calling `this.renderSkin()` inside `onClosedCaptionsInfoAvailable` affects the closed captions menu. It seems like we might not be addressing the root cause here. Can you explain why this change fixes the issue?

jpfranco:
Just to clarify a bit. My suggestion is to look for an alternative solution that addresses the root cause. This workaround seems risky to me, but if there's a good reason why we're doing this then we can discuss it. Please let us know.

KseniiaSokolovaEpam:
@jpfranco thank you! I think that you are right, so I found new way to fix this issue.

jpfranco:
LGTM, need a second review @pilievOoyala @aeng7 @dallen42

js/controller.js
```diff
@@ -638,7 +638,8 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
           }
         } else if (this.state.playerState === CONSTANTS.STATE.PLAYING) {
           this.state.screenToShow = CONSTANTS.SCREEN.PLAYING_SCREEN;
-        } else if (this.state.playerState === CONSTANTS.STATE.PAUSE) {
+        } else if (this.state.playerState === CONSTANTS.STATE.PAUSE &&
+          this.state.screenToShow !== CONSTANTS.SCREEN.CLOSEDCAPTION_SCREEN) {
```
aeng7: Should we use CLOSED_CAPTION_SCREEN instead? This will make it consistent with other CLOSED_CAPTION constants.

pilievOoyala:
LGTM.

PR #1058, opened 2018-04-13
Bugfix/player-3310 to master git commit -am Issue with playlist when switching from video with multi audio to one with only a single track

KseniiaSokolovaEpam:
Remove multiaudio from control Bar if multiAudio is not available.

[Original task|https://jira.corp.ooyala.com/browse/PLAYER-3310]

Unit test is added, tests are passed.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -197,6 +197,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       this.mb.subscribe(OO.EVENTS.RECREATING_UI, 'customerUi', _.bind(this.recreatingUI, this));
       this.mb.subscribe(OO.EVENTS.MULTI_AUDIO_FETCHED, 'customerUi', _.bind(this.onMultiAudioFetched, this));
       this.mb.subscribe(OO.EVENTS.MULTI_AUDIO_CHANGED, 'customerUi', _.bind(this.onMultiAudioChanged, this));
+      this.mb.subscribe(OO.EVENTS.REMOVE_MULTI_AUDIO, 'customerUi', _.bind(this.onRemoveMultiAudio, this));
       this.mb.subscribe(OO.EVENTS.ERROR, 'customerUi', _.bind(this.onErrorEvent, this));
       this.mb.addDependent(OO.EVENTS.PLAYBACK_READY, OO.EVENTS.UI_READY);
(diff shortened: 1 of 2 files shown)


jpfranco:
The approach that we use for Closed Captions and other asset-specific data is to clear the previous state on `EMBED_CODE_CHANGED`. I think that the video plugins shouldn't have to let the player know that they don't own the data from the previous asset, the new asset should just have a clean state by default.

I think that the the simplest way to fix this issue is to just clear the multiaudio state on the `EMBED_CODE_CHANGED` handler, that way we only have to modify a single project. Please let me know if this doesn't work.

KseniiaSokolovaEpam:
@jpfranco 
I am sorry, but I did not get your point. We check current state, not previous.

jpfranco:
@KseniiaSokolovaEpam what I'm trying to say is that we already have a mechanism for clearing the state of the previous video when a new video is selected, so it is not necessary to introduce a new event in order to clear multi-audio data.

The `EMBED_CODE_CHANGED` event is fired whenever a new video is set. If you take a look at the `onEmbedCodeChanged` handler in the skin, you'll notice that we use this event in order to remove all of the state from the previous video. So all that is needed in order to fix the issue is something like this:

```javascript
    onEmbedCodeChanged: function(event, embedCode, options) {
      this.state.videoQualityOptions.availableBitrates = null;
      this.state.videoQualityOptions.selectedBitrate = null;
      this.state.closedCaptionOptions.availableLanguages = null;
      this.state.closedCaptionOptions.cueText = null;
      this.state.closedCaptionsInfoCache = {};
      // Clear previous multi audio state. This is a new video so we shouldn't preserve 
      // any state from the previous one
      this.state.multiAudio = null;
      ...
```
The main reason why I'm suggesting this approach is that I think we should keep this type of behavior consistent between features. This pattern already works well for closed captions, so I think we should use it for multi-audio as well. Also note that with this approach we only modify a single project instead of four.

If there's another reason why you think this wouldn't work for this particular case then please let me know, but otherwise I think we should use this approach.

aeng7:
I agree with @jpfranco. We should start clean with a new video and update when notified to do so.

KseniiaSokolovaEpam:
@jpfranco Oh, great idea! Thank you!

PR #1059, opened 2018-04-17
Feature/player 3310 Issue with playlist when switching from video with multi audio to one with only a single track

KseniiaSokolovaEpam:
Set state multiAudio to null when embed code was changed.

Origin task is https://jira.corp.ooyala.com/browse/PLAYER-3310

Unit tests are passed.

### DIFF
--- a/tests/html5skin-test.js
+++ b/tests/html5skin-test.js
@@ -352,7 +352,6 @@ describe('Controller', function() {
       controller.onSetEmbedCode('newEmbedCode');
       controller.onEmbedCodeChanged('newEmbedCode');
       expect(controller.skin.state.currentPlayhead).toBe(0);
-      expect(controller.state.multiAudio).toBe(null);
     });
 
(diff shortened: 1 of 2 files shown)


tests/html5skin-test.js
```diff
@@ -352,6 +352,7 @@ describe('Controller', function() {
       controller.onSetEmbedCode('newEmbedCode');
       controller.onEmbedCodeChanged('newEmbedCode');
       expect(controller.skin.state.currentPlayhead).toBe(0);
+      expect(controller.state.multiAudio).toBe(null);
```
jpfranco: Can we move this to a separate test in the multi-audio section?  It looks like `multiAudio` state is null by default, so this wouldn't check that the state is being cleared. The test would need to set a multi audio state before calling `onEmbedCodeChanged`. The changes look good otherwise, thanks for the update.

PR #1060, opened 2018-04-17
Bugfix/player 3333 Pausing of live playback while changing language in minimized screen

KseniiaSokolovaEpam:
We do not pause a video after opening multiAudio screen.

Task is https://jira.corp.ooyala.com/browse/PLAYER-3333

Unit tests are passed.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -897,6 +897,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
      * @param currentTrack {{id: String, label: String, lang: String}} - current active audio track
      */
     setCurrentAudio: function(currentTrack) {
+      console.log('BBB setCurrentAudio start');
       // user selected track should be saved to localstorage
       var currentAudioTrack = JSON.stringify(currentTrack);
       OO.setItem(OO.CONSTANTS.SELECTED_AUDIO, currentAudioTrack);
@@ -1788,6 +1789,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
      * The function handles a click on multiAudio icon
      */
     toggleMultiAudioScreen: function() {
+      console.log('BBB toggleMultiAudioScreen start');
       if (this.state.screenToShow === CONSTANTS.SCREEN.MULTI_AUDIO_SCREEN) {
         if (this.state.playerState === CONSTANTS.STATE.END) {
           this.state.screenToShow = CONSTANTS.SCREEN.END_SCREEN;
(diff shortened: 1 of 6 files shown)


js/controller.js
```diff
@@ -1788,6 +1785,24 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       }
     },
 
+    /**
+     * The function handles a click on multiAudio icon
+     */
+    toggleMultiAudioScreen: function() {
```
jpfranco: Would it be possible to add this behavior to the `toggleScreen` instead of having a separate function for multi-audio? If doing this doesn't cause issues then I think it would be better to keep this logic generic, that we we don't have two different code paths to maintain.

tests/html5skin-test.js
```diff
@@ -352,6 +352,8 @@ describe('Controller', function() {
       controller.onSetEmbedCode('newEmbedCode');
       controller.onEmbedCodeChanged('newEmbedCode');
       expect(controller.skin.state.currentPlayhead).toBe(0);
+      controller.toggleMultiAudioScreen();
```
jpfranco: This test checks whether the playhead is reset on embed code change, so I don't  think this is the right place for these checks. Can we create a separate test in the multi audio section and check that toggling the screen on and off works correctly?

js/controller.js
KseniiaSokolovaEpam: Hello @jpfranco !
Thanks for the comment!
Yes, it is possible, but behavoir of this functions is different, so in this case we will have something like this:
```
    toggleScreen: function(screen) {
      this.isNewVrVideo = false;
      if (screen === CONSTANTS.SCREEN.MULTI_AUDIO_SCREEN) {
        this.toggleMultiAudioScreen();
      } else {
        if (this.state.screenToShow === screen) {
          this.closeScreen();
        }
        else {
          if (this.state.playerState === CONSTANTS.STATE.PLAYING) {
            this.pausedCallback = function() {
              this.state.pluginsElement.addClass('oo-overlay-blur');
              this.state.screenToShow = screen;
              this.renderSkin();
            }.bind(this);
          }
          else {
            this.state.screenToShow = screen;
            this.state.pluginsElement.addClass('oo-overlay-blur');
            this.renderSkin();
          }
        }
      }
    }
```
It is not beautiful, is not it?

tests/html5skin-test.js

js/controller.js
jpfranco: @KseniiaSokolovaEpam what I had in mind was to refactor the `toggleScreen` method in order to include the use case for multi audio, rather than call a separate `toggleMultiAudioScreen` function from within `toggleScreen`.

This is not that important though, so I'm ok with this as long as the second reviewer agrees.

PR #1063, opened 2018-04-20
[bugfix/PLAYER-3386] No longer able to drag on skin

KseniiaSokolovaEpam:
The PR fix the issue with video 360.

Original task is https://jira.corp.ooyala.com/browse/PLAYER-3386

No tests are needed.

### DIFF
--- a/js/skin.js
+++ b/js/skin.js
@@ -108,7 +108,7 @@ var Skin = React.createClass({
       return;
     }
     if (this.props.controller && this.props.controller.videoVr && this.state.isVrMouseDown) {
-      e.preventDefault();
+      event.preventDefault();
       this.setState({
         isVrMouseMove: true
       });


js/skin.js
```diff
@@ -108,7 +108,7 @@ var Skin = React.createClass({
       return;
     }
     if (this.props.controller && this.props.controller.videoVr && this.state.isVrMouseDown) {
-      e.preventDefault();
+      event.preventDefault();
```
jpfranco: I think this is a good opportunity to add a unit test for this method. It looks like we don't have good coverage for the VR360 code in skin.js

KseniiaSokolovaEpam: This function is tested already. I have chenged name for param and forgot chenge the param in the place.

aldigun: Just to confirm, is there a unit test for this?

jpfranco: @KseniiaSokolovaEpam I'm not seeing unit tests for the `handleVrPlayerMouseDown` function from `skin.js`. We try to always add a unit test on bug fixes since a bug indicates that there's an error condition which could've possibly been avoided with unit tests.

KseniiaSokolovaEpam: @jpfranco ok, thanks for the comment. I added the test for the bug.

jpfranco: @KseniiaSokolovaEpam thanks for the update.

jpfranco:
@pilievOoyala @aeng7 @dallen42 need a second review.

aeng7:
LGTM

PR #1070, opened 2018-05-10
[bugfix/PLAYER-3296] Add localized headers to CC and MA menus

yaroslavgribov:
Fixes an issue where title from constants was used instead of retrieving localized string.

### Implements
- [PLAYER-3296](https://jira.corp.ooyala.com/browse/PLAYER-3296)

### Tests
- Added a test to verify that string from translations is used in rendering


pilievOoyala:
LGTM.  @aeng7 @jpfranco  Can you review too?

PR #1072, opened 2018-05-14
[feature/PLAYER-3482] The menu should auto-focus on the first element, but only when triggered via keyboard

KseniiaSokolovaEpam:
When multiAudio menu is opened by keyboard first element from the list should be focused.
Original task is https://jira.corp.ooyala.com/browse/PLAYER-3482
Unit tests are passed.

### DIFF
--- a/js/components/closed-caption-multi-audio-menu/tab.js
+++ b/js/components/closed-caption-multi-audio-menu/tab.js
@@ -2,6 +2,7 @@ var React = require('react');
 var ScrollArea = require('react-scrollbar/dist/no-css');
 var Icon = require('../icon');
 var AccessibleButton = require('../accessibleButton');
+var AccessibleMenu = require('../higher-order/accessibleMenu');
 var classnames = require('classnames');
 var CONSTANTS = require('../../constants/constants');
 
@@ -21,7 +22,10 @@ var Tab = React.createClass({
           speed={1}
           horizontal={false}
         >
-          <ul className="oo-cc-ma-menu__list">
+          <ul
+            className="oo-cc-ma-menu__list"
+            role={CONSTANTS.ARIA_ROLES.MENU}
+          >
             {this.props.itemsList.map(function(el, index) {
               return (
                 <li
@@ -57,6 +61,8 @@ var Tab = React.createClass({
     );
   }
 });
+
+Tab = AccessibleMenu(Tab, { useRovingTabindex: true });
 
 Tab.defaultProps = {
   skinConfig: {
--- a/js/components/closed-caption-multi-audio-menu/tab.js
+++ b/js/components/closed-caption-multi-audio-menu/tab.js
@@ -1,7 +1,9 @@
 var React = require('react');
 var ScrollArea = require('react-scrollbar/dist/no-css');
 var Icon = require('../icon');
+var AccessibleButton = require('../accessibleButton');
 var classnames = require('classnames');
+var CONSTANTS = require('../../constants/constants');
 
 var Tab = React.createClass({
   handleClick: function(id) {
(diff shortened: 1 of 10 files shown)


js/components/closed-caption-multi-audio-menu/tab.js
```diff
@@ -19,24 +22,36 @@ var Tab = React.createClass({
...
+                    ariaLabel={el.label}
+                    ariaChecked={el.enabled}
+                    onClick={this.handleClick.bind(this, el.id)}
+                  >
+                    <Icon
+                      skinConfig={this.props.skinConfig}
+                      icon="selected"
+                      className={classnames({ 'oo-icon-hidden': !el.enabled })}
+                    />
+                    <span className="oo-cc-ma-menu__name" title={el.label}>
```
jpfranco: Do we need the title attribute?

js/components/closed-caption-multi-audio-menu/tab.js
```diff
@@ -19,24 +22,36 @@ var Tab = React.createClass({
...
                     {el.label}
                   </span>
```
jpfranco: Can we fix the indentation here?

js/components/closed-caption-multi-audio-menu/tab.js
```diff
@@ -19,24 +22,36 @@ var Tab = React.createClass({
           speed={1}
           horizontal={false}
         >
-          <ul className="oo-cc-ma-menu__list">
+          <ul
+            className="oo-cc-ma-menu__list"
+            role={CONSTANTS.ARIA_ROLES.MENU}
+          >
             {this.props.itemsList.map(function(el, index) {
```
jpfranco: Can we use a more descriptive name for `el`? It's not clear whether this is a DOM element or an audio track, etc.

js/components/closed-caption-multi-audio-menu/tab.js
```diff
@@ -19,24 +22,36 @@ var Tab = React.createClass({
...
               return (
                 <li
                   key={index}
```
jpfranco: Is it possible to use `el.id` instead of `index` here? Using indexes as keys is not recommended: https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

js/components/closed-caption-multi-audio-menu/tab.js
```diff
@@ -19,24 +22,36 @@ var Tab = React.createClass({
...
-                  onClick={this.handleClick.bind(this, el.id)}
                   className={classnames('oo-cc-ma-menu__element', {
```
jpfranco: I recommend adding a `role="presentation"` attribute to the `li` element, otherwise NVDA will not detect the correct number of items in this menu.

scss/components/_cc-ma-menu.scss
```diff
@@ -100,4 +101,13 @@
     right: 14px;
     font-size: 16px;
   }
+}
+.oo-ma-btn {
```
aeng7: Does ma stand for multi audio? Can we write out 'multi-audio' for this class? It would be more descriptive and have less chance at possible conflicts/confusion with new classes in the future.

js/components/closed-caption-multi-audio-menu/tab.js
KseniiaSokolovaEpam: \+

js/components/closed-caption-multi-audio-menu/tab.js
KseniiaSokolovaEpam: I changed it to 'item' and really it can be everything (and audio, and CC, and something also what we pass to the component as parameter for the list).

js/components/closed-caption-multi-audio-menu/tab.js
KseniiaSokolovaEpam: The attribute is used in styles to prevent jitter of lines when hovering (according to the design, the font becomes bold when hovering).

js/components/closed-caption-multi-audio-menu/tab.js
KseniiaSokolovaEpam: \+

KseniiaSokolovaEpam: Very interesting! Thank you!

js/components/closed-caption-multi-audio-menu/tab.js
KseniiaSokolovaEpam: \+

scss/components/_cc-ma-menu.scss
KseniiaSokolovaEpam: \+

pilievOoyala:
LGTM

PR #1074, opened 2018-05-16
Player 2103

aeng7:
-changed how we hide and show the ad plugins div to support non-blocking ads


js/controller.js
```diff
@@ -1217,6 +1217,8 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
         height: '',
         width: ''
       });
+      this.cancelTimer();
+      this.hideControlBar();
```
pilievOoyala: You don't have to fix this, but does it work with freewheel? Which ad manager did you test?

pilievOoyala:
LGTM

js/controller.js
aeng7: Tested with IMA. We'll need to test this with the other ad managers. This is just a user enhancement change in case an ad appears while the control bar is up. We could remove this if necessary or causes a lot of regressions.

dallen42:
LGTM

PR #1075, opened 2018-05-17
[PLAYER-3418] Expected UI Controls during SSAI Ad Playback

ipadilla4:


cacel:
@ipadilla4 Could you please resolve the conflicts in the PR? Thanks

js/components/adPanel.js
```diff
@@ -14,6 +14,13 @@ var React = require('react'),
     Utils = require('./utils'),
     Icon = require('../components/icon');
 
+var state = {
```
dallen42: @ipadilla4 the convention here seems to be to use this.setState(), see for example in discoveryPanel.js. @aeng7 can you confirm that is correct

dallen42:
@ipadilla4 can you add some unit tests for the new SSAI specific changes in adPanel.js?

js/components/adPanel.js
aeng7: Agreed, we should use React's component state whenever possible.

aeng7: Note that some of these can potentially be props.

dallen42: @ipadilla4 can you investigate converting some of these into props to address the infinite loop issue you mentioned earlier?

ipadilla4: Ok I used props before but I was having issues with object persistence, maybe I was using the wrong "upper" component. I will try this approach.

js/components/adPanel.js
```diff
@@ -90,13 +103,30 @@ var AdPanel = React.createClass({
...
 
     if (this.props.skinConfig.adScreen.showAdCountDown) {
-      var remainingTime;
+      var remainingTime = 0;
       if (isLive) {
-        remainingTime = parseInt(
-          (this.props.adStartTime + this.props.adVideoDuration * 1000 - new Date().getTime()) / 1000
-        );
+        remainingTime = parseInt((this.props.adStartTime + this.props.adVideoDuration * 1000 - new Date().getTime())/1000);
+        if (isSSAI) {
```
dallen42: @ipadilla4 is this behavior actually correct? It seems like you are trying to use this state to keep track of the amount of time the ad was paused and add it to the timer, is that correct? If so the result would be the timer is always increasing when the ad is paused... is that the expected user experience? Is that what non-ssai ads do currently?

ipadilla4: @dallen42 for live non-ssai ads we use current time (new Date().getTime()) to calculate remaining time. Since now we can pause our live ads, we need a way to calculate the total time it was paused so we can add it to remaining time, so when user resume the counter keeps decreasing from the time was paused.

ipadilla4: @dallen42 or maybe we can preserve the time when ad was paused and then decrease that number every second when user resumes the ad. I can try this approach with props and see how it behaves.

dallen42:
@ipadilla4 can we add a few unit tests to test the live countdown pause functionality?

ipadilla4:
This one is ready to be reviewed. Please take a look at it.

js/controller.js
```diff
@@ -806,6 +813,13 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
 
     onVcPlay: function(event, source) {
       this.state.currentVideoId = source;
+      if (this.state.wasPaused){
```
dallen42: @ipadilla4 should have space between `)` and `{`

js/controller.js
```diff
@@ -822,7 +836,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
         this.state.isInitialPlay = false;
         this.renderSkin();
       }
-      if (source === OO.VIDEO.ADS) {
+      if (source === OO.VIDEO.ADS || this.state.isPlayingAd){
```
dallen42: @ipadilla4 also space between `)` and `{` here

js/controller.js
```diff
@@ -1390,6 +1408,27 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       this.renderSkin();
     },
 
+    getAdRemainingTime: function(){
```
dallen42: @ipadilla4 space between `)` and `{` here

tests/components/adPanel-test.js
```diff
@@ -13,6 +13,9 @@ describe('AdPanel', function() {
     var mockController = {
       state: {
         isMobile: false
+      },
+      getAdRemainingTime: function(){
```
dallen42: @ipadilla4 space before `{`

tests/components/adPanel-test.js
```diff
@@ -56,6 +59,9 @@ describe('AdPanel', function() {
       onAdsClicked: function(source) {
         learnMoreClicked = true;
         clickSource = source;
+      },
+      getAdRemainingTime: function(){
```
dallen42: @ipadilla4 space before `{`

js/controller.js
```diff
@@ -806,6 +813,13 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
 
     onVcPlay: function(event, source) {
       this.state.currentVideoId = source;
+      if (this.state.wasPaused){
+        this.state.adPauseDuration = Date.now() - this.state.adPausedTime;
+        //we calculate new ad end time, based on the time that ad was paused.
```
dallen42: @ipadilla4 minor grammar issue, should be: "time that *the* ad was paused"

js/controller.js
```diff
@@ -806,6 +813,13 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
 
     onVcPlay: function(event, source) {
       this.state.currentVideoId = source;
+      if (this.state.wasPaused) {
```
aeng7: Should we also check that the previous video was an ad before proceeding inside this if statement?

js/controller.js
```diff
@@ -1390,6 +1408,27 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
     },
 
+    getAdRemainingTime: function() {
+      var remainingTime = 0;
+
+      var isLive = (this.state.currentAdsInfo.currentAdItem) ? this.state.currentAdsInfo.currentAdItem.isLive : false;
+      var isSSAI = (this.state.currentAdsInfo.currentAdItem) ? this.state.currentAdsInfo.currentAdItem.ssai : false;
+
+      if (isLive) {
+        remainingTime = parseInt((this.state.adStartTime + this.state.adVideoDuration * 1000 - Date.now()) / 1000);
```
aeng7: Is the adStartTime in ms but the adVideoDuration in seconds? Just double checking

tests/html5skin-test.js
```diff
@@ -703,6 +703,36 @@ describe('Controller', function() {
       controller.onShowAdControls(event, false, false);
       expect(controller.state.forceControlBarVisible).toBe(false);
     });
+
```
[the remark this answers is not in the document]
ipadilla4: Done

js/controller.js
ipadilla4: Agree, this `wasPaused` flag is set only when an ad was paused, but I will ad the validation to ensure it was an ssai ad.

ipadilla4: I will also change the name to `adWasPaused`, to be more clear.

js/controller.js
ipadilla4: Changed adVideoDuration to ms.

dallen42:
LGTM

js/controller.js
```diff
@@ -806,6 +813,13 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
 
     onVcPlay: function(event, source) {
       this.state.currentVideoId = source;
+      if (this.state.adWasPaused && this.state.currentAdsInfo.currentAdItem.ssai) {
```
dallen42: @ipadilla4 do we need a null check here for `this.state.currentAdsInfo` and `this.state.currentAdsInfo.currentAdItem`? If not disregard

js/controller.js
```diff
@@ -893,12 +907,16 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
         this.state.playerState = CONSTANTS.STATE.PAUSE;
         this.renderSkin();
-      } else if (videoId === OO.VIDEO.ADS) {
+      } else if (videoId === OO.VIDEO.ADS || this.state.isPlayingAd) {
         // If we pause during an ad (such as for clickthroughs or when autoplay fails)
         // we'll show the control bar so that the user has an indication that the video
         // must be unpaused to resume
         this.state.config.adScreen.showControlBar = true;
         this.state.adPauseAnimationDisabled = false;
+        if (this.state.currentAdsInfo.currentAdItem && this.state.currentAdsInfo.currentAdItem.ssai) {
```
dallen42: @ipadilla4 also not sure we need a null check for `this.state.currentAdsInfo` and `this.state.currentAdsInfo.currentAdItem` here

js/controller.js
```diff
@@ -1390,6 +1410,27 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       this.renderSkin();
     },
 
+    getAdRemainingTime: function() {
```
aeng7: Can we add a jsdoc for this function? I understand some other functions don't have jsdocs, but we'd like to stay on top of this moving forward.

aeng7:
LGTM other than a minor comment about a jsdoc. I'll merge now, but please add the jsdoc in a future PR @ipadilla4 . Thanks!

aeng7:
LGTM. Thanks @ipadilla4 !

PR #1076, opened 2018-05-21
Feature/player 3533 Change the logic of showing multi-audio icon

KseniiaSokolovaEpam:
The name for the flag **showMultiAudioIcon** was changed to **hideMultiAudioIcon**.
By default multi-audio icon will be shown automatically if there are multiple audio tracks. 

- hideMultiAudioIcon = true - will hide MA icon even if there are several audio tracks in the video
- hideMultiAudioIcon = false - will show MA icon if there are several audio tracks 
- If flag hideMultiAudioIcon is not set its value devaults to hideMultiAudioIcon = false

Unit tests are passed.


aeng7:
LGTM

pilievOoyala:
LGTM

PR #1077, opened 2018-05-21
Player 3498

pilievOoyala:


js/controller.js
```diff
@@ -1266,15 +1265,20 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
     },
 
-    onShowAdControls: function(event, showAdControls) {
+    onShowAdControls: function(event, showAdControls, autoHide) {
       this.state.showAdControls = showAdControls;
+      this.state.forceControlBarVisible = false;
       if (showAdControls && this.state.config.adScreen.showControlBar) {
         this.state.pluginsElement.removeClass('oo-full');
         this.state.pluginsClickElement.removeClass('oo-full');
+        if (typeof autoHide !== 'undefined') {
```
aldigun: something bugs me about this condition. autoHide can be an object, number, etc and it will be a valid condition. Is autoHide suppose to be a boolean?

pilievOoyala: I think the !autohide will covert it to a boolean in the next line. But I can change it to '=== boolean' instead.

aldigun: Okay, other than this, LGTM. I will give a pre-approval.

aeng7:
LGTM

PR #1078, opened 2018-05-22
[player-3531]

aeng7:
-fixed an issue where the plugins div could support variable heights for ads


aeng7:
We don't have any css unit tests at the moment, nor an easy way to set them up.

pilievOoyala:
woot! easy fix.  LGTM

PR #1079, opened 2018-06-01
[bugfix/PLAYER-3580] Audio instead of MultiAudio as a text for tooltip

KseniiaSokolovaEpam:
Use translated "Audio" instead of untranslated "Multi Audio"

[Original task])(https://jira.corp.ooyala.com/browse/PLAYER-3580)

No tests needed


aeng7:
LGTM

PR #1080, opened 2018-06-05
[PLAYER-3645] Control bar refactor/cleanup

jpfranco:
This update is meant to simplify button declarations inside the control bar. So far I mostly focused on removing the boilerplate code related to tooltips, but the end goal is to eventually refactor the control bar buttons into a separate component that just takes a few props and renders the icon, tooltip, etc.

No unit tests are needed since nothing new was added.


pilievOoyala:
LGTM. Let's merge this after the build today.

js/constants/constants.js
```diff
@@ -155,6 +155,23 @@ module.exports = {
...
+  CONTROL_BAR_KEYS: {
+    PLAY_PAUSE: 'playPause',
+    LIVE: 'live',
+    VOLUME: 'volume',
+    TIME_DURATION: 'timeDuration',
+    FLEXIBLE_SPACE: 'flexibleSpace',
+    MORE_OPTIONS: 'moreOptions',
+    QUALITY: 'quality',
+    DISCOVERY: 'discovery',
+    CLOSED_CAPTION: 'closedCaption',
```
aeng7: Should we rename this to closedCaptions to make it consistent with other closed captions constants? Though I already notice a few discrepancies like discover vs discovery

js/components/controlBar.js
```diff
@@ -858,149 +782,103 @@ var ControlBar = React.createClass({
...
+            icon="share"
+            style={dynamicStyles.iconCharacter}
             onMouseOver={this.highlight}
-            onMouseOut={this.removeHighlight}
-            key="stereo"
-            focusId={CONSTANTS.FOCUS_IDS.STEREO}
-            ariaLabel={stereoAriaLabel}
-          >
-            <Icon {...this.props} icon={stereoIcon} style={dynamicStyles.iconCharacter} />
-            <Tooltip
```
aeng7: What happened with the tooltip for stereoscopic? I don't see it in the new changes.

js/components/controlBar.js
```diff
@@ -550,284 +550,208 @@ var ControlBar = React.createClass({
...
             <Icon
               {...this.props}
-              icon="ellipsis"
+              icon="quality"
               style={dynamicStyles.iconCharacter}
               onMouseOver={this.highlight}
-              onMouseOut={this.removeHighlight}
-            />
+              onMouseOut={this.removeHighlight} />
             <Tooltip
```
aeng7: What happened with the tooltip for flexible space? I'm assuming it was removed because it wasn't needed?

aeng7:
LGTM other than a few comments. I'll approve once @jpfranco answers the comments.

js/constants/constants.js
jpfranco: Yes, I think `closedCaptions` is the correct name, but this matches the name from the buttons section of the skin.json. I also kept the value that was already being used in case it was being referenced elsewhere.

js/components/controlBar.js
jpfranco: I removed it because it had an empty tooltip with no text. I double-checked and the `Stereo` button only shows up on mobile (where tooltips are disabled), so that's probably why no one had noticed before.

js/constants/constants.js
jpfranco: @aeng7 actually, it looks like the multi audio and VR360 ones don't follow this rule, I'll fix them. But the idea is that they need to match the name from the skin.json

aeng7:
LGTM

pilievOoyala:
Still looks good

PR #1081, opened 2018-06-14
bugfix/PLAYER-3622 Focus is not restored to Multi-audio menu after clicking Close icon

KseniiaSokolovaEpam:
Focus is not saved on the btn when we use NVDA, so we set the focus forcibly.

[Original task](https://jira.corp.ooyala.com/browse/PLAYER-3622)

Unit tests are passed.

### DIFF
--- a/js/components/controlBar.js
+++ b/js/components/controlBar.js
@@ -859,9 +859,7 @@ var ControlBar = React.createClass({
               ariaLabel={CONSTANTS.ARIA_LABELS.MULTI_AUDIO}
               ariaHasPopup={true}
               ariaExpanded={this.props.controller.state.multiAudioOptions.showPopover ? true : null}
-              onClick={function(evt) {
-                this.handleMultiAudioClick(evt);
-              }.bind(this)}
+              onClick={this.handleMultiAudioClick}
             >
               <Icon
                 {...this.props}
--- a/js/components/controlBar.js
+++ b/js/components/controlBar.js
@@ -34,7 +34,6 @@ var ControlBar = React.createClass({
   },
 
   componentDidMount: function() {
-    console.log('BBB controlBar this.props.controller.toggleButtons', this.props.controller.toggleButtons);
     window.addEventListener('orientationchange', this.closeOtherPopovers);
     window.addEventListener('orientationchange', this.setLandscapeScreenOrientation, false);
     document.addEventListener('keydown', this.handleControlBarKeyDown);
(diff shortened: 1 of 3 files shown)


js/components/controlBar.js
```diff
@@ -858,7 +859,9 @@ var ControlBar = React.createClass({
               ariaLabel={CONSTANTS.ARIA_LABELS.MULTI_AUDIO}
               ariaHasPopup={true}
               ariaExpanded={this.props.controller.state.multiAudioOptions.showPopover ? true : null}
-              onClick={this.handleMultiAudioClick}
+              onClick={function(evt) {
```
aeng7: Is this change necessary? From the looks of the old code, it looks like the event argument should be passed to handleMultiAudioClick.

KseniiaSokolovaEpam: Oh, yes! You are right! Thank you.

js/components/controlBar.js
```diff
@@ -248,6 +248,7 @@ var ControlBar = React.createClass({
     } else {
       this.props.controller.toggleMultiAudioScreen();
     }
+    this.handleControlBarFocus(evt);
```
jpfranco: I'm remembering just now that we have a big refactor coming up where the `handleControlBarFocus` function will be removed from this component (you can see this update in the PLAYER-3475 branch).  Could we switch to the approach that we use for the closed captions menu? You can see how this was implemented in `closedCaptionsPanel.js`

tests/components/controlBar-test.js
```diff
@@ -1130,6 +1130,52 @@ describe('ControlBar', function() {
     expect(popoverStateChanged).toBe(true);
   });
 
+  it('should store multiAudio button focus state', function() {
+    var mockController = {
```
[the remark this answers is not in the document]
KseniiaSokolovaEpam: Was changed for the [PR](https://github.com/ooyala/html5-skin/pull/1082)

js/components/controlBar.js
KseniiaSokolovaEpam: @jpfranco oh,  i am sorry, I did not get your idea about the approach. But anyway we checked CC and the bug is existing and for CC now.

jpfranco: Oh ok, it seems to be working for me. Does that only happen when using NVDA?

The problem is that this code has changed a lot, so this focus handler is not even in this file anymore. Please take a look at this PR: https://github.com/ooyala/html5-skin/pull/1083. This logic has been moved to the `PreserveKeyboardFocus` component.

Is it possible to update your fix so that it doesn't rely on the `handleControlBarFocus` function?

aeng7: Note that 1083 has been merged

aeng7:
@KseniiaSokolovaEpam @jpfranco should this PR be updated due to the recent skin changes?

KseniiaSokolovaEpam:
@aeng7 Now function "handleControlBarFocus" does not exists. (I meant that not only last changes are blockers for merge). So I think that would be better to close  the PR and create new one.

PR #1082, opened 2018-06-18
[bugfix/PLAYER-3751] Cardboard icon is visible on tablets

KseniiaSokolovaEpam:
There is function "isMobile" which checkes if a user device is phones or tablets. And returns true if the device is phone or tablet. But we need to show a carbord icon only on phones (exept tablets).

So now we have function "getUserDevice" for checking type of the user devices ( which returns one of values: phone, tablet or desktop). And if the device is phone we show the icon.

Unit tests are passed.

And there are 2 changes for ESLint (in utils.js line 617 and controlBar.js line 638)

### DIFF
--- a/js/components/controlBar.js
+++ b/js/components/controlBar.js
@@ -19,6 +19,7 @@ var React = require('react'),
 var ControlBar = React.createClass({
   getInitialState: function() {
     this.isMobile = this.props.controller.state.isMobile;
+    this.isPhone = Utils.getUserDevice() === 'phone';
     this.domNode = null;
     this.responsiveUIMultiple = this.getResponsiveUIMultiple(this.props.responsiveView);
     this.moreOptionsItems = null;
@@ -106,7 +107,7 @@ var ControlBar = React.createClass({
     evt.preventDefault();
     if (this.props.controller) {
       this.props.controller.toggleFullscreen();
-      if (this.vr && this.isMobile && this.props.controller.isVrStereo) {
+      if (this.vr && this.isPhone && this.props.controller.isVrStereo) {
         this.toggleStereoVr();
       }
     }
@@ -634,7 +635,10 @@ var ControlBar = React.createClass({
       ),
 
       flexibleSpace: (
-        <div key={CONSTANTS.CONTROL_BAR_KEYS.FLEXIBLE_SPACE} className="oo-flexible-space oo-control-bar-flex-space" />
+        <div
+          key={CONSTANTS.CONTROL_BAR_KEYS.FLEXIBLE_SPACE}
+          className="oo-flexible-space oo-control-bar-flex-space"
+        />
       ),
 
       moreOptions: (
@@ -831,7 +835,7 @@ var ControlBar = React.createClass({
         </a>
       ),
 
-      stereoscopic: !!(this.vr && this.isMobile) && (
+      stereoscopic: !!(this.vr && this.isPhone) && (
         <AccessibleButton
           key={CONSTANTS.CONTROL_BAR_KEYS.STEREOSCOPIC}
           className="oo-video-type oo-control-bar-item oo-vr-stereo-button"
--- a/js/components/utils.js
+++ b/js/components/utils.js
@@ -376,6 +376,28 @@ var Utils = {
     return this.isAndroid() || this.isIos();
   },
 
+
+  /**
+   * Get type of user device.
+   *
+   * @returns {string} - name of the user device, may be one of the values 'desktop', 'phone' or 'tablet'.
+   */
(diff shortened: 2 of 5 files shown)


js/components/controlBar.js
```diff
@@ -106,7 +107,7 @@ var ControlBar = React.createClass({
     evt.preventDefault();
     if (this.props.controller) {
       this.props.controller.toggleFullscreen();
-      if (this.vr && this.isMobile && this.props.controller.isVrStereo) {
+      if (this.vr && this.isPhone && this.props.controller.isVrStereo) {
```
jpfranco: We already have a `Utils.isIPhone()` helper function. Could we use that here instead?

js/components/utils.js
```diff
@@ -376,6 +376,28 @@ var Utils = {
     return this.isAndroid() || this.isIos();
   },
 
+
+  /**
+   * Get type of user device.
+   *
+   * @returns {string} - name of the user device, may be one of the values 'desktop', 'phone' or 'tablet'.
+   */
+  getUserDevice: function() {
```
jpfranco: Would recommend using the existing `Utils.isIPhone()` function if possible.

tests/components/controlBar-test.js
```diff
@@ -129,136 +129,155 @@ describe('ControlBar', function() {
...
+  describe('Vr on phones', function(){
+    beforeEach(function() {
+      window.navigator.userAgent = 'phone';
+    });
+    afterEach(function() {
+      window.navigator.userAgent = 'desktop';
+    });
+
+    it('render one stereo button if content vr', function() {
+      var mockController = {
```
jpfranco: Can we use baseMockController and baseMockProps and only set the properties that are needed for the test? This will help eliminate boiler plate code and keep tests short. I recommend the same for the tests below.

js/components/controlBar.js
KseniiaSokolovaEpam: @jpfranco no, I told about any phones (not only iPhones.)

js/components/utils.js
KseniiaSokolovaEpam: oh, in this function we don't determine if a device is iPhone, but determine if the device is mobile phone or tablet, or desktop.

js/components/controlBar.js
jpfranco: Oh ok, sorry. I read that wrong.

aeng7:
LGTM other than @jpfranco's comments

tests/components/controlBar-test.js
KseniiaSokolovaEpam: \+

js/components/utils.js
KseniiaSokolovaEpam: @jpfranco ok, i got your point. I will change the code. Thanks!

KseniiaSokolovaEpam: @jpfranco I checked sizes for devices and I undestood that a smaller device may have bigger size in px than bigger one. http://screensiz.es/ . That means that a tablet may be checked as a phone and a phone as a tablet.

So, now I think that if we want to check if device is phone, it would be better to use userAgent (as we do in the function).
Or maybe you have any other ideas?

jpfranco: Oh ok I see. So in some cases the resolution is higher than the screen size. I think we can check the user agent then. Thanks for updating all the old unit tests by the way.

aeng7:
@KseniiaSokolovaEpam looks like there are some conflicts that need to be resolved. Please look into this, thanks!

KseniiaSokolovaEpam:
I see 2 "LGTM" and I resolved the conflicts, so I will merge the PR.

tests/components/controlBar-test.js
```diff
@@ -49,78 +47,39 @@ describe('ControlBar', function() {
     baseMockProps = {
       isLiveStream: false,
       controller: baseMockController,
-      skinConfig: JSON.parse(JSON.stringify(defaultSkinConfig))
+      skinConfig: defaultSkinConfig
```
jpfranco: @KseniiaSokolovaEpam sorry, I'm just seeing this now. We can fix this later, but `Utils.clone` returns a shallow copy of the object, so if you modify any nested properties of the skin config they will be modified for all test cases. This means that a test can affect the outcome of the next one.

We should keep the `JSON.parse(JSON.stringify` line. It's inefficient, but for the purpose of unit tests that doesn't matter.

jpfranco:
@KseniiaSokolovaEpam we usually review any new commits. It's ok to self-merge if there haven't been any changes after the approvals, but otherwise please let the original reviewers merge. Thanks!

PR #1083, opened 2018-06-19
[PLAYER-3475] Adding skip controls

jpfranco:
[TICKET](https://jira.corp.ooyala.com/browse/PLAYER-3475)

This is the base implementation of the skip controls for the skin.

Included in this PR:

- Skip controls component
- Ability to configure skip buttons and skip times
- Skip button accessibility
- New ControlButton component which encapsulates features that are common between control bar buttons and skip buttons
- Refactored ControlBar in order to switch to ControlButton component and remove some boiler plate code

Not included yet: 
- Unit tests. All unit test updates are just for compatibility with new code
- Some small missing features and tweaks

I will include the new unit tests in the next PR. There will be another PR for Mjolnir, but this can be merged independently. The skip buttons need to be turned on with the skin.json, so it should be safe to merge this even though it's not 100% complete yet.


js/components/controlBar.js
```diff
@@ -577,21 +445,16 @@ var ControlBar = React.createClass({
 
     var controlItemTemplates = {
       playPause: (
-        <AccessibleButton
+        <ControlButton
+          {...commonButtonProps}
           key={CONSTANTS.CONTROL_BAR_KEYS.PLAY_PAUSE}
-          className="oo-play-pause oo-control-bar-item"
```
aeng7: Do we no longer use the class oo-control-bar-item? Some customers may have some custom CSS on this class, so we may want to document this change if so.

js/components/controlBar.js
```diff
@@ -603,23 +466,15 @@ var ControlBar = React.createClass({
 
       volume: (
         <div key={CONSTANTS.CONTROL_BAR_KEYS.VOLUME} className="oo-volume oo-control-bar-item">
```
aeng7: If we're no longer using the class oo-control-bar-item, should we remove this one as well?

js/components/controlBar.js
```diff
@@ -638,48 +493,35 @@ var ControlBar = React.createClass({
...
             }.bind(this)}
             style={selectedStyle}
             className={qualityClass}
-            focusId={CONSTANTS.FOCUS_IDS.VIDEO_QUALITY}
+            focusId={CONSTANTS.CONTROL_BAR_KEYS.QUALITY}
             ariaLabel={CONSTANTS.ARIA_LABELS.VIDEO_QUALITY}
             ariaHasPopup={true}
             ariaExpanded={this.props.controller.state.videoQualityOptions.showPopover ? true : null}
+            icon="quality"
+            tooltip={CONSTANTS.SKIN_TEXT.VIDEO_QUALITY}
```
aeng7: Really nice, simplifying the icon and tooltip components

js/components/controlBar.js
```diff
@@ -772,28 +601,21 @@ var ControlBar = React.createClass({
...
-              ref={function(e) {
+            <ControlButton
+              {...commonButtonProps}
+              onRef={function(e) {
                 this.setToggleButtons(CONSTANTS.MENU_OPTIONS.MULTI_AUDIO, e);
               }.bind(this)}
               style={selectedStyle}
               className={multiAudioClass}
-              focusId={CONSTANTS.FOCUS_IDS.MULTI_AUDIO}
+              focusId={CONSTANTS.CONTROL_BAR_KEYS.AUDIO_AND_CC}
```
aeng7: Was it deliberate to change the focusId name to match the icon name? Just curious

js/components/controlButton.js
```diff
@@ -0,0 +1,316 @@
...
+      );
+      Utils.removeHighlight(iconElement, baseOpacity, baseColor);
+    }
+  },
+
+  /**
+   * Extracts the responsive UI multiple value from the skin config depending on the
+   * responsive view that is currently active.
+   * @private
+   * @return {Number} The numeric value of the UI multiple that matches the current responsive view.
```
aeng7: If "multiple" is referring to "multiplier," can we update the function name/jsdoc and variable names in the function to reference "multiplier" instead? Or if there is a decision to use "multiple," let me know.

js/components/controlButton.js
```diff
@@ -0,0 +1,316 @@
...
+  /**
+   * Chooses a component (either AccessibleButton or NonAccessibleButton) to use
+   * as base for the ControlButton depending on whether or not accessibility is
+   * enabled for this particular instance. Accessibility is enabled by default
+   * unless the ariaHidden prop is set to false explicitly.
+   * @private
+   * @return {Component} An AccessibleButton or NonAccessibleButton component, depending on the case
+   */
+  getButtonComponent: function() {
+    var Component;
```
aeng7: Is it standard to capitalize component when it is used like this?

js/components/controlButton.js
```diff
@@ -0,0 +1,316 @@
...
+  },
+
+  render: function() {
+    var Component = this.getButtonComponent();
+    var className = classNames('oo-control-bar-item', this.props.className);
+    var iconStyles = this.getIconStyles();
+    var areTooltipsEnabled = this.areTooltipsEnabled();
+
+    if (areTooltipsEnabled && this.props.tooltip) {
+      var responsiveUiMultiple = this.getResponsiveUiMultiple();
```
aeng7: Should these be declared outside of this if statement with a default value?

js/components/higher-order/holdOnClick.js
```diff
@@ -0,0 +1,125 @@
...
+ * mouse click as well as with the Space and Enter keys.
+ * @public
+ * @param {Component} ComposedComponent The React component to extend.
+ * The component must support adding handlers the following events: click, keydown, mousedown, mouseup
+ * @return {Component} A new component that supports holding the onClick handler
+ */
+var holdOnClick = function(ComposedComponent) {
+
+  var CLICK_HOLD_START_DELAY = 500;
+  var CLICK_HOLD_FREQUENCY = 100;
```
aeng7: Should we have these values be configurable?

js/components/skipControls.js
```diff
@@ -0,0 +1,276 @@
...
+        disabled={!this.props.config.hasPreviousVideos}
+        onClick={this.onPreviousVideo}>
+      </ControlButton>
+    );
+
+    buttonTemplate[CONSTANTS.SKIP_CTRLS_KEYS.SKIP_BACKWARD] = (
+      <HoldControlButton
+        {...this.props}
+        className="oo-center-button oo-skip-backward"
+        icon="replay"
```
aeng7: Are skip_backward and skip_forward just using the replay icon as placeholders? Or is it our final icon?

js/components/skipControls.js
```diff
@@ -0,0 +1,276 @@
...
+   * of the button.
+   */
+  getSortedButtonEntries: function() {
+    var buttons = [];
+    var buttonConfig = Utils.getPropertyValue(
+      this.props.skinConfig,
+      'skipControls.buttons',
+      {}
+    );
+    // Find de ids and indexes of all enabled buttons
```
aeng7: What are de ids?

js/components/skipControls.js
```diff
@@ -0,0 +1,276 @@
...
+    for (var buttonId in buttonConfig) {
+      var button = buttonConfig[buttonId];
+
+      if (button && button.enabled && this.shouldDisplayButton(buttonId)) {
```
aeng7: Is it possible to put the button.enabled check into this.shouldDisplayButton?

js/components/skipControls.js
```diff
@@ -0,0 +1,276 @@
...
+    buttons.sort(function(a, b) {
+      return a.index - b.index;
+    });
+    return buttons;
+  },
+
+  render: function() {
+    var buttons = this.getSortedButtonEntries();
+    // Nothing to render if we don't have buttons
+    if (!buttons.length) {
```
aeng7: What happens from an end user perspective when this happens? Will they just be sitting on a blank screen? Is this something we want to consider avoiding?

js/controller.js
```diff
@@ -149,6 +149,16 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
         delayedContentData: null
       },
 
+      scrubberBar: {
+        isHovered: false
```
aeng7: Would it make more sense to call this isHovering?

js/controller.js
```diff
@@ -1630,7 +1644,12 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       OO.log('onRelatedVideosFetched is called');
       if (relatedVideos.videos) {
         this.state.discoveryData = { relatedVideos: relatedVideos.videos };
-        this.state.upNextInfo.upNextData = relatedVideos.videos[0];
+
+        if (relatedVideos.upNextVideo) {
+          this.state.upNextInfo.upNextData = relatedVideos.upNextVideo;
+        } else {
+          this.state.upNextInfo.upNextData = relatedVideos.videos[0];
```
aeng7: This may be null or undefined. Is this ok? Should we also check for relatedVideos.videos.length in the parent if statement?

js/views/playingScreen.js
```diff
@@ -209,6 +216,72 @@ var PlayingScreen = React.createClass({
...
+    this.skipControlsClientRect = clientRect;
+    this.tryCheckMouseOverControls();
+  },
+
+  /**
+   * Checks to see whether or not the mouse is over the skip controls element and
+   * prevents the controls from autohiding if that is the case.
+   *
+   * IMPORTANT:
+   * There's a much simpler way to do this which is just listening to the
```
aeng7: Thanks for including this note

scss/components/_skip-controls.scss
```diff
@@ -0,0 +1,80 @@
...
+    color: #fff;
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    font-size: 0.55em;
+    text-align: center;
+    box-sizing: border-box;
+    backface-visibility: hidden;
```
aeng7: Why does this line have a red background? Do you see the same?

js/components/controlBar.js
jpfranco: `ControlButton`'s will have a `oo-control-bar-item` class by default. You can set an additional class, but they all have this as a base class.

js/components/controlBar.js
jpfranco: In this case we need to set the class manually since this is not a `ControlButton`.

js/components/controlBar.js
jpfranco: Yes, the focus id just needs to be any unique id, so we never really needed to have a separate set of ids for this purpose. I recently added the `CONTROL_BAR_KEYS` values, that's why we weren't using these before. It's simpler if there's a single id per button.

js/components/controlButton.js
jpfranco: Yes I just copied this over from the old code. I think it should be multiplier. Will update this.

js/components/controlButton.js
jpfranco: Yes, since this is a reference that is used directly in the JSX, where components are usually capitalized.

js/components/controlButton.js
jpfranco: I can update this if you prefer it, but I think it's a bit cleaner this way. These don't need a default value when tooltips aren't rendered, so the automatic hoisting achieves the same result. I'm ok with changing this though.

js/components/higher-order/holdOnClick.js
jpfranco: This is not currently a requirement so I think they'll be predefined for the first version. These values mimic the behavior of the browser's `keydown` event so that both mouse and keyboard controls behave consistently.

js/components/skipControls.js
jpfranco: Yes sorry, forgot to mention this. The font icons haven't been updated yet, so I'm currently just using existing icons and flipping them. This is temporary.

js/components/skipControls.js
jpfranco: These are the button names from the skin config. Will clarify this in the comments.

js/controller.js
jpfranco: Will change

js/controller.js
jpfranco: This is the way it's always been, but I'll add a check

js/components/higher-order/holdOnClick.js
aeng7: Got it, sounds good. No need to feature creep

js/components/controlButton.js
aeng7: Either way is fine. The default value was the main point rather than the hoisting

js/components/controlBar.js
aeng7: Got it

js/components/controlBar.js
aeng7: Got it

scss/components/_skip-controls.scss
jpfranco: I think it's because `backface-visibility` is an experimental technology. This can be considered as progressive enhancement though, browsers that don't support this (or deprecate it) will just ignore it. This is needed in order to prevent the button text from flickering during animations

aeng7: Sounds good

js/components/controlButton.js
```diff
@@ -0,0 +1,316 @@
...
+  },
+
+  /**
+   * Applies the highlighted styles to the button icon on hover.
+   * TODO
+   * This was moved over from ControlBar but the original implementation sets the
+   * styles directly on the rendered element, which is an anti-pattern in React.
+   * Eventually we should update this in order to accomplish the same with setState().
+   * @private
+   * @param {Event} event The mousenter event object that triggered the action
```
dallen42: @jpfranco Is it `mousenter` or `mouseenter`? Looks like you use `onMouseEnter` as the name for the later function...

jpfranco: Yes, it's with double `e`. Will fix this

js/controller.js
```diff
@@ -2094,6 +2114,62 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
+    setScrubberBarHoverState: function(isHovering) {
+      if (this.state.scrubberBar.isHovering !== isHovering) {
+        this.state.scrubberBar.isHovering = isHovering;
+        this.renderSkin();
+      }
+    },
+
+    /**
+     * "Previous video" button handler. Either rewinds the video or requests the
+     * previous video if the playhead is close to the beggining of the video or the
```
dallen42: @jpfranco mispelled "beginning"

js/constants/constants.js
```diff
@@ -249,7 +254,18 @@ module.exports = {
...
+    DEFAULT_SKIP_FORWARD_TIME: 10, // In seconds
+    MIN_SKIP_TIME: 1, // Min allowed value for skip controls, in seconds
+    MAX_SKIP_TIME: 60, // Max allowed value for skip controls, in seconds
+    // If the previous video button is clicked more than once, this
+    // is the maximum amount of time (in milliseconds) between calls that would
+    // cause the controller to request the previous video instead of rewinding
+    REQUEST_PREVIOUS_TIME_TRESHOLD: 1500,
+    // When the previous video button is clicked, if the playhead value (in seconds) is
+    // below this constant the controller will request the previous video instead of rewinding
+    REQUEST_PREVIOUS_PLAYHEAD_TRESHOLD: 2
```
dallen42: @jpfranco misspelled "threshold" in the constant name

js/controller.js
jpfranco: Fixed

js/constants/constants.js
jpfranco: Fixed. Good catch, I think I've never spelled this word properly

PR #1085, opened 2018-06-26
Player 3765

aeng7:
**Please do not merge yet**, opening for review only. We're still figuring out how to integrate this with our automation and build processes.

Updated Node to latest stable (8.11.3) and NPM to 5.6.0. Updated ReactJS to 16.4.1.

Overview of changes:
-React.createClass was deprecated. Using an alternative convenience package, create-react-class as a placeholder until we update to newer React workflows
-React.PropTypes was deprecated. Using an alternative convenience package, prop-types
-We now use babelify instead of reactify in our build script
-The jsdom we are now using does not allow reading and/or writing of certain window.navigator properties. Some APIs such as window.open are also not supported. Workarounds and mocks are included in our test setup file at /scripts/jest/setup.js
-TestUtils was deprecated. Using Airbnb's testing framework, Enzyme, as a replacement. Enzyme documentation at:
https://airbnb.io/enzyme/docs/api/mount.html
The following are some of the general changes moving to Enzyme:

1. TestUtils.renderIntoDocument and other render methods have been replaced by Enzyme.mount, which returns a ReactWrapper object
2. TestUtils.findRenderedComponentWith* have been replaced by ReactWrapper.find(selector)
3. TestUtils.findRenderedDOMComponentWith* and ReactDOM.findDOMNode have been replaced by ReactWrapper.getDOMNode
4. TestUtils.Simulate was replaced by ReactWrapper.simulate wherever possible.
5. We now use ReactWrapper.ref to retrieve a ref

-I've updated a few files to fix some benign console errors that our unit tests were throwing. Note that a lot of these errors are still thrown, mainly due to the NonAccessibleButton. These will need to be addressed at a later date. Unit tests still work fine and pass successfully. 

Note that there were a small number of unit tests I could not port over to Enzyme. These have been marked with TODOs


aeng7:
Please do not merge yet, opening for review only. We're still figuring out how to integrate this with our automation and build processes.

js/components/thumbnailContainer.js
```diff
@@ -107,6 +109,18 @@ var ThumbnailContainer = React.createClass({
     }
   },
 
+  getClientWidth: function(element) {
```
jpfranco: I think these might be a bit generic for thumbnailcontainer. Would it be better to move these to utils?

aeng7: Sounds good, will do.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.1",
   "description": "HTML5 skin over an Ooyala video player",
   "engines": {
-   "node" : ">=0.10.33 <=0.10.46"
+    "node": ">=8.11.3"
```
jpfranco: This is not needed yet, but we probably should look into using  `package-lock.json` instead of shrinkwrap now that we have a node version that supports it. Not sure if it's worth the effort though, just thinking out loud.

package.json
```diff
@@ -24,44 +24,54 @@
...
+    "gulp-rename": "^1.3.0",
+    "gulp-replace": "^1.0.0",
+    "gulp-sass": "^4.0.1",
+    "gulp-shell": "^0.6.5",
+    "gulp-sourcemaps": "^2.6.4",
+    "gulp-uglify": "^3.0.0",
+    "gulp-util": "^3.0.8",
+    "jest": "^23.1.0",
+    "jest-cli": "^23.1.0",
+    "jquery": "^3.3.1",
```
jpfranco: Not 100% sure, but I think the skin actually gets the jquery reference from the core. If that's the case then it might make sense to keep the old version in order to test with the one that will actually be loaded by the player.

package.json
```diff
@@ -24,44 +24,54 @@
   },
   "dependencies": {
     "classnames": "*",
+    "create-react-class": "^15.6.3",
     "deepmerge": "^1.4.0",
     "lodash.debounce": "4.0.3",
     "lodash.values": "4.1.0",
-    "react": "0.14.7",
-    "react-dom": "0.14.7",
```
jpfranco: Looks like we were only 16 major versions behind :)

tests/components/closed-caption/closedCaptionPopover-test.js
```diff
@@ -81,38 +81,66 @@ describe('ClosedCaptionPopover', function() {
...
 
+    //TODO: I couldn't find an easy way to simulate the keydown event with the proper event target,
+      //so I'm calling onKeyDown manually for the below tests for now
     it('should focus on previous menu item when pressing UP or LEFT arrow keys', function() {
       var activeIndex = popoverMenuItems.length - 1;
-      document.activeElement = popoverMenuItems[activeIndex];
-      popover.dispatchEvent(getMockKeydownEvent(document.activeElement, CONSTANTS.KEYCODES.UP_ARROW_KEY));
+      popoverMenuItems[activeIndex].focus();
+      wrapper.instance().onKeyDown({
+        keyCode: CONSTANTS.KEYCODES.UP_ARROW_KEY,
```
jpfranco: Could we keep the `getMockKeydownEvent` in order to avoid setting up the even to object each times. If the object changes in the future it will be easier to modify

tests/components/controlBar-test.js
```diff
@@ -272,20 +252,20 @@ describe('ControlBar', function() {
...
     );
 
-    var volumeButton = TestUtils.findRenderedDOMComponentWithClass(DOM, 'oo-mute-unmute').firstChild;
-    TestUtils.Simulate.click(volumeButton);
+    var volumeButton = wrapper.find('.oo-mute-unmute').hostNodes();
+    volumeButton.simulate('click');
     expect(muteClicked).toBe(true);
-    var volumeBars = TestUtils.scryRenderedDOMComponentsWithClass(DOM, 'oo-volume-bar');
+    var volumeBars = wrapper.find('.oo-volume-bar').hostNodes();
     // JEST doesn't support dataset at the time of writing
```
jpfranco: I think this comment might be obsolete now

tests/components/videoQualityPanel-test.js
```diff
@@ -268,65 +272,96 @@ describe('VideoQualityPanel', function() {
...
     });
 
+    //TODO: I couldn't find an easy way to simulate the keydown event with the proper event target,
+      //so I'm calling onKeyDown manually for the below tests for now
     it('should focus on previous menu item when pressing UP or LEFT arrow keys', function() {
       var activeIndex = qualityButtons.length - 1;
-      document.activeElement = qualityButtons[activeIndex];
-      qualityPanel.dispatchEvent(getMockKeydownEvent(document.activeElement, CONSTANTS.KEYCODES.UP_ARROW_KEY));
+      qualityButtons[activeIndex].focus();
+      wrapper.instance().onKeyDown({
```
jpfranco: Same comment as above, would be better to keep the `getMockKeydownEvent` helper function?

package.json
aeng7: Yep, you're correct. I'll change this back.

aeng7: Node is throwing errors when attempting to install our older version (1.8.3). I'll look into this a little to see if it is easy to resolve this. However, it may not be possible.

package.json
aeng7: We should've waited until React 16.4.7 was released prior to the update to reach that exact 16 major versions behind goal

tests/components/closed-caption/closedCaptionPopover-test.js
aeng7: Will do

tests/components/controlBar-test.js
aeng7: Removed

tests/components/videoQualityPanel-test.js
aeng7: Will do

js/views/pauseScreen.js
```diff
@@ -309,7 +311,7 @@ var PauseScreen = React.createClass({
           <ControlBar
             {...this.props}
             controlBarVisible={this.state.controlBarVisible}
-            playerState={this.state.playerState}
+            playerState={this.props.playerState}
```
aldigun: Is this correctly using props instead of state?

aeng7: Yep, that is how it is passed in other views, as a prop. Considering all views are created in skin.js and are created in the same way, and our new unit test framework was complaining about this, I made this change.

PR #1086, opened 2018-06-26
[player-3811]

aeng7:
-SET_CURRENT_AUDIO will now be published with the current video id


aeng7:
Should be merged along with mjolnir PR 500

aeng7:
@KseniiaSokolovaEpam @DmitriiMaslov I've made a change to how SET_CURRENT_AUDIO works. Previously it would trigger on the focused video element, which would cause issues if an ad video element comes into focus prior to the audio change. Please take a look at this. Let me know if you have any questions, thanks!

KseniiaSokolovaEpam:
LGTM

PR #1087, opened 2018-06-27
[PLAYER-3475] Updated icons and UI fixes

jpfranco:
Updated the icon fonts and implemented the last UX suggestions made by Fernando. Unit tests are still pending.


js/components/scrubberBar.js
```diff
@@ -222,17 +222,21 @@ var ScrubberBar = React.createClass({
...
-  handleScrubberBarMouseLeave: function(evt) {
+  handleScrubberBarMouseOut: function(evt) {
     if (!this.props.controller.state.thumbnails) return;
     this.props.controller.setScrubberBarHoverState(false);
     this.setState({
       hoveringX: 0
     });
   },
 
+  handleScrubberBarMouseLeave: function(evt) {
```
aeng7: How come mouseLeave is treated differently from mouseOut?

js/controller.js
```diff
@@ -2149,14 +2149,16 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
     rewindOrRequestPreviousVideo: function() {
       var currentTimestamp = Utils.getCurrentTimestamp();
       var timeElapsed = currentTimestamp - this.state.skipControls.requestPreviousTimestamp;
-      // Button has been clicked once more in a short amount of time or playhead
-      // is below a certain threshold, request previous video.
+      // Button has been clicked once more in a short amount of time, playhead
+      // is below a certain threshold, or player is stalled after seeking, request previous video.
```
aeng7: If the user seeks waaay forward via the control bar, and then starts seeking backwards rapidly via backward button into available but unbuffered content, is it possible for the user to go back to the previous video? If so, I think this is undesirable, right?

jpfranco: This applies to the "Back" button only, not the "Seek back" button. It is expected for the Back button to take you to the previous video if you click on it more than once

aeng7: How come the back button is affected by the current seeking state of the video? Just curious

js/components/scrubberBar.js
```diff
@@ -398,17 +402,16 @@ var ScrubberBar = React.createClass({
       <div
         className="oo-scrubber-bar-container"
         ref="scrubberBarContainer"
-        onMouseEnter={scrubberBarMouseEnter}
-        onMouseLeave={scrubberBarMouseLeave}
-        onMouseMove={scrubberBarMouseMove}
-      >
+        onMouseOver={scrubberBarMouseOver}
+        onMouseOut={scrubberBarMouseOut}
+        onMouseLeave={this.handleScrubberBarMouseLeave}
```
aldigun: why this is the only one that is using "this."?

jpfranco: It is the only one that was added recently. The other two were like that before, but I had changed them on the previous PR, and now I reverted them because the change had caused an issue. 
The old code was storing the handler in another variable for some reason, but this is unecessary

js/controller.js
jpfranco: This is because the first click rewinds the video, but the playhead might take some time to update. This check ensures that you move to the previous video when clicking again, even if the playhead is not below the threshold.

aeng7:
LGTM

PR #1089, opened 2018-06-27
[update-readme-with-new-node]

aeng7:
-updating readme to reference Node version 8.11.3

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,9 +59,7 @@ This simple test HTML page can also be hosted on your environment to showcase ht
 
 5. After cloning, just `cd` in there, install dependencies by running `npm install`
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,7 +51,7 @@ This simple test HTML page can also be hosted on your environment to showcase ht
 ## Developer Setup
 1. [Install git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)
 
-2. [Install Node.js] (https://nodejs.org/download/release/v0.10.33/) v0.10.33
+2. [Install Node.js] (https://nodejs.org/download/release/v8.11.3/) v8.11.3
```
jpfranco: Do you know if we still need the dedupe part?

aeng7: I don't believe we do. I'll remove it from the readme.

PR #1090, opened 2018-06-28
[PLAYER-3807] - Volume controls stop responding after a pause/resume

ipadilla4:
When pause event came from "ad mode", currentVideoId was "ads", preventing volume controls to work fine. We need to ensure currentVideoId is always "main" for SSAI ads.


js/controller.js
```diff
@@ -830,12 +830,16 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
 
     onVcPlay: function(event, source) {
       this.state.currentVideoId = source;
-      if (this.state.adWasPaused && this.state.currentAdsInfo && this.state.currentAdsInfo.currentAdItem && this.state.currentAdsInfo.currentAdItem.ssai) {
-        this.state.adPauseDuration = Date.now() - this.state.adPausedTime;
-        //we calculate new ad end time, based on the time that the ad was paused.
-        this.state.adEndTime = this.state.adEndTime + this.state.adPauseDuration; //milliseconds
-        this.state.adWasPaused = false;
-        this.state.adPauseDuration = 0;
+      if (this.state.currentAdsInfo.currentAdItem.ssai) {
```
dallen42: @ipadilla4 we should not check `this.state.currentAdsInfo.currentAdItem.ssai` before executing the null checks for `this.state.currentAdsInfo && this.state.currentAdsInfo.currentAdItem` that are inside the nested if. I would put `this.state.currentAdsInfo && this.state.currentAdsInfo.currentAdItem` as the top level check, with individual checks for `this.state.currentAdsInfo.currentAdItem.ssai` (during which we set currentVideoId to OO.VIDEO.MAIN) and `this.state.adWasPaused` (where we handle the pause logic) inside.

dallen42: @ipadilla4 I think you need the null checks that are in the nested if statement here. This check will cause an exception if `this.state.currentAdsInfo.currentAdItem` or `this.state.currentAdsInfo` is null. Is the solution to just set `this.state.currentVideoId` to `OO.VIDEO.MAIN`? If so, then I think we could remove this outer if and move `this.state.currentVideoId = OO.VIDEO.MAIN` inside the previous `if (this.state.adWasPaused && this.state.currentAdsInfo && this.state.currentAdsInfo.currentAdItem && this.state.currentAdsInfo.currentAdItem.ssai)` statement

dallen42:
@ipadilla4 is there any unit test we can add for this?

js/controller.js
```diff
@@ -830,12 +830,16 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
+        //When pause come from ad source, for SSAI we still want source to be main, for consistency.
+        this.state.currentVideoId = OO.VIDEO.MAIN;
```
aeng7: Is something changing the currentVideoId from OO.VIDEO.MAIN for ssai?

js/controller.js
ipadilla4: Yes this is needed only for ssai ads. I forgot to add the null checks to the outer if. We do need two ifs because `onVcPlay` method is called twice, one for `"main"` source and one for `"ads"` . When parameter `source` is `"ads"`, `this.state.adWasPaused` will be false.

js/controller.js
ipadilla4: Well I realized `onVcPlay` method get called twice, one with `main` as source and another one with `ads` source.  So we are preventing `currentVideoId` is overwritten with `ads` for ssai ads.

js/controller.js
```diff
@@ -830,12 +830,16 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
       this.state.currentVideoId = source;
-      if (this.state.adWasPaused && this.state.currentAdsInfo && this.state.currentAdsInfo.currentAdItem && this.state.currentAdsInfo.currentAdItem.ssai) {
-        this.state.adPauseDuration = Date.now() - this.state.adPausedTime;
-        //we calculate new ad end time, based on the time that the ad was paused.
-        this.state.adEndTime = this.state.adEndTime + this.state.adPauseDuration; //milliseconds
-        this.state.adWasPaused = false;
-        this.state.adPauseDuration = 0;
+      if (this.state.currentAdsInfo && this.state.currentAdsInfo.currentAdItem && this.state.currentAdsInfo.currentAdItem.ssai) {
+        //When pause come from ad source, for SSAI we still want source to be main, for consistency.
+        this.state.currentVideoId = OO.VIDEO.MAIN;
```
dallen42: @ipadilla4 this looks ok for a quick patch but I am concerned that it does not address the root cause. Can we find out why we are receiving video id for both `OO.VIDEO.MAIN` and `OO.VIDEO.AD` here? I suspect there is something we need to change in the event itself to distinguish between SSAI and CSAI ads.

ipadilla4: I found out that from core plugin we are sending this `OO.VIDEO.ADS` videoId in the event `willResumeAds`. I agree that we need to address the root cause so I will double check that we always get `OO.VIDEO.MAIN` when having an ssai ad.

dallen42: @ipadilla4 got it, I think the solution then is to have a check in willResumeAds if we are playing ssai ads, and not publish `OO.VIDEO.ADS` if that is the case. We might be able to use `notifySSAIAdPlaying` and `notifySSAIAdPlayed` to keep track of this, or `notifyLinearAdStarted` and `notifyLinearAdEnded` but we would have to modify those to differentiate ssai ads

ipadilla4: @dallen42 yes we are currently using `notifyLinearAdStarted` and `notifyLinearAdEnded`, but we don't send videoId as 'OO.VIDEO.ADS`, we only send ad properties when publishing those events. So I think for ssai ads we are good.

dallen42:
@ipadilla4 do we still need this PR or can we close it?

PR #1093, opened 2018-06-29
Player 3808

aeng7:
-updated PauseScreen component to be a class rather than a function
-added a Babel plugin to support public class fields syntax
-added support for auto hiding pause screens
-removed mixins from PauseScreen during the transition to a class. A post about why here: https://reactjs.org/blog/2016/07/13/mixins-considered-harmful.html

### DIFF
--- a/js/views/pauseScreen.js
+++ b/js/views/pauseScreen.js
@@ -16,6 +16,8 @@ var React = require('react'),
     ViewControlsVr = require('../components/viewControlsVr');
 
 class PauseScreen extends React.Component{
+  //Alex: When transitioning to a class, I've removed the usage of mixins since they have been
+  //deprecated. I've brought in the animate and resize mixin functionalities directly into this class.
   constructor(props) {
     super(props);
     this.state = {
@@ -60,18 +62,31 @@ class PauseScreen extends React.Component{
     clearTimeout(this.animateTimer);
   }
 
+  /**
+   * Originally part of the animationMixin. Sets the animate state to true. Used
+   * to animate the fade of the pause button in the center of the screen upon pausing.
+   * @private
+   */
   startAnimation = () => {
     this.setState({
       animate: true
     });
   };
 
+  /**
+   * Show the pause screen by setting the hidden state to false.
+   * @private
+   */
   showPauseScreen = () => {
     this.setState({
       hidden: false
     });
   };
 
+  /**
+   * Hide the pause screen by setting the hidden state to true.
+   * @private
+   */
   hidePauseScreen = () => {
     this.setState({
       hidden: true
@@ -179,16 +194,22 @@ class PauseScreen extends React.Component{
     }
   };
 
+  /**
+   * MouseLeave event handler. Hides the pause screen if the autoHide skin config is set to true.
+   * @private
+   */
   handleMouseLeave = () => {
     if (this.props.skinConfig.pauseScreen.autoHide) {
       this.hidePauseScreen();
     }
   };
 
+  /**
+   * MouseEnter event handler. Shows the pause screen.
+   * @private
+   */
   handleMouseEnter = () => {
-    if (this.props.skinConfig.pauseScreen.autoHide) {
-      this.showPauseScreen();
-    }
+    this.showPauseScreen();
   };
 
   render() {
@@ -309,8 +330,7 @@ class PauseScreen extends React.Component{
           onMouseDown={this.handlePlayerMouseDown}
           onTouchStart={this.handlePlayerMouseDown}
           onMouseUp={this.handlePlayerMouseUp}
-          onTouchEnd={this.handleTouchEnd}
-        />
+          onTouchEnd={this.handleTouchEnd} />
 
         <Watermark
           {...this.props}
--- a/js/views/pauseScreen.js
+++ b/js/views/pauseScreen.js
@@ -9,30 +9,33 @@ var React = require('react'),
     UpNextPanel = require('../components/upNextPanel'),
     TextTrackPanel = require('../components/textTrackPanel'),
     Watermark = require('../components/watermark'),
-    ResizeMixin = require('../mixins/resizeMixin'),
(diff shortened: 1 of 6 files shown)


aeng7:
I'll need to update the skin config submodule after https://github.com/ooyala/skin-config/pull/210 is merged

js/views/pauseScreen.js
```diff
@@ -9,30 +9,35 @@ var React = require('react'),
...
-var createReactClass = require('create-react-class');
-var PropTypes = require('prop-types');
 
-var PauseScreen = createReactClass({
-  mixins: [ResizeMixin, AnimateMixin],
-
-  getInitialState: function() {
-    return {
+class PauseScreen extends React.Component{
+  //Alex: When transitioning to a class, I've removed the usage of mixins since they have been
```
jpfranco: I think this might be a good candidate for an HOC

js/views/pauseScreen.js
```diff
@@ -109,52 +149,70 @@ var PauseScreen = createReactClass({
...
     }
-  },
+  };
 
-  render: function() {
+  /**
+   * MouseLeave event handler. Hides the pause screen if the autoHide skin config is set to true.
+   * @private
+   */
+  handleMouseLeave = () => {
```
jpfranco: Would there be a way for the controls to auto hide when the video is in fullscreen mode since the mouse can't leave the container?

js/views/pauseScreen.js
```diff
@@ -109,52 +149,70 @@ var PauseScreen = createReactClass({
...
+      this.hidePauseScreen();
+    }
+  };
+
+  /**
+   * MouseEnter event handler. Shows the pause screen.
+   * @private
+   */
+  handleMouseEnter = () => {
+    this.showPauseScreen();
```
jpfranco: Could this be called conditionally depending on configuration in order to avoid re-rendering?

js/views/pauseScreen.js
```diff
@@ -253,8 +311,17 @@ var PauseScreen = createReactClass({
...
+    var pauseScreenClass = ClassNames({
+      'oo-state-screen': true,
+      'oo-pause-screen': true,
+      'oo-pause-screen-hidden': this.state.hidden
+    });
+
     return (
-      <div className="oo-state-screen oo-pause-screen">
+      <div className={pauseScreenClass}
+        onMouseEnter={this.handleMouseEnter}
```
jpfranco: Would this cover cases in which the pause screen is triggered via keyboard controls?

scss/components/_pause-screen.scss
```diff
@@ -1,11 +1,18 @@
 
 .oo-pause-screen {
+  opacity: 1;
```
jpfranco: For consistency, would it be better to use the same autohide animation that we use on the playing screen instead of fading the pause screen?

js/views/pauseScreen.js
```diff
@@ -43,17 +48,52 @@ var PauseScreen = createReactClass({
     if (this.state.containsText) {
       this.props.controller.addBlur();
     }
-  },
 
-  componentWillUnmount: function() {
+    this.animateTimer = setTimeout(this.startAnimation, 1);
```
jpfranco: Do we need a constant for the timer delay?

js/views/pauseScreen.js
```diff
@@ -323,5 +390,6 @@ var PauseScreen = createReactClass({
       </div>
     );
   }
-});
+}
+
```
jpfranco: Do we also need to enable this behavior on the end screen? I think that from an end-user perspective it might be perceived as being the same or at least a similar state

js/views/pauseScreen.js
aeng7: I agree too. I should've mentioned this might be temporary until we come up with a design solution, but an alternative in the meanwhile is just replicate our mixins as HOCs. What do you think?

js/views/pauseScreen.js
aeng7: This sounds like a product question. I'll bring it up with Steve.

js/views/pauseScreen.js
aeng7: Will do.

js/views/pauseScreen.js
aeng7: I'll ask Steve how autohide should work with Keyboard controls.

scss/components/_pause-screen.scss
aeng7: Will do.

js/views/pauseScreen.js
aeng7: I just copied this over from the animate mixin. I'll add one in here or in the HOC if we go that route.

js/views/pauseScreen.js
aeng7: I'll bring this up with Steve.

js/views/pauseScreen.js
jpfranco: Sure, sounds good to me

js/views/pauseScreen.js
jpfranco: Ok thanks. If he does want to enable that we could probably re-use the logic from the playing screen, it already handles this scenario

js/views/pauseScreen.js
aeng7: Steve says we should leave the end screen as-is right now.

js/views/pauseScreen.js
aeng7: After talking with Steve, we'll handle keyboard controls in a separate ticket.

js/views/pauseScreen.js
aeng7: It looks like Amazon changed their behavior. Steve would like us to mimic the new one, which is autohide after 1.5 seconds if the mouse is not over one of our buttons.

aeng7:
I'm closing this PR as we adjust to the new requirements

PR #1094, opened 2018-06-29
feature/PLAYER-3794 ISO-639-2 instead of ISO-639-1

KseniiaSokolovaEpam:
**IMPORTANT!**
Before merging the PR, [PR](https://github.com/ooyala/html5-common/pull/80) to html5-common should be merged.

We have an [issue](https://jira.corp.ooyala.com/browse/PLAYER-3794). 
The issue happened because of we use ISO-639-1, and the standard does not have special codes.
In the PR we use codes for ISO-639-2 as main codes. And the map with languages was incresed according the table from [wiki](https://en.wikipedia.org/wiki/List_of_ISO_639-2_codes) .
Some of function were changed. Now we check if the code is one of special codes, and we show a label of a track according by [rules for the task](https://jira.corp.ooyala.com/browse/PLAYER-3134)

A concomitant PR is https://git.corp.ooyala.com/projects/PBW/repos/mjolnir/pull-requests/503

Unit tests are passed.


js/components/closed-caption-multi-audio-menu/helpers.js
```diff
@@ -41,9 +42,8 @@ function getDisplayLanguage(languagesList, languageCode) {
   ) {
     var matchingLanguage = _.find(languagesList, function(language) {
       return (
-        language['1'] === languageCode ||
-        language['2'] === languageCode ||
         language['2T'] === languageCode ||
+        language['1'] === languageCode ||
```
aeng7: Just to double check, do we no longer care about language['2']?

js/components/closed-caption-multi-audio-menu/helpers.js
```diff
@@ -57,31 +57,72 @@ function getDisplayLanguage(languagesList, languageCode) {
...
+  );
+  localizedLanguage = localizedLanguage ?
+    localizedLanguage
+    :
+    phrase ? phrase : '';
+  return localizedLanguage;
+}
+
+/**
+ *
```
aeng7: Can we add a description for this function?

aeng7:
LGTM outside of a few minor comments/questions

js/components/closed-caption-multi-audio-menu/helpers.js
KseniiaSokolovaEpam: \+

js/components/closed-caption-multi-audio-menu/helpers.js
KseniiaSokolovaEpam: I found a table on [wiki](https://en.wikipedia.org/wiki/List_of_ISO_639-2_codes) . It looks like the most full table with codes. I have grabbered it. So now in our map (language.js file) we have all single values for ISO-639-2 from first column and values with asterisk from first column , they are marked as "2B", and values without asterisk are marked as "2T".

About language['x']:
There is keys "1", "2B", "2T", "3" in languages.js . These are types of ISO-639 .
You can read about the types here: 
https://en.wikipedia.org/wiki/ISO_639-1
https://en.wikipedia.org/wiki/ISO_639-2
https://en.wikipedia.org/wiki/ISO_639-3

And we check if the langCode is one of these codes.

aeng7: Thanks for the great explanation @KseniiaSokolovaEpam !

aeng7:
Note there is a conflict in the mjolnir PR

pilievOoyala:
LGTM but please remove that debug log file first and then you can merge this yourself.

KseniiaSokolovaEpam:
@pilievOoyala ok (I deleted unnecessary Stereoscopic, Video and npm-debug.log files), thank you!

PR #1095, opened 2018-06-29
[PLAYER-3835] Fix for rewind with loop + controller unit tests for Skip Ctrls

jpfranco:
Fixes an issue in which rewind doesn't happen if `loop=true`. Also added the first set of unit tests for the Skip Controls. Will continue adding more of these.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -2159,7 +2159,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
         this.mb.publish(OO.EVENTS.REQUEST_PREVIOUS_VIDEO);
       } else {
         this.updateSeekingPlayhead(0);
-        this.mb.publish(OO.EVENTS.REPLAY);
+        this.seek(0);
       }
       this.state.skipControls.requestPreviousTimestamp = currentTimestamp;
     },
--- a/tests/html5skin-test.js
+++ b/tests/html5skin-test.js
@@ -1036,6 +1036,115 @@ describe('Controller', function() {
     });
   });
 
+  describe('Scrubber Bar', function() {
+    var spyRender;
+
+    beforeEach(function() {
+      spyRender = sinon.spy(controller, 'renderSkin');
+    });
+
+    afterEach(function() {
+      spyRender.restore();
+    });
+
+    it('should update scrubber bar hover state and render when setScrubberBarHoverState() is called', function() {
+      controller.state.scrubberBar.isHovering = false;
+      controller.setScrubberBarHoverState(true);
+      expect(controller.state.scrubberBar.isHovering).toBe(true);
+      expect(spyRender.callCount).toBe(1);
+    });
+
+    it('should NOT render when setScrubberBarHoverState() is called with same value', function() {
+      controller.state.scrubberBar.isHovering = false;
+      controller.setScrubberBarHoverState(false);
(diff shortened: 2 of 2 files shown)


js/controller.js
```diff
@@ -2159,7 +2159,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
         this.mb.publish(OO.EVENTS.REQUEST_PREVIOUS_VIDEO);
       } else {
         this.updateSeekingPlayhead(0);
-        this.mb.publish(OO.EVENTS.REPLAY);
```
aldigun: REPLAY event is not published?

jpfranco: Not anymore, it will just perform a seek. I think it makes sense since `REPLAY` currently happens only after the whole video has been watched, so rewinding is not the same thing. If you think this should be fired I can look for another solution though.

tests/html5skin-test.js
```diff
@@ -1036,6 +1036,115 @@ describe('Controller', function() {
...
+        expect(spyUpdatePlayhead.calledWith(0)).toBe(true);
+        expect(spySeek.calledWith(0)).toBe(true);
+      });
+
+      it('should request previous video on first "Previous Video" click if playhead is below threshold', function() {
+        controller.state.mainVideoPlayhead = CONSTANTS.UI.REQUEST_PREVIOUS_PLAYHEAD_THRESHOLD - 1;
+        controller.rewindOrRequestPreviousVideo();
+        expect(spyUpdatePlayhead.callCount).toBe(0);
+        expect(spySeek.callCount).toBe(0);
+        expect(spyPublish.callCount).toBe(1);
```
aeng7: Should we check what the contents of this publish are?

tests/html5skin-test.js
```diff
@@ -1036,6 +1036,115 @@ describe('Controller', function() {
...
+        controller.state.seeking = true;
+        controller.state.mainVideoPlayhead = CONSTANTS.UI.REQUEST_PREVIOUS_PLAYHEAD_THRESHOLD + 1;
+        controller.rewindOrRequestPreviousVideo();
+        expect(spyUpdatePlayhead.callCount).toBe(0);
+        expect(spySeek.callCount).toBe(0);
+        expect(spyPublish.callCount).toBe(1);
+        expect(spyPublish.calledWith(OO.EVENTS.REQUEST_PREVIOUS_VIDEO)).toBe(true);
+      });
+
+      it('should request previous video on second "Previous Video" click if time elapsed since last call is below treshold', function() {
```
aeng7: What is this testing? Is it when the first click is above the threshold and the second click is below the threshold (as if the user seeked to earlier or something)?

aeng7:
LGTM other than a small question

tests/html5skin-test.js
jpfranco: This is testing that the previous video is requested if the function is called with a time difference below the time threshold. The playhead is set above the `REQUEST_PREVIOUS_PLAYHEAD_THRESHOLD` in order to simulate an initial scenario in which the player rewinds.

aeng7: Got it

js/controller.js
pilievOoyala: I think that makes sense. For analytics I think we want to separate out when the user wants to watch the content again versus rewinding halfway through.

PR #1096, opened 2018-07-03
[PLAYER-3831][v4] [VAST] [Ad countdown] - Ad countdown is stuck at 0

ipadilla4:
Bug fix for ad countdown.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -1445,7 +1445,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
           }
         }
       } else {
-        remainingTime = parseInt(this.state.adVideoDuration - (this.state.adVideoPlayhead * 100));
+        remainingTime = parseInt(this.state.adVideoDuration - (this.state.adVideoPlayhead));
       }
       return remainingTime;
     },
--- a/js/controller.js
+++ b/js/controller.js
@@ -1445,11 +1445,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
           }
         }
       } else {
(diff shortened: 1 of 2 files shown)


js/controller.js
```diff
@@ -1445,7 +1445,11 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
           }
         }
       } else {
-        remainingTime = parseInt(this.state.adVideoDuration - (this.state.adVideoPlayhead * 100));
+        if (isSSAI) {
+          remainingTime = parseInt(this.state.adVideoDuration - (this.state.adVideoPlayhead * 100));
+        }else {
```
aeng7: There should be a space between } and else

js/controller.js
```diff
@@ -1445,7 +1445,11 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
           }
         }
       } else {
-        remainingTime = parseInt(this.state.adVideoDuration - (this.state.adVideoPlayhead * 100));
+        if (isSSAI) {
+          remainingTime = parseInt(this.state.adVideoDuration - (this.state.adVideoPlayhead * 100));
```
aeng7: What units are adVideoPlayhead in for SSAI? Is it possible to have the adVideoPlayhead be the same units for both SSAI and non-SSAI? This will help maintenance down the road.

ipadilla4: Actually I just checked that SSAI is always "Live" even for VOD, so those assets will never enter in this else condition. I will remove this if condition.

aeng7:
LGTM

PR #1098, opened 2018-07-06
[bugfix/PLAYER-3877] only mute/unmute on iOs

KseniiaSokolovaEpam:
**IMPORTANT!** 
Merge only in case if my approach in [task](https://jira.corp.ooyala.com/browse/PLAYER-3877) is approved.

Now on iOs we can not change sound by JavaScript ([link](http://blog.millermedeiros.com/unsolved-html5-video-issues-on-ios/)). So we do not show volume slider on iOs devices and an user can only switch on/switch off sound programmatically and have to use manual buttons for changing volume.

Unit tests are passed.

### DIFF
--- a/tests/components/controlBar-test.js
+++ b/tests/components/controlBar-test.js
@@ -110,7 +110,7 @@ describe('ControlBar', function() {
       OO_setWindowNavigatorProperty('userAgent', 'desktop');
     });
 
-    it('render one stereo button if content vr', function() {
+    it('vr: render one stereo button if content vr', function() {
       baseMockProps.skinConfig.buttons.desktopContent = [{'name':'stereoscopic', 'location':'controlBar', 'whenDoesNotFit':'keep', 'minWidth':35 }];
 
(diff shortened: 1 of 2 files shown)


tests/components/controlBar-test.js
```diff
@@ -191,6 +191,33 @@ describe('ControlBar', function() {
...
+      var volumeButton = wrapper.find('.oo-mute-unmute').hostNodes();
+      volumeButton.simulate('click');
+      expect(isMuted).toBe(true);
+      expect(isSliderShown).toBe(false);
+      volumeButton.simulate('click');
+      expect(isMuted).toBe(false);
+      expect(isSliderShown).toBe(false);
+      OO_setWindowNavigatorProperty('platform', '');
+    });
+
```
aeng7: Can we add a unit test to test that the volume controls are hidden on iOS?

aeng7:
LGTM other than a comment about unit tests

PR #1099, opened 2018-07-09
[PLAYER-3475] Unit tests for Skip Controls feature

jpfranco:
Included in this PR:

- Added unit tests for Skip Controls
- Fixed all remaining warnings that used to show up when running unit tests
- Updated new Skip Control-related components in order to use newer React patterns

### DIFF
--- a/js/components/controlButton.js
+++ b/js/components/controlButton.js
@@ -1,11 +1,11 @@
-var React = require('react');
-var classNames = require('classnames');
-var AccessibleButton = require('./accessibleButton');
-var Icon = require('./icon');
-var Tooltip = require('./tooltip');
-var Utils = require('./utils');
-var CONSTANTS = require('../constants/constants');
-var PropTypes = require('prop-types');
+const React = require('react');
+const classNames = require('classnames');
+const AccessibleButton = require('./accessibleButton');
+const Icon = require('./icon');
+const Tooltip = require('./tooltip');
+const Utils = require('./utils');
+const PropTypes = require('prop-types');
+const CONSTANTS = require('../constants/constants');
 
 /**
  * Template component that is used for buttons that don't have accessibility enabled.
--- a/js/components/controlButton.js
+++ b/js/components/controlButton.js
@@ -249,7 +249,8 @@ class ControlButton extends React.Component {
     const iconStyles = this.getIconStyles();
     const areTooltipsEnabled = this.areTooltipsEnabled();
 
-    let responsiveUiMultiplier, tooltipVerticalOffset;
+    let responsiveUiMultiplier;
+    let tooltipVerticalOffset;
 
     if (areTooltipsEnabled && this.props.tooltip) {
       responsiveUiMultiplier = this.getResponsiveUiMultiplier();
--- a/js/components/controlButton.js
+++ b/js/components/controlButton.js
@@ -5,7 +5,6 @@ var Icon = require('./icon');
 var Tooltip = require('./tooltip');
 var Utils = require('./utils');
 var CONSTANTS = require('../constants/constants');
-var createReactClass = require('create-react-class');
 var PropTypes = require('prop-types');
 
 /**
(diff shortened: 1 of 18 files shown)


js/components/controlButton.js
```diff
@@ -29,27 +28,36 @@ class NonAccessibleButton extends React.Component {
       </a>
     );
   };
-};
+}
```
pilievOoyala: Shouldn't we keep that semi colon?

js/components/controlButton.js
```diff
@@ -201,49 +209,51 @@ var ControlButton = createReactClass({
...
+    const className = classNames('oo-control-bar-item', this.props.className);
+    const iconStyles = this.getIconStyles();
+    const areTooltipsEnabled = this.areTooltipsEnabled();
 
-  render: function() {
-    var Component = this.getButtonComponent();
-    var className = classNames('oo-control-bar-item', this.props.className);
-    var iconStyles = this.getIconStyles();
-    var areTooltipsEnabled = this.areTooltipsEnabled();
+    let responsiveUiMultiplier, tooltipVerticalOffset;
```
pilievOoyala: Can you put these are on two separate lines?

js/components/tooltip.js
```diff
@@ -4,6 +4,7 @@ var deepmerge = require('deepmerge');
 var Utils = require('./utils');
 var createReactClass = require('create-react-class');
 var PropTypes = require('prop-types');
+var CONSTANTS = require('../constants/constants');
```
pilievOoyala: Do you want to update these to 'const' as well?

js/components/controlButton.js
jpfranco: Not in this case, class declarations don't end with semicolons

js/components/controlButton.js
jpfranco: Will update

pilievOoyala:
LGTM, just some minor stuff if you choose to fix

js/components/tooltip.js
jpfranco: I just updated the new files. We'll have to migrate the older ones gradually, but it's ok to keep them like this for now.

tests/components/skipControls-test.js
```diff
@@ -0,0 +1,340 @@
...
+    renderComponent();
+    expect(wrapper.find('.oo-skip-controls.oo-inactive').length).toBe(1);
+  });
+
+  it('should set pointer events to none on buttons when component is inactive', function() {
+    enableAllButtons();
+    props.isInactive = true;
+    renderComponent();
+    wrapper.find(ControlButton).forEach((button) => {
+      expect(button.props().style).toEqual({ pointerEvents: 'none' });
```
aeng7: Should we do a contains check instead of equality in case this style prop definition gets expanded?

tests/components/skipControls-test.js
```diff
@@ -0,0 +1,340 @@
...
+    props.controller.state.duration = 0;
+    renderComponent();
+    expect(wrapper.find(ControlButton).length).toBe(2);
+    expect(wrapper.find('.oo-previous-video').hostNodes().length).toBe(1);
+    expect(wrapper.find('.oo-skip-backward').hostNodes().length).toBe(0);
+    expect(wrapper.find('.oo-skip-forward').hostNodes().length).toBe(0);
+    expect(wrapper.find('.oo-next-video').hostNodes().length).toBe(1);
+  });
+
+  it('should NOT render Previous/Next video buttons when video is single', function() {
```
aeng7: What happens if the video is married?

aeng7:
LGTM other than one minor comment about a unit test

tests/components/skipControls-test.js
jpfranco: lol, we have a separate story for that

tests/components/skipControls-test.js
jpfranco: Updated

jpfranco:
@aeng7 updated the PR

PR #1100, opened 2018-07-10
Player 3519

pilievOoyala:
There is no unit test because of errors when trying to render the DOM in the unit test environment.

### DIFF
--- a/tests/html5skin-test.js
+++ b/tests/html5skin-test.js
@@ -1186,6 +1186,7 @@ describe('Controller', function() {
     var spyPublish;
 
     beforeEach(function() {
+      controller.state.elementId = elementId;
       spyPublish = sinon.spy(OO.mb, 'publish');
     });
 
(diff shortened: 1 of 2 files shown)


aeng7:
@pilievOoyala what errors are you observing when attempting to create unit tests? Our components should render fine in the jsdom included in our test environment.

jpfranco:
LGTM apart from the comment. I think the `loadConfigData` calls other functions that do some jquery stuff, so it might be a hassle to test it. It might be possible to mock some of it though

pilievOoyala:
I'm getting an errror from the render function.

Controller › Config settings › test that the chosen ui language is sent on the message bus

    Invariant Violation: Target container is not a DOM element.

      1507 | 
      1508 |       // load player
    > 1509 |       this.skin = ReactDOM.render(
           |                            ^
      1510 |         React.createElement(Skin, {
      1511 |           skinConfig: this.state.config,
      1512 |           localizableStrings: Localization.languageFiles,

      at invariant (node_modules/fbjs/lib/invariant.js:42:15)
      at legacyRenderSubtreeIntoContainer (node_modules/react-dom/cjs/react-dom.development.js:17238:34)
      at Object.render (node_modules/react-dom/cjs/react-dom.development.js:17317:12)
      at Html5Skin.loadConfigData (js/controller.js:1509:28)
      at Object.<anonymous> (tests/html5skin-test.js:129:18)

And here is what my test looks like. Maybe I set it up wrong.

describe('Config settings', function() {
    it('test that the chosen ui language is sent on the message bus', function() {
      var spy = sinon.spy(controller.mb, 'publish');
      controller.loadConfigData('customerUi', {"skin":{"inline":{"localization":{"defaultLanguage":""}}}}, {}, {}, {});
      console.log(spy.args);
      //expect(mb.published(asdf))
    });
  });

pilievOoyala:
Unit tests update after @aeng7 showed me how to get them to work.

tests/html5skin-test.js
```diff
@@ -1180,4 +1181,28 @@ describe('Controller', function() {
...
+    beforeEach(function() {
+      spyPublish = sinon.spy(OO.mb, 'publish');
+    });
+
+    afterEach(function() {
+      spyPublish.restore()
+    });
+
+    it('test that the chosen ui language is sent on the message bus', function() {
+      controller.state.elementId = elementId;
```
aeng7: Can we add this to the beforeEach block? This way if someone else adds new unit tests, they are less likely to face the same issue you did.

aeng7:
LGTM

PR #1102, opened 2018-07-10
[bugfix/PLAYER-3847] Not able to change audio track after failover to subchannel2

KseniiaSokolovaEpam:
Now we have a situation for HA video: when second subchannel starts playing function '_removeFocusFromElement' is called; videoId is 'reload' now. After that function 'vcFocusVideoElement' is called, and videoId is becoming 'main', and it is last time when Object { main: {…}, reload: {…} } exists. After that the object is { main: {…} }, but in skin state 'currentVideoId' (which we use for setting audio) is still 'reload'. And when user click on a track from MA list we return from the function because Object['reload'] does not exist.

So, now we change value of currentVideoId in skin to value of videoId in mjolnir.

In the test we check if the value of currentVideoId is correct.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -1388,6 +1388,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
         this.state.pluginsElement.removeClass('oo-showing');
         this.state.pluginsClickElement.removeClass('oo-showing');
       }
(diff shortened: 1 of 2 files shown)


aeng7:
LGTM

pilievOoyala:
LGTM

PR #1103, opened 2018-07-17
one touch pause fix

knooyala:
Before: single tap on screen provides a pause only before video was rotated; after that pause is not works.
Now: play/pause function works in any time.

Tests: already exists in playingScreen-test.js ('creates a PlayingScreen and checks touchEnd'); another tests could not be created because of all changes are provides by internal variable in function.


js/skin.js
```diff
@@ -136,13 +136,16 @@ var Skin = createReactClass({
...
     if (this.props.controller && this.props.controller.isVrStereo) {
       return;
     }
+
     if (this.props.controller && this.props.controller.videoVr) {
-      var isVrMouseMove = this.state.isVrMouseMove;
-      if (Utils.isIos()) {
+      var isVrMouseMove = this.state.isVrMouseMove,
+          isTouchEnd = typeof arguments[0] === 'object' && arguments[0].type === 'touchend';
+
```
KseniiaSokolovaEpam: Could you add "var" for second variable instead of list of variable (according by our rules in eslint)?

KseniiaSokolovaEpam:
Could you remove npm-debul.log file?

js/skin.js
KseniiaSokolovaEpam: Could we use only isTouchEnd instead of Utils.isIos()?

KseniiaSokolovaEpam: Could you add tests for the changes?

js/skin.js
```diff
@@ -136,13 +136,16 @@ var Skin = createReactClass({
    * @public
    * @description the function is called when we stop the rotation
    */
-  handleVrPlayerMouseUp: function() {
+  handleVrPlayerMouseUp: function() {
```
aeng7: Extra added whitespace

aeng7:
LGTM other than @KseniiaSokolovaEpam's comments

js/skin.js
```diff
@@ -136,13 +136,16 @@ var Skin = createReactClass({
...
+  handleVrPlayerMouseUp: function() {    
     if (this.props.controller && this.props.controller.isVrStereo) {
       return;
     }
+
     if (this.props.controller && this.props.controller.videoVr) {
-      var isVrMouseMove = this.state.isVrMouseMove;
-      if (Utils.isIos()) {
+      var isVrMouseMove = this.state.isVrMouseMove,
+          isTouchEnd = typeof arguments[0] === 'object' && arguments[0].type === 'touchend';
```
jpfranco: Can we add the event parameter to the function instead reading `arguments[0]`? This will make it easier to understand which parameter we're working with

knooyala: +

js/skin.js
knooyala: \+

js/skin.js
knooyala: It's risky to use only isTouchEnd instead of Utils.isIos(). Something can broke.
Tests already exists.

aeng7:
@knooyala, please add some unit tests. I see your comment on the tests in the top level comment, but we can test for the state change that is occurring (isVrMouseMove) in the situation this PR is attempting to address.

tests/views/playingScreen-test.js
```diff
@@ -204,6 +204,50 @@ describe('PlayingScreen', function() {
     expect(clicked).toBe(true);
   });
 
+////********
+  it('creates a PlayingScreen, move video and checks touchEnd', function() {
```
aeng7: From an initial glance, this test does not appear to test the changes made in this PR. It looks like we should test that if ```handleVrPlayerMouseUp``` is called via a touchEnd event, we want to set the ```isVrMouseMove``` state to false. A state change is significant enough to add a unit test for.

aeng7:
@knooyala please fix the conflicts for this PR. Thanks!

tests/views/playingScreen-test.js
knooyala: Unfortunately, I can't simulate touchMove event. It should to work by the logic, but it doesn't work in tests. May be, a do something wrong, may be it's just an bug or restriction of JEST.
I found same problems on stackoverflow but without resolving. 
So, that is the best testing way I found.

knooyala: Unfortunately, I can't simulate touchMove event. It should to work by the logic, but it doesn't work in tests. May be, a do something wrong, may be it's just an bug or restriction of JEST.
I found same problems on stackoverflow but without resolving. 
So, that is the best testing way I found.

aeng7: Ah, can you add a TODO on this then? Otherwise sounds good

aeng7:
LGTM

pilievOoyala:
LGTM

PR #1106, opened 2018-07-18
Player 3808c

aeng7:
Created withAutoHide HOC to handle screen auto hiding functionality. Applied this to the Pause Screen.


js/controller.js
```diff
@@ -2505,13 +2506,15 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
 
     showControlBar: function() {
       this.state.controlBarVisible = true;
+      this.renderSkin();
```
jpfranco: Do we need a separate `showControlBarAndRender` in order to avoid rendering in other cases where the code doesn't need it?

js/skin.js
```diff
@@ -394,6 +395,7 @@ var Skin = createReactClass({
               seeking={this.state.seeking}
               upNextInfo={this.state.upNextInfo}
               isLiveStream={this.state.isLiveStream}
+              controlBarAutoHide={this.props.skinConfig.controlBar.autoHide}
```
jpfranco: Not super important, but I think it might be better to read this directly from skin config rather than have a separate prop. I've struggled with some unit tests that use props that come from two different sources

js/views/higher-order/withAutoHide.js
```diff
@@ -0,0 +1,194 @@
+const React = require('react');
+const ReactDOM = require('react-dom');
+const CONSTANTS = require('../../constants/constants');
+
+const withAutoHide = function(ComposedComponent) {
```
jpfranco: Can we add a jsdoc for this component?

js/views/higher-order/withAutoHide.js
```diff
@@ -0,0 +1,194 @@
...
+    /**
+     * Shows the control bar if not mobile or if mobile and was triggered
+     * by a touchend event.
+     * @public
+     * @param event The event object from the event that triggered this
+     */
+    showControlBar(event) {
+      if (!this.props.controller.state.isMobile || (event && event.type === 'touchend')) {
+        this.props.controller.showControlBar();
+        ReactDOM.findDOMNode(this.refs.AutoHideScreen).style.cursor = 'auto';
```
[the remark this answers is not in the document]
aeng7: Will do

js/views/higher-order/withAutoHide.js
```diff
@@ -0,0 +1,194 @@
...
+      this.handleKeyDown = this.handleKeyDown.bind(this);
+      this.startHideControlBarTimer = this.startHideControlBarTimer.bind(this);
+      this.cancelHideControlBarTimer = this.cancelHideControlBarTimer.bind(this);
+      this.handleMouseOut = this.handleMouseOut.bind(this);
+      this.handlePlayerMouseMove = this.handlePlayerMouseMove.bind(this);
+      this.handleTouchEnd = this.handleTouchEnd.bind(this);
+    }
+
+    componentDidMount() {
+      document.addEventListener('mousemove', this.handlePlayerMouseMove, false);
```
jpfranco: Is it possible to limit these to the screen's container, or are they needed at the document level?

js/controller.js
jpfranco: The `renderSkin` function ends up calling `setState` on skin.js. I agree that we probably do want to render though, I'm not sure how the code that called this without rendering was working. It might be worth checking whether the value actually changed though

aeng7: I'll add in a check to see if the value changed

js/views/higher-order/withAutoHide.js
aeng7: Will add

js/views/higher-order/withAutoHide.js
aeng7: I copied these from its original location in playingScreen.js:

https://github.com/ooyala/html5-skin/blob/master/js/views/playingScreen.js#L47

I'd like to avoid changing the behavior unless necessary.

js/controller.js
jpfranco: @pilievOoyala it still calls all the render methods and has to do some work in order to determine what has changed. Due to the way our player is structured, a render at this level probably ends up executing the render method of all active components.

scss/components/_state-screen.scss
```diff
@@ -80,6 +81,13 @@
     top: 5%;
     left: 5%;
 
+    transition: 0.2s opacity ease-in-out, 0.3s transform ease-in-out;
```
jpfranco: Sorry I missed this the last time. Is it deliberate to use a different duration for the transform transition?

js/views/playingScreen.js
```diff
@@ -446,11 +391,8 @@ var PlayingScreen = createReactClass({
     return (
       <div
         className={className}
-        ref="PlayingScreen"
-        onTouchStart={this.handleTouchStart}
```
jpfranco: I think this `onTouchStart` shouldn't be removed. There's a handler that sets the `hasCheckedMouseOverControls` var

jpfranco:
@aeng7 one last question, does this feature need to be configurable via skin.json?

scss/components/_state-screen.scss
aeng7: Ah I copied this from the skip controls css to keep things consistent:

https://github.com/ooyala/html5-skin/blob/master/scss/components/_skip-controls.scss#L8

Should I update both with 0.3s for the opacity transform? It seems like 0.3s is the go to transition/transform time elsewhere in our skin

jpfranco: Oh ok sorry, I forgot I had done that, but it was deliberate. If possible it might be good to move this to a variable since they now both use the same transition, but  not super important

js/views/playingScreen.js
aeng7: Fixed

scss/components/_state-screen.scss
aeng7: I'll add this in

js/views/higher-order/withAutoHide.js
```diff
@@ -0,0 +1,201 @@
...
+          break;
+        default:
+          break;
+      }
+    }
+
+    /**
+     * Handles the mouseout event.
+     * @private
+     * @param event The mouseout event object
```
pilievOoyala: Your param labeling is inconsistent. I think we need the {Event} on all these.  You just have the var name and desc.

pilievOoyala:
LGTM, just minor jsdoc thing. You can merge and fix that later or just fix it and merge yourself.

aeng7:
@pilievOoyala I'll update the jsdocs later.

PR #1107, opened 2018-07-19
[PLAYER-3645] Switching video quality panel to new component/styles

jpfranco:
The video quality panel will now use the same component as the playback speed panel in order for them to have the same behavior and styles.

Most of the CSS classes used for the video quality panel were replaced with more generic equivalents and are no longer needed, but I kept the actual classes and html structure for backward compatibility. The only exception to this is the "Auto Quality" button, which no longer looks different from the other options and thus doesn't need a special class. 

**Note:** I will include unit tests in a future PR


aeng7:
LGTM

pilievOoyala:
LGTM

PR #1108, opened 2018-07-20
Feat/player 3621 change param value for MultiAudio

KseniiaSokolovaEpam:
Value of use Roving Tabindex for Multi Audio Component was changed for enable possibility to use TAB.

Unit tests are not necessary.


aeng7:
LGTM

PR #1109, opened 2018-07-20
[PLAYER-3645] Allowing UI to react to API speed changes

jpfranco:
These updates allow handling the case in which playback speed is set via API to a value that is not one of the predefined values. The UI will now update and include the new value in the list.

Note: Unit tests are still pending, will add in a future PR.


dallen42:
@jp do we need any unit tests for this?

jpfranco:
@dallen42 oh sorry, I forgot to mention. The unit tests for the whole playback speed feature are still pending. I will add these in a future PR

aeng7:
LGTM

PR #1111, opened 2018-07-31
Player 4041

aeng7:
-fixed an issue where an ad resumes playback immediately after a clickthrough
-cleaned up usage of plugins elements in html5skin-test.js

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -1644,6 +1644,9 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
         }.bind(this)
       );
 
+      //PLAYER-4041: It seems like 'click' events started when a div's pointerEvents is initially 'none' but is changed to
+      //'auto' prior to the 'click' event ending will trigger 'click' event listeners on Android. We'll instead listen to
+      //'touchend' and 'touchcancel' on Android.
       if (OO.isAndroid) {
         this.state.pluginsClickElement.on('touchend touchcancel', this.resumePlaybackAfterClickthrough.bind(this));
       } else {
--- a/js/controller.js
+++ b/js/controller.js
@@ -1643,12 +1643,13 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
           this.hideControlBar();
         }.bind(this)
       );
-      this.state.pluginsClickElement.click(
-        function() {
-          this.state.pluginsClickElement.removeClass('oo-showing');
-          this.mb.publish(OO.EVENTS.PLAY);
-        }.bind(this)
-      );
+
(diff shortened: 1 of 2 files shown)


js/controller.js
```diff
@@ -1643,12 +1643,16 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
-          this.state.pluginsClickElement.removeClass('oo-showing');
-          this.mb.publish(OO.EVENTS.PLAY);
-        }.bind(this)
-      );
+
+      //PLAYER-4041: It seems like 'click' events started when a div's pointerEvents is initially 'none' but is changed to
+      //'auto' prior to the 'click' event ending will trigger 'click' event listeners on Android. We'll instead listen to
+      //'touchend' and 'touchcancel' on Android.
+      if (OO.isAndroid) {
+        this.state.pluginsClickElement.on('touchend touchcancel', this.resumePlaybackAfterClickthrough.bind(this));
```
knooyala: Should touchcancel handle this operation? May be user changed his mind and don't want to perform action?

aeng7: I think the touchcancel event implementation is browser specific:

https://developer.mozilla.org/en-US/docs/Web/Events/touchcancel

Including touchcancel was more of a fallback in case the end user triggers the event instead of touchend. If triggered, the video will just unpause.

tests/html5skin-test.js
```diff
@@ -126,6 +111,7 @@ describe('Controller', function() {
 
   describe('Closed Captions', function() {
     beforeEach(function() {
+      controller.createPluginElements();
```
jpfranco: Not super important, but I think it might be simpler if we added this to the global `beforeEach` at the controller level instead of to each individual test group

jpfranco:
Oops, Peter was already revewing

tests/html5skin-test.js
aeng7: There are some tests that require ```createPluginElements``` to be called after some setup, like https://github.com/ooyala/html5-skin/pull/1111/files#diff-63a4cbb7eb24c1dc5f84f0a7439d0d34R1245 the 'Show player controls over ads' tests.

jpfranco: Ok, got it

js/controller.js
aeng7: @knooyala I'd like to get this merged so QA can test it. Let me know what you think, thanks!

aeng7: @knooyala we're going to merge this to get QA started. Let me know if you have any concerns and we'll continue discussion

PR #1112, opened 2018-07-31
Bugfix/player 3750 - restore original position of video 360 after ads played

knooyala:
before: after ads plaing video orientation resets to initial position
after: controller save position and restore it after ads was played
test: added

+ merged with master branch => some other code lines was affected


js/controller.js
```diff
@@ -1299,6 +1324,24 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
       this.renderSkin();
+
+      // restore video360 pos by deviceOrientation changes (only for mobile devices)
+      // It's uses after video 360 was moved by device motion, after ads playback
+      if(this.state.deviceOrientation.allow) {
+        var orientationType = Utils.getOrientationType();
+        var params = [0, 0, 0];
+
+        params[0] = this.state.deviceOrientation.alpha;
+        params[2] = orientationType === 'landscape-primary' ?
```
KseniiaSokolovaEpam: What happens if the orientation is "landscape-secondary"?

js/controller.js
```diff
@@ -2389,7 +2431,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
 
       // validate language is available before update and save
       if (language && availableLanguages && _.contains(availableLanguages.languages, language)) {
-        this.state.closedCaptionOptions.language =
+        this.state.closedCaptionOptions.language =
```
KseniiaSokolovaEpam: Remove extra space

knooyala: \+

js/controller.js
knooyala: it's another position; in this case axises are changes by places

js/controller.js
```diff
@@ -498,6 +503,26 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
         }
         this.checkDeviceOrientation = false;
       }
+
+      // for Ads
+       
+      // mobile device detected
+      this.state.deviceOrientation.allow = true;
+      // current position of device
+      if(!this.state.deviceOrientation.freeze) {
```
jpfranco: Can we add spaces between `if(`? Same comment for occurrences below.

js/controller.js
```diff
@@ -1299,6 +1324,24 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       // In case ad was skipped or errored while stalled
       this.setBufferingState(false);
       this.renderSkin();
+
+      // restore video360 pos by deviceOrientation changes (only for mobile devices)
+      // It's uses after video 360 was moved by device motion, after ads playback
+      if(this.state.deviceOrientation.allow) {
```
jpfranco: Could we move this to a separate function? Something like `restoreDeviceOrientation()`. There's a lot of logic in this handler already, so it's a bit hard to reason about the flow

js/controller.js
jpfranco: Can we add this value to constants.js?

js/controller.js
```diff
@@ -1299,6 +1324,24 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
+        var params = [0, 0, 0];
+
+        params[0] = this.state.deviceOrientation.alpha;
+        params[2] = orientationType === 'landscape-primary' ? 
+          this.state.deviceOrientation.beta : 
+          this.state.deviceOrientation.gamma;
+
+        // To apply there need a time gap
+        setTimeout(function() {
+          this.onTouchMove(params);
```
jpfranco: Could we separate this logic from the `touchmove` event? In general I think it's better not to call event handlers manually since some other part of the code might be expecting these to be triggered by actual DOM event

tests/html5skin-test.js
```diff
@@ -307,6 +307,31 @@ describe('Controller', function() {
       expect(stopBufferingTimerSpy.callCount).toBe(8);
       expect(controller.state.bufferingTimer).toBeFalsy();
     });
+
+    it('should ads played, device was moved - video should continue to play in last orientation', function() {
```
jpfranco: Can we move this to a more appropriate section? This seems to be on the `Buffering state` tests, which are unrelated

pilievOoyala:
LGTM aside from the other questions

jpfranco:
Also forgot to mention, I think html5-common was removed in a recent PR. Was it re-added by accident or is it needed again?

js/controller.js

PR #1113, opened 2018-07-31
[bugfix/PLAYER-4018] langauge code 'und' is translated to 'qafaraf' instead of 'undefined langauge'

KseniiaSokolovaEpam:
If we get langCode === '' we should not use it to find name in the language map ; we need to set displayLanguage to ''.

[Origin task](https://jira.corp.ooyala.com:8443/browse/PLAYER-4018)

Unit tests are passed.


pilievOoyala:
LGTM

PR #1116, opened 2018-08-03
[PLAYER-3645] Highlighting Playback Speed button when menu is active

jpfranco:
I had missed this little detail. The control bar buttons get the highlight color when their respective popover menu is open.

### DIFF
--- a/js/components/controlBar.js
+++ b/js/components/controlBar.js
@@ -401,19 +401,20 @@ var ControlBar = createReactClass({
 
     var qualityClass = ClassNames({
       'oo-quality': true,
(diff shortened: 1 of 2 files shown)


pilievOoyala:
LGTM

js/components/controlBar.js
```diff
@@ -401,19 +401,20 @@ var ControlBar = createReactClass({
 
     var qualityClass = ClassNames({
       'oo-quality': true,
-      'oo-control-bar-item': true,
```
jpfranco: `ControlButton`'s have these classes by default, so these were getting duplicated

PR #1117, opened 2018-08-03
[player-4080]

aeng7:
-will set multiAudio state to null if we receive less than 2 tracks (effectively hiding the multi audio icon)

PLAYER-4080 was occurring because the video element is throwing an audio change event on a single audio track, which sends out a multi audio changed event with only 1 audio track. This makes our skin show the multi audio icon. This looks to be a native Safari issue. To workaround this, I've modified the skin to set the multi audio state to null if we receive a multi audio changed event with less than 2 tracks.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -1097,7 +1097,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
         // if param hideMultiAudioIcon is set to false
         if (typeof multiAudio !== 'undefined') {
           let multiAudioValue = null;
-          if (multiAudio.tracks) {
+          if (this.containsMultiAudio(multiAudio)) {
             multiAudioValue = { tracks: multiAudio.tracks };
           }
           this.state.multiAudio = multiAudioValue;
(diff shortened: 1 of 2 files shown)


js/controller.js
```diff
@@ -1116,11 +1116,25 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
       }
     },
 
+    /**
+     * Checks to see if we have multiple audio tracks
+     * @param multiAudio The multiAudio object that is fetched from the current video
+     * @returns {boolean} True if the provided object contains more than one track, false otherwise
+     */
+    containsMultiAudio: function(multiAudio) {
+      return !!(multiAudio && multiAudio.tracks && multiAudio.tracks.length > 1);
```
pilievOoyala: Can you stick the result in a temp var and return that?

js/controller.js
```diff
@@ -1116,11 +1116,25 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
+          this.state.multiAudio = null;
+        }
+
         this.renderSkin();
       }
     },
 
+    /**
+     * Checks to see if we have multiple audio tracks
+     * @param multiAudio The multiAudio object that is fetched from the current video
```
pilievOoyala: Can you put the type?

js/controller.js
aeng7: Added

js/controller.js
aeng7: Added

pilievOoyala:
LGTM

PR #1119, opened 2018-08-08
bf/PLAYER-3750 View direction is reset to initial position after AD for tilting

KseniiaSokolovaEpam:
When an user use only tilting we do not have correct value for viewing direction. So we forcibly check the viewing direction when AD starts and set this value when AD ends.

(And some changes for eslint.)

Unit tests are passed.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -475,16 +475,16 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
      * Uses for video 360 on mobile devices for setting necessary coordinates (relevant with start device orientation)
      * Before playing this.vrMobileOrientationChecked is equal false,
      * if need to check value for device orientation set this.checkDeviceOrientation = true
-     * @param {object} e The event object
+     * @param {object} event The event object
      */
-    handleVrMobileOrientation: function(e) {
+    handleVrMobileOrientation: function(event) {
       if (!this.vrMobileOrientationChecked || this.checkDeviceOrientation) {
-        var beta = e.beta;
-        var gamma = e.gamma;
-        var yaw = this.state.vrViewingDirection['yaw'];
-        var pitch = this.state.vrViewingDirection['pitch'];
-        var dir = beta;
-        var orientationType = Utils.getOrientationType();
+        let beta = event.beta;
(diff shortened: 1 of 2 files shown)


js/controller.js
```diff
@@ -1313,6 +1319,15 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       this.trySetAnamorphicFixState(true);
       // In case ad was skipped or errored while stalled
       this.setBufferingState(false);
+
```
dallen42: @KseniiaSokolovaEpam can we remove extra line here?

js/controller.js
```diff
@@ -475,25 +475,26 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
-    handleVrMobileOrientation: function(e) {
+    handleVrMobileOrientation: function(event) {
       if (!this.vrMobileOrientationChecked || this.checkDeviceOrientation) {
-        var beta = e.beta;
-        var gamma = e.gamma;
-        var yaw = this.state.vrViewingDirection['yaw'];
-        var pitch = this.state.vrViewingDirection['pitch'];
-        var dir = beta;
-        var orientationType = Utils.getOrientationType();
+        let beta = event.beta;
```
jpfranco: We don't need to change this, but I think it's a good convention to use `const` for any values that are not re-assigned and `let` for values that change. This helps with reasoning about which variables have changed

js/controller.js
```diff
@@ -1313,6 +1319,15 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       this.trySetAnamorphicFixState(true);
       // In case ad was skipped or errored while stalled
       this.setBufferingState(false);
+
+
+      var vrViewingDirectionList = [
```
jpfranco: Could we add a more descriptive comment here? For someone without context it's hard to understand what this is meant to do. It might also make sense to move this to own its own function with a descriptive name

pilievOoyala:
LGTM.

js/controller.js

js/controller.js

js/controller.js

PR #1121, opened 2018-08-17
[update-readme]

aeng7:
-Updated readme with simpler examples by removing Discovery, Freewheel, and OSMF plugins


pilievOoyala:
LGTM

sample.html
```diff
@@ -19,8 +17,7 @@
     "playerBrandingId": "26e2e3c1049c4e70ae08a242638b5c40",
     'debug':true,
     "skin": {
-      "config": "",
```
pilievOoyala: Why are we removing this?

pilievOoyala: Do we not need the iframe example?

pilievOoyala:
@aeng7  You can merge, I just have 1 question.

sample.html
aeng7: @aldigun said to remove it. I'm assuming it's because it is outside the scope of a simple example.

PR #1122, opened 2018-08-23
[bugfix/PLAYER-4243] Clicking on video in playlist shows the ad controls on main video

KseniiaSokolovaEpam:
If click on another video from a playlist during an ads is playing an user will see AD_SCREEN. It happens because of we check state "isPlayingAd" (controller.js, line 906) and if it is true, scrrenToShow is set to AD_SCREEN.
So we have to reset the value for isPlayingAd and for case when ads is not ended yet but an user switch it.

[Original task](https://jira.corp.ooyala.com/browse/PLAYER-4243)

Unit tests are passed.


tests/html5skin-test.js
```diff
@@ -346,6 +346,13 @@ describe('Controller', function() {
       expect(controller.state.screenToShow).toBe(CONSTANTS.SCREEN.LOADING_SCREEN);
     });
 
+    it('should reset isPlayingAd state when embed code is set after video has started', function() {
+      controller.state.screenToShow = CONSTANTS.SCREEN.INITIAL_SCREEN;
+      controller.state.initialPlayHasOccurred = true;
+      controller.onSetEmbedCode('newEmbedCode');
+      expect(controller.state.isPlayingAd).toBe(false);
```
jpfranco: Looks good, I just think we should set to `controller.state.isPlayingAd` to `true` at the start of the test, otherwise we might not be testing that the value is being changed since it's `false` by default.

pilievOoyala:
LGTM aside from JP's comment.

tests/html5skin-test.js
KseniiaSokolovaEpam: oh, thank you!!

\+

PR #1123, opened 2018-08-27
bugfix/PLAYER-4220 View direction is taken back by a tap after tilting

KseniiaSokolovaEpam:
Plugin "eslint-plugin-require-jsdoc-except" was added to exept jsdoc validation for constructor, componentWilMount etc.

PauseScreen, playingScreen were updated according to eslint rules.

common function "handleTouchEnd" for playingScreen and pauseScreen was moved to skin.js and changed to fix the [bug](https://jira.corp.ooyala.com:8443/browse/PLAYER-4220) .

Unit tests were changed and added. Unit tests passed.


js/controller.js
```diff
@@ -507,6 +507,18 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       this.vrMobileOrientationChecked = false;
     },
 
+    /**
+     * Pass into onTouchMove current controller state values for vrViewingDirection
+     */
+    setControllerVrViewingDiraction: function() {
```
zelibobla: Not diraction but direction I guess

js/skin.js
```diff
@@ -128,8 +128,8 @@ var Skin = createReactClass({
         isVrMouseMove: true
       });
       if (typeof this.props.controller.onTouchMove === 'function') {
-        var coords = Utils.getCoords(event);
-        var params = this.getDirectionParams(coords.x, coords.y);
+        let coords = Utils.getCoords(event);
+        let params = this.getDirectionParams(coords.x, coords.y);
```
zelibobla: coords and params are const inside this block as well

js/skin.js
```diff
@@ -251,12 +251,12 @@ var Skin = createReactClass({
...
-    var selector = '[' + CONSTANTS.KEYBD_FOCUS_ID_ATTR + ']:enabled';
-    var focusableElements = this.domElement.querySelectorAll(selector);
+    const selector = '[' + CONSTANTS.KEYBD_FOCUS_ID_ATTR + ']:enabled';
+    let focusableElements = this.domElement.querySelectorAll(selector);
 
     if (focusableElements.length) {
-      var firstFocusableElement = focusableElements[0];
-      var lastFocusableElement = focusableElements[focusableElements.length - 1];
+      let firstFocusableElement = focusableElements[0];
+      let lastFocusableElement = focusableElements[focusableElements.length - 1];
```
zelibobla: Both elements provided by reference. It is not going to change down by code, so it's a const.

js/skin.js
```diff
@@ -287,8 +287,42 @@ var Skin = createReactClass({
     }
   },
 
+  /**
+   * call handleTouchEnd when touchend was called on selectedScreen
+   * @public
+   * @param {Event} event - event object
+   */
+  handleTouchEnd: function(event) {
+    event.preventDefault(); // to prevent mobile from propagating click to discovery shown on pause
```
zelibobla: Actually it won't stop propagation, but stop default browser listeners to hear event exactly on touched element (not parent). I'm not aware yet about the reason of this method in general, but if you wanted to stop propagation you need event.stopPropagation() method. Is .preventDefault() is exactly what you wanted then probably we need another comment for this line.

js/skin.js
```diff
@@ -287,8 +287,42 @@ var Skin = createReactClass({
     }
   },
 
+  /**
+   * call handleTouchEnd when touchend was called on selectedScreen
```
zelibobla: It doesn't explain general purpose of the method. I suggest: "Manage controls visibility and switch active screen if required, once the screen touched."
Notice how much now we would be happy to decompose this method on two:
- handleTouchEnd
- checkVrDirection (this is already implemented in controller)

js/skin.js
```diff
@@ -287,8 +287,42 @@ var Skin = createReactClass({
...
+      if (shouldToggle) {
+        this.props.controller.togglePlayPause(event);
+      }
+      // set value for this.state.isVrMouseMove in this.handleVrPlayerMouseUp
+    }
+    if (this.props.controller.videoVr) { // only for vr on mobile
+      // Check current a vr video position (an user could change position using tilting)
+      if (
+        typeof this.props.controller.checkVrDirection === 'function' &&
+        typeof this.props.controller.setControllerVrViewingDiraction === 'function'
```
zelibobla: We are owners of the controller.js, so it's a bit unclear why do we check if we didn't forget to implement method of our controller? Consider we should trust ourselves.

js/views/pauseScreen.js
```diff
@@ -104,41 +108,20 @@ class PauseScreen extends React.Component {
...
-        shouldToggle = true;
-      }
-      if (shouldToggle) {
-        this.props.controller.togglePlayPause(e);
-      }
-    }
-  }
-
   /**
    * call handleVrTouchEnd when touchend was called on selectedScreen and videoType is Vr
```
zelibobla: Guess we don't need this method. Just bypass this.props.handleVrPlayerMouseUp to the 52 and 61 line where you use it. If required binding, arrow functions are much shorter way to get it.

js/views/pauseScreen.js
```diff
@@ -104,41 +108,20 @@ class PauseScreen extends React.Component {
...
-  handleVrTouchEnd(e) {
-    this.props.handleVrPlayerMouseUp(e);
+  handleVrTouchEnd(event) {
+    this.props.handleVrPlayerMouseUp(event);
   }
 
   /**
    * remove the button on pause screen for correct checking mouse movement
    */
   hideVrPauseButton() {
```
zelibobla: This method is a kind overcomplicated, and it runs setTimeout even if not required. Guess it should be:
if (!this.props.controller.videoVr || !this.pauseButton){
  return;
}
setTimeout(() => this.pauseButton.style.display = 'none', 1000);

js/views/pauseScreen.js
```diff
@@ -147,43 +130,50 @@ class PauseScreen extends React.Component {
...
   }
 
   /**
    * call handleVrMouseUp when mouseup was called on document
-   * @param {Event} e - event object
+   * @param {Event} event - event object
    */
-  handleVrMouseUp(e) {
-    this.props.handleVrPlayerMouseUp(e);
+  handleVrMouseUp(event) {
```
zelibobla: The same as for `handleVrTouchEnd`. Not required wrapper.

js/views/pauseScreen.js
```diff
@@ -246,18 +241,22 @@ class PauseScreen extends React.Component {
...
       <div className={titleClass} style={titleStyle}>
         {this.props.contentTree.title}
       </div>
     );
-    var descriptionMetadata = (
-      <div className={descriptionClass} ref="description" style={descriptionStyle}>
+    const descriptionMetadata = (
+      <div
+        className={descriptionClass}
+        ref={text => this.description = text}
```
zelibobla: Even simple logic inside templates is a kind of antipattern.

js/views/pauseScreen.js
```diff
@@ -318,7 +318,7 @@ class PauseScreen extends React.Component {
         </div>
 
         <button
-          id="oo-pause-button"
+          ref={btn => this.pauseButton = btn}
```
zelibobla: Logic inside templates is a kind of antipattern

js/views/playingScreen.js
```diff
@@ -119,55 +119,42 @@ class PlayingScreen extends React.Component {
   }
 
   /**
-   * call handleTouchEnd when touchend was called on selectedScreen
-   * @param {Event} event - event object
+   * call handleVrTouchEnd when touchend was called on document and videoType is Vr
```
zelibobla: Description should explain the purpose of the method body, not when it called.

js/views/playingScreen.js
```diff
@@ -258,42 +244,46 @@ class PlayingScreen extends React.Component {
...
    */
-  setAnimationDuration(vrDuration, defaultDuration) {
-    var style = {};
-    defaultDuration = Utils.ensureNumber(defaultDuration, 3);
+  setAnimationDuration(vrDuration, userDefaultDuration) {
+    let style = {};
+    const functionDefaultfDuration = 3; // default value for Duration if userDefaultDuration is undefined
+    const defaultDuration = Utils.ensureNumber(userDefaultDuration, functionDefaultfDuration);
     if (
       this.props.controller.state.config.animationDurations !== null &&
```
zelibobla: Hope `const userDefined = this.props.controller.state.config.animationDurations` would make your code here much easier to read.

js/views/playingScreen.js
```diff
@@ -364,27 +357,27 @@ class PlayingScreen extends React.Component {
       !this.state.isVrIconHidden &&
       this.props.controller.isNewVrVideo
     ) {
-      var defaultDuration = 3;
-      var style = this.setAnimationDuration('vrIcon', defaultDuration);
+      let defaultDuration = 3;
+      let style = this.setAnimationDuration('vrIcon', defaultDuration);
       vrIcon = (
-        <div id="vrIconContainer" className="oo-state-screen-vr-container" style={style}>
+        <div ref={icon => this.vrIconContainer = icon} className="oo-state-screen-vr-container" style={style}>
```
zelibobla: Logic inside templates

js/controller.js

js/skin.js
```diff
@@ -77,7 +77,7 @@ var Skin = createReactClass({
   },
 
   switchComponent: function(args) {
-    var newState = args || {};
+    let newState = args || {};
```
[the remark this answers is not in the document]

js/skin.js
```diff
@@ -100,7 +100,7 @@ var Skin = createReactClass({
       return;
     }
     if (this.props.controller && this.props.controller.videoVr) {
-      var coords = Utils.getCoords(event);
+      let coords = Utils.getCoords(event);
```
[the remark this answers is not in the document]

js/skin.js

js/skin.js

js/skin.js

js/skin.js

js/skin.js
KseniiaSokolovaEpam: we need to check it according by style docs.

js/views/pauseScreen.js
KseniiaSokolovaEpam: \+

js/views/pauseScreen.js
KseniiaSokolovaEpam: \+

js/views/pauseScreen.js
KseniiaSokolovaEpam: \+

js/views/pauseScreen.js
KseniiaSokolovaEpam: it would be better to save function for 'ref' now.

js/views/pauseScreen.js
KseniiaSokolovaEpam: it would be better to save function for 'ref' now.

js/views/playingScreen.js

js/views/playingScreen.js
KseniiaSokolovaEpam: the same as above

js/skin.js

js/skin.js

js/skin.js
KseniiaSokolovaEpam: :-)

js/views/pauseScreen.js
KseniiaSokolovaEpam: \+

js/views/pauseScreen.js
KseniiaSokolovaEpam: \+

js/views/pauseScreen.js
KseniiaSokolovaEpam: \+

js/skin.js
```diff
@@ -200,16 +200,16 @@ var Skin = createReactClass({
...
-    var pageY = Utils.ensureNumber(_pageY, 0);
-    var dx = pageX - this.state.xVrMouseStart;
-    var dy = pageY - this.state.yVrMouseStart;
-    var maxDegreesX = 90;
-    var maxDegreesY = 120;
-    var degreesForPixelYaw = maxDegreesX / this.state.componentWidth;
-    var degreesForPixelPitch = maxDegreesY / this.state.componentHeight;
-    var yaw = this.getVrViewingDirectionParamValue('yaw') + dx * degreesForPixelYaw;
-    var pitch = this.getVrViewingDirectionParamValue('pitch') + dy * degreesForPixelPitch;
+    let pageX = Utils.ensureNumber(_pageX, 0);
```
jpfranco: Not super important, but it looks like none of these are re-assigned, so they could as well be all `const`

js/views/pauseScreen.js
```diff
@@ -104,86 +107,60 @@ class PauseScreen extends React.Component {
...
-      var pauseButton = document.getElementById('oo-pause-button');
-      setTimeout(function() {
-        if (pauseButton) {
-          pauseButton.style.display = 'none';
-        }
-      }, 1000);
+    if (!(this.props.controller.videoVr && this.pauseButton)) {
+      return;
     }
+    setTimeout(() => this.pauseButton.style.display = 'none', 1000);
```
jpfranco: Is there a reason why we modify the element's style directly instead of using `setState()`?

js/views/pauseScreen.js
jpfranco: Was this updated?

js/views/pauseScreen.js
jpfranco: I think this is an improvement over the deprecated string ref that we had before. In the future we should probably just use `createRef()` and then declare the ref such as `<div ref={this.description} >`

js/skin.js

js/skin.js
jpfranco: I agree with @zelibobla , I think for our own methods we probably only need to check in cases in which it's an optional prop that might've not been passed.

js/views/pauseScreen.js
KseniiaSokolovaEpam: it is possible to use setState, but the idea was to use the same time as in css, if we use setState we will have a delay

js/views/pauseScreen.js
KseniiaSokolovaEpam: oh, I missed out it, thank you!

\+

js/views/pauseScreen.js
KseniiaSokolovaEpam: I tried to use it, but had a problem (this.description was returned as undefined), and I decided to check it in future because it did not connect with the task.

.eslintrc
```diff
@@ -70,14 +70,22 @@
...
       "error",
       {
         "require": {
           "MethodDefinition": true,
           "ClassDeclaration": true,
           "FunctionDeclaration": true
-        }
+        },
+        "ignore": [
+          "constructor",
```
pilievOoyala: Do we have another place where we can document the params that you pass into a constructor?

pilievOoyala: The description for the constructor can just be "Constructor" but I think the params is the important part. Especially if it's overloaded.

js/skin.js
```diff
@@ -287,8 +287,44 @@ var Skin = createReactClass({
     }
   },
 
+  /**
+   * This function changes a screen (pause/play) and
```
pilievOoyala: Change the wording a little bit:
This function toggles the current screen (pause/play) and

js/skin.js
```diff
@@ -287,8 +287,44 @@ var Skin = createReactClass({
...
+   * sets current direction for vr video (it is necessary for tilting)
+   * when touchend was called on selected screen
```
pilievOoyala: change 'was' to 'is'

js/views/playingScreen.js
```diff
@@ -257,45 +233,42 @@ class PlayingScreen extends React.Component {
...
     this.props.handleVrPlayerFocus();
   }
 
-  unmuteClick(event) {
-    this.props.controller.handleMuteClick();
-  }
-
   /**
    *
    * @param {number} vrDuration - key for duraction in config
```
pilievOoyala: duraction -> duration

js/skin.js

js/skin.js

js/views/playingScreen.js

.eslintrc
KseniiaSokolovaEpam: The interesting idea, but now the params is props, without any specification. There is a plan to use [airbnb eslint|https://www.npmjs.com/package/eslint-config-airbnb] in future.

pilievOoyala: Sounds good.  We have the eslint settings that Yaroslav started. But we didn't implement that for all the repos (I think), but that's a separate discussion.

pilievOoyala:
LGTM

PR #1124, opened 2018-09-03
[bugfix/PLAYER-4226] Page is scrolled when user tries to navigate the video by touch&drag

KseniiaSokolovaEpam:
Page was scrolled when user tried to navigate the video 360 via touch&drag in window mode.

Document touch event listeners are now passive by default in Safari 11.1, which is bundled with iOS 11.3 (this change is documented in the [Safari 11.1](https://developer.apple.com/library/archive/releasenotes/General/WhatsNewInSafari/Articles/Safari_11_1.html) ) so because of that preventDefault in the callback did not work to prevent scrolling.

Tests are not needed.

### DIFF
--- a/js/views/pauseScreen.js
+++ b/js/views/pauseScreen.js
@@ -45,7 +45,7 @@ class PauseScreen extends React.Component {
     this.handleResize();
     this.hideVrPauseButton();
     document.addEventListener('mousemove', this.handlePlayerMouseMove, false);
-    document.addEventListener('touchmove', this.handlePlayerMouseMove, false);
+    document.addEventListener('touchmove', this.handlePlayerMouseMove, { passive: false });
     document.addEventListener('mouseup', this.props.handleVrPlayerMouseUp, false);
     document.addEventListener('touchend', this.props.handleVrPlayerMouseUp, false);
   }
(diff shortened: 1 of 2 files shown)


js/views/pauseScreen.js
```diff
@@ -45,7 +45,7 @@ class PauseScreen extends React.Component {
     this.handleResize();
     this.hideVrPauseButton();
     document.addEventListener('mousemove', this.handlePlayerMouseMove, false);
-    document.addEventListener('touchmove', this.handlePlayerMouseMove, false);
```
aeng7: Will this affect other mobile/tablet environments such as Android or the Microsoft Surface?

pilievOoyala:
LGTM except for alex's question.

js/views/pauseScreen.js
KseniiaSokolovaEpam: @aeng7 touch event is supported for these [browsers](https://caniuse.com/#feat=touch)
passiveSupported works for these [browsers](https://caniuse.com/#feat=passive-event-listener) . As I see all supported touch event browsers support passiveSupported.

But you right that this changes will affect mobile vr video behavior. Because of that I asked QA Engineers to do regression after merge.

aeng7: Sounds good, thanks for the update!

PR #1125, opened 2018-09-06
bugfix/PLAYER-3966 View direciton is not reset to initial from the first time via on-screen contorl

KseniiaSokolovaEpam:
Prevent dowble call rotation method.
Move to es6 syntax.

The [associated PR](https://git.corp.ooyala.com/projects/PBW/repos/video-plugins-internal/pull-requests/350)

Unit tests passed.

### DIFF
--- a/js/components/viewControlsVr.js
+++ b/js/components/viewControlsVr.js
@@ -1,23 +1,24 @@
 var React = require('react');
-var DirectionControlVr = require('./directionControlVr');
 var Icon = require('../components/icon');
 var classnames = require('classnames');
 var _ = require('underscore');
 var createReactClass = require('create-react-class');
 var PropTypes = require('prop-types');
 
+import { DirectionControlVr } from './directionControlVr';
+
 var ViewControlsVr = createReactClass({
   /**
    * @method ViewControlsVr#handleVrViewControlsClick
    * @public
-   * @param {Event} e - event
+   * @param {Event} event - event object
    * @param {boolean} isRotated - true - if need to rotate; false - stop rotation
    * @param {string} direction - direction for rotation: "left", "right", "up", "down"
    */
-  handleVrViewControlsClick: function(e, isRotated, direction) {
-    if (e.type === 'touchend' || !this.props.controller.state.isMobile) {
-      e.stopPropagation(); // W3C
-      e.cancelBubble = true; // IE
+  handleVrViewControlsClick: function(event, isRotated, direction) {
+    if (event.type === 'touchend' || !this.props.controller.state.isMobile) {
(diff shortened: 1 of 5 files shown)


js/components/viewControlsVr.js
```diff
@@ -1,23 +1,24 @@
...
    * @param {string} direction - direction for rotation: "left", "right", "up", "down"
    */
-  handleVrViewControlsClick: function(e, isRotated, direction) {
-    if (e.type === 'touchend' || !this.props.controller.state.isMobile) {
-      e.stopPropagation(); // W3C
-      e.cancelBubble = true; // IE
+  handleVrViewControlsClick: function(event, isRotated, direction) {
+    if (event.type === 'touchend' || !this.props.controller.state.isMobile) {
+      event.stopPropagation(); // W3C
+      event.cancelBubble = true; // IE
```
jpfranco: Is `cancelBubble` still needed? I think IE11 should be able to handle `stopPropagation()`

js/views/playingScreen.js
```diff
@@ -79,11 +79,12 @@ class PlayingScreen extends React.Component {
    */
   handleVrAnimationEnd(ref, stateName) {
     if (ref) {
+      const _this = this;
```
jpfranco: Would it be better to declare `animationEndHandler` as a method of this component? Ideally we would also remove the listener on `componentWillUnmount` so using a method would allow us to do so and also allow us to avoid the `_this = this` declaration.

jpfranco:
Can we add a unit test for the error condition that was fixed?

aeng7:
LGTM other than @jpfranco's comments

js/views/playingScreen.js
zelibobla: JPFranco AFAIK, moving animationEndHandler into class level method will not help to fix execution context `this`, due to it's being called in event handling flow, when context is `window`. Shortest way to achieve fixed context is to set in 82 line something like: `const animationEndHandler = () => {...};`

js/components/viewControlsVr.js
KseniiaSokolovaEpam: \+

js/views/playingScreen.js
jpfranco: @zelibobla yes, of course. What I had in mind was to then either bind the handler in the constructor or use a class property (although that would require an additional babel plugin). The main point here is that it's probably a good idea to remove the listener when the component un-mounts. 

Anyway, I think its unlikely that the component will actually unmount with an active listener, but I think it's a good practice to remove DOM listeners on mount. I think this is fine though

PR #1126, opened 2018-09-07
Player 4170 - Audio Only Player

aeng7:
Main changes:

-There is a new controller state ```audioOnly```, which is set to true if the skin config has it set to true
-There is a new screen ```AudioOnlyScreen``` that is used if the player state is ```PLAYING```, ```PAUSED``` ,```END``` ,```START``` ,```INITIAL```
-Control Bar now has a ```skipControls``` item definition. The scrubber bar is now an optional part of the Control Bar.
-Skip Controls now have a ```playPause``` item definition
-Various video only styles have been moved to classes with the suffix ```-video```. These were styles that were marked with ```!important``` or any other styles I did not want to spend too much time working around for the audio only player. These can be looked into in the future

There are a few open questions in the PLAYER-4170 ticket for Fernando. Work on these questions will start once he has answered them.


js/components/controlBar.js
```diff
@@ -168,17 +171,21 @@ var ControlBar = createReactClass({
...
-      evt.cancelBubble = true; // IE
-      if (!this.props.controller.state.volumeState.volumeSliderVisible) {
-        this.props.controller.showVolumeSliderBar();
+    if (this.props.clickToVolumeScreen) {
+      this.props.controller.toggleScreen(CONSTANTS.SCREEN.VOLUME_SCREEN, true);
+    } else {
+      if (this.isMobile) {
+        this.props.controller.startHideControlBarTimer();
+        evt.stopPropagation(); // W3C
+        evt.cancelBubble = true; // IE
```
jpfranco: This is probably not needed anymore

js/components/controlBar.js
```diff
@@ -359,32 +358,16 @@ var ControlBar = createReactClass({
...
-    if (
-      this.props.duration === null ||
-      typeof this.props.duration === 'undefined' ||
-      this.props.duration === ''
-    ) {
-      totalTime = Utils.formatSeconds(0);
-    } else {
-      totalTime = Utils.formatSeconds(this.props.duration);
-    }
+    var totalTime = this.props.getTotalTime();
```
jpfranco: Is it necessary to call a function in this case or could the formatted value be simply passed as a prop?

js/components/controlBar.js
```diff
@@ -480,7 +454,7 @@ var ControlBar = createReactClass({
             tooltip={mutedInUi ? CONSTANTS.SKIN_TEXT.UNMUTE : CONSTANTS.SKIN_TEXT.MUTE}
             onClick={this.handleVolumeIconClick}>
           </ControlButton>
-          <VolumeControls {...this.props} />
+          {!this.props.hideVolumeControls ? <VolumeControls {...this.props} /> : null}
```
jpfranco: Completely subjective opinion here, but I think the 
`{ condition && <Component /> }`
pattern is more readable. Returning `false` in these statements is valid

js/components/controlBar.js
```diff
@@ -727,14 +701,37 @@ var ControlBar = createReactClass({
               null
           }
           height={this.props.skinConfig.controlBar.logo.height} />
+      ),
+
+      skipControls: (
+        <SkipControls
+          key={CONSTANTS.CONTROL_BAR_KEYS.SKIP_CONTROLS}
+          audioOnly={this.props.audioOnly}
```
jpfranco: I haven't seen this part of the code yet, but would it make more sense to make the component configurable as to meet audio only requirements instead of making it aware of audio only logic?

js/components/controlBar.js
```diff
@@ -879,33 +876,38 @@ var ControlBar = createReactClass({
         alignment = CONSTANTS.TOOLTIP_ALIGNMENT.LEFT;
       }
       tooltipAlignments[collapsedControlBarItems[k].name] = alignment;
-      finalControlBarItems.push(controlItemTemplates[collapsedControlBarItems[k].name]);
+      var item = controlItemTemplates[collapsedControlBarItems[k].name];
+      if (this.props.equalSpacing) {
+        item = (
+          <div className="oo-flex-row" key={collapsedControlBarItems[k].name}>
```
jpfranco: Is it necessary to add an extra container here? I think this might make it slightly more complex to maintain/update the CSS and we also have two places where the key is set. Could we do something like:

- Add a modifier class at the control bar level and use it customize the styles of `oo-control-bar-item` elements 
- Add an additional class to the control bar buttons when this mode is active

js/components/controlBar.js
```diff
@@ -879,33 +876,38 @@ var ControlBar = createReactClass({
...
+
     return finalControlBarItems;
   },
 
   render: function() {
     var controlBarClass = ClassNames({
       'oo-control-bar': true,
-      'oo-control-bar-hidden': !this.props.controlBarVisible
+      'oo-control-bar-hidden': !this.props.controlBarVisible,
+      'oo-control-bar-video': !this.props.simpleControlBar
```
jpfranco: Could we use a different name for this prop? Without looking at the code it's not clear what will happen when this is set

js/components/controlBar.js
```diff
@@ -879,33 +876,38 @@ var ControlBar = createReactClass({
...
     });
 
     var controlBarItems = this.populateControlBar();
-    var controlBarStyle = {
```
jpfranco: Does removing this have any effect on the regular player mode?

js/components/controlBar.js
```diff
@@ -879,33 +876,38 @@ var ControlBar = createReactClass({
...
-        style={controlBarStyle}
         onFocus={this.props.onFocus}
         onBlur={this.props.onBlur}
         onMouseUp={this.handleControlBarMouseUp}
         onTouchEnd={this.handleControlBarMouseUp}>
-        <ScrubberBar {...this.props} />
+        {!this.props.hideScrubberBar ? <ScrubberBar {...this.props} /> : null}
 
-        <div className="oo-control-bar-items-wrapper">
+        <div className="oo-control-bar-items-wrapper oo-flex-row-parent">
```
jpfranco: Does this class need to be added conditionally for audio only mode?

js/components/scrubberBar.js
```diff
@@ -402,6 +403,10 @@ var ScrubberBar = createReactClass({
       );
     }
 
+    if (!this.props.audioOnly) {
+      scrubberBarClassName += ' oo-scrubber-bar-video';
```
jpfranco: I know that this is how the old code was doing it, but could we use the `className` utility here?

js/components/skipControls.js
```diff
@@ -146,6 +156,13 @@ class SkipControls extends React.Component {
       buttonStyle.pointerEvents = 'none';
     }
 
+    var playButtonDetails = Utils.getPlayButtonDetails(this.props.controller.state.playerState);
+    var playIcon = playButtonDetails.icon;
```
jpfranco: Another subjective comment, but I think this might be cleaner if we used `playButtonDetails.icon` directly in the component in order to avoid the extra lines. Otherwise I think these should be declared as `const`

js/components/skipControls.js
```diff
@@ -202,6 +220,20 @@ class SkipControls extends React.Component {
...
+
+    buttonTemplate[CONSTANTS.SKIP_CTRLS_KEYS.PLAY_PAUSE] = (
+      <ControlButton
+        {...this.props}
+        key={CONSTANTS.CONTROL_BAR_KEYS.PLAY_PAUSE}
+        className="oo-play-pause"
+        focusId={CONSTANTS.CONTROL_BAR_KEYS.PLAY_PAUSE}
+        ariaLabel={playPauseAriaLabel}
+        icon={playIcon}
+        tooltip={playBtnTooltip}
```
jpfranco: Tooltips are not currently enabled for this component, so it might be better to remove this otherwise it would be the only one to show up

js/components/skipControls.js
```diff
@@ -227,7 +259,7 @@ class SkipControls extends React.Component {
       buttonId === CONSTANTS.SKIP_CTRLS_KEYS.SKIP_FORWARD
     );
 
-    const isDisabled = (
+    const isDisabled = !this.props.audioOnly && (
```
jpfranco: Could this be something more generic like `forceShowButtons` instead of checking for audio only?

js/components/skipControls.js
```diff
@@ -244,9 +276,15 @@ class SkipControls extends React.Component {
    */
   getSortedButtonEntries() {
     const buttons = [];
+    let key;
+    if (this.props.audioOnly) {
```
jpfranco: Would it be better to pass the right button configuration as a prop instead of checking for audio only? The component could just use that object instead of looking for it directly on `skinConfig`

js/components/skipControls.js
```diff
@@ -276,7 +314,9 @@ class SkipControls extends React.Component {
 
     const className = classNames('oo-skip-controls', {
       'oo-inactive': this.props.isInactive,
-      'oo-in-background': this.props.isInBackground
+      'oo-in-background': this.props.isInBackground,
+      'oo-absolute-centered': true,
```
jpfranco: Since these are related to layout, could we expose a `className` prop so that parent components can set the class that works with their layout?

js/components/volumePanel.js
```diff
@@ -0,0 +1,79 @@
...
+class VolumePanel extends React.Component {
+
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+
+    return (
+      <div className="oo-absolute-centered">
```
jpfranco: Can we add a class name that identifies this component, such as `oo-volume-panel`. This will make it easier for customers to customize the styles of this component.

js/controller.js
```diff
@@ -2205,7 +2216,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       }
     },
 
-    toggleScreen: function(screen) {
+    toggleScreen: function(screen, doNotPause) {
```
jpfranco: Can you add a jsdoc since you're modifying the function signature?

js/skin.js
```diff
@@ -340,268 +384,338 @@ const Skin = createReactClass({
...
+        switch (this.state.screenToShow) {
+          case CONSTANTS.SCREEN.INITIAL_SCREEN:
+          case CONSTANTS.SCREEN.START_SCREEN:
+          case CONSTANTS.SCREEN.PLAYING_SCREEN:
+          case CONSTANTS.SCREEN.PAUSE_SCREEN:
+          case CONSTANTS.SCREEN.END_SCREEN:
+            screen = (
+              <AudioOnlyScreen
+                {...this.props}
+                handleVrPlayerMouseDown={this.handleVrPlayerMouseDown}
```
jpfranco: Should we ignore or disable these VR events for audio only?

js/skin.js
```diff
@@ -340,268 +384,338 @@ const Skin = createReactClass({
...
+            break;
+          default:
+            screen = <div />;
+        }
       }
     }
 
+    var className = this.state.responsiveClass;
+    if (this.props.controller.state.audioOnly) {
+      className += " oo-audio-only";
```
jpfranco: Can we use the `classNames` utility here?

js/views/audioOnlyScreen.js
```diff
@@ -0,0 +1,83 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      controlBarVisible: true,
+      animate: false
+    };
+  }
+
+  render() {
+    var titleStyle = {
```
jpfranco: Can we use `const` for these declarations?

js/views/audioOnlyScreen.js
```diff
@@ -0,0 +1,83 @@
...
+      color: this.props.skinConfig.startScreen.titleFont.color
+    };
+    var infoPanelClass = ClassNames({
```
jpfranco: Is it necessary to use `ClassNames` here considering that none of the classes are dynamic? I think it helps with readability, but just wondering if any of there are supposed to change under certain conditions

scss/components/_control-bar.scss
```diff
@@ -169,8 +149,26 @@
...
+  background: $control-bar-bg;
+  @include vendor-prefixes (background, $control-bar-gradient);
+  @include vendor-prefixes (transition, $medium-transition);
+  height: $control-bar-height;
+  padding: 0 15px;
+
+  .oo-control-bar-items-wrapper {
+    margin-top: 10px;
+
+    & > * {
```
jpfranco: Is it possible to replace the `*` operator with something more explicit? If we add different elements to this container in the future they will inherit this style automatically and might be hard to debug

scss/components/_state-screen.scss
```diff
@@ -1,3 +1,21 @@
+.oo-state-screen-audio {
+  height: 100%;
+
+  .oo-state-screen-audio-title {
+    @include vendor-prefixes (transition, $fade-up-transition);
+    pointer-events: none;
+    width: 100%;
+    font-size: 1em;
+
+    .oo-state-screen-audio-info {
```
jpfranco: Another personal opinion but I prefer to avoid more than one level of nesting. This style will result in  a rule like '.oo-state-screen-audio .oo-state-screen-audio-title .oo-state-screen-audio-info` which has high specificity and is hard to override. If you had to add a media query for this element, for example, you'd have to recreate this hierachy in order for the style to take effect

js/views/audioOnlyScreen.js
```diff
@@ -0,0 +1,83 @@
...
+        : {this.props.contentTree.description}
+      </div>
+    );
+    return (
+      <div className="oo-state-screen-audio oo-flex-column-parent">
+        <div className={infoPanelClass}>
+          {titleMetadata}
+        </div>
+        <div className="oo-interactive-container oo-flex-column oo-flex-column-parent">
+          <ControlBar
```
jpfranco: Since the control bar displays all buttons in a row, will this approach scale if we later need to organize the elements with a different layout (i.e. multiple rows, etc.)?

scss/base/_utilities.scss
```diff
@@ -26,6 +26,42 @@
...
+
+.oo-flex-column {
+  @include flexbox;
+  @include flex-direction(column);
+  flex: 1;
+}
+
+.oo-flex-column:first-child > * { margin-bottom: auto; }
+
+.oo-flex-column:last-child > * { margin-top: auto; }
```
jpfranco: Same as previous comment, I think it would better to avoid the `*` selector if possible

js/components/controlBar.js
aeng7: Ah, yep, there are a few things that this could be applied to. I missed this when cleaning up the code. I'll add these new configs in. However, we may still need the audioOnly prop to retrieve the audio only button config. Do you know of a way to do this otherwise?

js/components/controlBar.js
aeng7: Will do

js/components/controlBar.js
aeng7: I forgot to add a comment about this. The height for the control bar is defined in our css. I wanted QA to see if removing this affected anything, since there were two sources of truth. Nice catch though!

js/components/controlBar.js
aeng7: Nope, the control bar had flexbox styles on it before. I've moved them to oo-flex-row-parent.

js/components/scrubberBar.js
aeng7: Sounds good, will do.

js/components/skipControls.js
aeng7: Will add

js/components/skipControls.js
aeng7: Will do

js/components/skipControls.js
aeng7: Yep, will add

js/components/skipControls.js
aeng7: I guess I wanted to try to keep the configurability of the skip controls. I'll double check with Steve, but if we don't want this to be configurable, I'll make this change.

js/components/skipControls.js
aeng7: Will add

js/components/volumePanel.js
aeng7: Missed this, will add.

js/controller.js
aeng7: Haha deja vu, will add

js/components/controlBar.js
jpfranco: Yes so definitely at some point we'll need the audioOnly check, I just think it might be better at a higher level. The button config could be passed to the component as a prop (instead of it getting it directly from the skin config) that way the check happens earlier and the SkipControls don't need to know about audio only

js/skin.js
aeng7: Sounds good, will add

js/skin.js
aeng7: Yep

js/views/audioOnlyScreen.js
aeng7: Yep

js/views/audioOnlyScreen.js
aeng7: I'll check this

scss/components/_control-bar.scss
aeng7: I'm not sure, this was part of the original control bar styling. I copied this from https://github.com/ooyala/html5-skin/pull/1126/files/c5ea5e71f2fd39b8872ef99dc7df98a39d662f10#diff-256e213b9742bfdba7b41aa6315beb3eL32. I'd prefer not to change this in this PR, but we should look into this in the future.

scss/components/_state-screen.scss
aeng7: Will take a look

js/components/controlBar.js
aeng7: Agreed, will make these changes

js/components/controlBar.js
aeng7: I'm going to add this back in. I did not notice that it's taking the height from the skin config

js/components/controlBar.js
```diff
@@ -313,18 +319,10 @@ var ControlBar = createReactClass({
...
-      playIcon = 'pause';
-      playPauseAriaLabel = CONSTANTS.ARIA_LABELS.PAUSE;
-    } else if (this.props.playerState === CONSTANTS.STATE.END) {
-      playIcon = 'replay';
-      playPauseAriaLabel = CONSTANTS.ARIA_LABELS.REPLAY;
-    } else {
-      playIcon = 'play';
-      playPauseAriaLabel = CONSTANTS.ARIA_LABELS.PLAY;
-    }
+    var playButtonDetails = Utils.getPlayButtonDetails(this.props.playerState);
```
pilievOoyala: just a note, but should we make the way we spell button consistent here?
You have playButtonDetails and playBtnTooltip.  The length of the var is about the same if they are consistent.

js/components/skipControls.js
```diff
@@ -146,6 +156,10 @@ class SkipControls extends React.Component {
       buttonStyle.pointerEvents = 'none';
     }
 
+    const playButtonDetails = Utils.getPlayButtonDetails(this.props.controller.state.playerState);
+
```
pilievOoyala: remove the extra line

js/controller.js
```diff
@@ -1040,6 +1042,8 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
         this.state.screenToShow = CONSTANTS.SCREEN.SHARE_SCREEN;
       } else {
         this.state.screenToShow = CONSTANTS.SCREEN.END_SCREEN;
+        //TODO: END_SCREEN_SHOWN is not consumed by anyone, should we remove? If not, should
+        //we publish this when the end screen is actually shown?
```
pilievOoyala: We probably should move the publish to after the renderskin. Also, I'd rather be thorough and send the message without anyone listening then having to figure out where to add this line later if we do need it.

js/controller.js
```diff
@@ -2216,7 +2227,12 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
             this.state.screenToShow = screen;
             this.renderSkin();
           }.bind(this);
-          this.mb.publish(OO.EVENTS.PAUSE);
+          if (doNotPause) {
+            this.pausedCallback();
```
pilievOoyala: This seems a little counterintuitive. Why should you call the callback saying that we paused successfully if we aren't supposed to pause?

js/views/audioOnlyScreen.js
aeng7: Nope, none of these currently change. Should I leave the ```ClassNames``` usage since it improves readability?

jpfranco: @aeng7 I would say yes, although ` 'oo-inactive': false` is probably unnecesary

js/components/controlBar.js
aeng7: Ah I moved these lines, so that is how they were spelled before. Will make this change.

js/controller.js
aeng7: The render skin is not instant and is async. It would be a bigger change than moving the publish after the ```renderSkin``` call. I think this is outside the scope of this PR and was just something I noticed. Can we make this change in some future PR?

js/controller.js
aeng7: The ```pausedCallback``` is set right above this line. This is normally called after the player pauses. We now have the concept of a content screen that does not pause the video. To accomplish this, I've modified the ```toggleScreen``` function to not always pause the video. When this is the case, we just immediately call the callback we just defined and set it to null.

js/components/controlBar.js
```diff
@@ -727,14 +696,44 @@ var ControlBar = createReactClass({
...
           }
           height={this.props.skinConfig.controlBar.logo.height} />
+      ),
+
+      skipControls: (
+        <SkipControls
+          key={CONSTANTS.CONTROL_BAR_KEYS.SKIP_CONTROLS}
+          buttonConfig={Utils.getPropertyValue(
+            this.props.skinConfig,
+            'skipControls.buttonsWithPlayPause',
```
jpfranco: Is this the same for both audio only and the regular player?

js/components/controlBar.js
```diff
@@ -727,14 +696,44 @@ var ControlBar = createReactClass({
...
+            {}
+          )}
+          forceShowButtons={true}
```
jpfranco: Should this only be set for the audio-only player?

js/components/controlBar.js
jpfranco: Sorry, nevermind. Didn't realize that this is the control bar

aeng7: Yep, this is the control bar version of the skip controls, with the play/pause icon. This will not affect the regular skip controls shown on the playing/pause screens. Currently the control bar skip controls is only used for the audio only player.

js/components/controlBar.js
```diff
@@ -727,14 +696,44 @@ var ControlBar = createReactClass({
...
+          maxWidth={200}
+          className={'oo-absolute-centered oo-control-bar-item'}
```
jpfranco: We can probably change this later, but this could be just `className="
oo-absolute-centered oo-control-bar-item"`

js/components/skipControls.js
```diff
@@ -274,15 +302,21 @@ class SkipControls extends React.Component {
...
 
-    const className = classNames('oo-skip-controls', {
+    const className = classNames('oo-skip-controls', this.props.className, {
       'oo-inactive': this.props.isInactive,
       'oo-in-background': this.props.isInBackground
     });
     const buttonTemplate = this.getButtonTemplate();
 
+    const skipControlsStyle = {};
+    if (this.props.maxWidth) {
```
jpfranco: Is it necessary to set this using an inline style? Could this style be added to the custom class?

js/components/volumePanel.js
```diff
@@ -0,0 +1,77 @@
...
+
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+
+    return (
+      <div className="oo-absolute-centered oo-volume-panel">
+        <ControlBar
```
jpfranco: Sorry I hadn't noticed this before. Is it necessary to use a whole control bar for the volume button? The control bar is a fairly heavy component so it might be more efficient and a bit simpler to use the volume component on its own

js/components/controlBar.js
aeng7: Added

scss/responsive/_xsmall.scss
```diff
@@ -1,4 +1,5 @@
 .oo-xsmall {
+  height: 100%;
```
jpfranco: Does this have any effect on the regular player?

js/components/skipControls.js
aeng7: Added

scss/skin/_default.scss
```diff
@@ -56,6 +54,11 @@
   }
 }
 
+.oo-video-player {
+  // default aspect ratio 16:9
+  @include set-aspect-ratio(16, 9);
```
jpfranco: Would setting this on a different class have any effect on the auto aspect ratio feature of the player?

scss/responsive/_xsmall.scss
aeng7: I haven't noticed anything with the regular player.

scss/skin/_default.scss
aeng7: I pulled this out from the .oo-player class into its own .oo-video-player class, which is not applied for the audio only player (since we do not want auto aspect ratios for the audio player), but is applied to the same div that .oo-player is for the video player. Original location here:

https://github.com/ooyala/html5-skin/blob/master/scss/skin/_default.scss#L18

This should not have any effect for the auto aspect ratio feature for the video player.

jpfranco: Oh ok, got it. So this `oo-video-player` class is set on the same container as `oo-player` right?

js/components/volumePanel.js
aeng7: There are styles involving the control bar and calculations done in the control bar component for the volume controls. This component is also likely to be deprecated soon as the volume bar requirement has changed. Is it ok if we leave it as is for now?

scss/skin/_default.scss
aeng7: Yep

js/components/volumePanel.js
jpfranco: Ok sounds good. I think we should eventually refactor that to make a standalone volume bar

js/controller.js
pilievOoyala: Sure

js/controller.js
pilievOoyala: Ok, sounds good 👍

pilievOoyala:
LGTM

PR #1127, opened 2018-09-07
[PLAYER-4216] (PR 2 of 3) Fixes for MainHtml5 CC refactor

jpfranco:
Contains a few fixes needed in order for the MainHtml5 CC changes to work properly:

- No longer showing CC button if CC language list is empty
- Setting closed captions right before going fullscreen on iOS in order for the UI to be able to detect the selected language
- Allowing turning closed captions on with the `CHANGE_CLOSED_CAPTION_LANGUAGE` event


js/controller.js
```diff
@@ -2406,7 +2414,8 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       }
     },
 
-    setClosedCaptionsLanguage: function() {
+    setClosedCaptionsLanguage: function(params) {
```
aeng7: Since you're modifying the function signature, can you add a jsdoc for this?

js/controller.js
```diff
@@ -2441,7 +2451,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       this.renderSkin();
     },
 
-    onChangeClosedCaptionLanguage: function(event, language) {
+    onChangeClosedCaptionLanguage: function(event, language, params = {}) {
```
aeng7: Since you're modifying the function signature, can you add a jsdoc for this?

js/components/controlBar.js
```diff
@@ -800,11 +800,16 @@ var ControlBar = createReactClass({
...
       // do not show CC button if no CC available
       if (
-        (this.props.controller.state.isOoyalaAds ||
-          !this.props.controller.state.closedCaptionOptions.availableLanguages) &&
-        defaultItems[k].name === 'closedCaption'
+        defaultItems[k].name === CONSTANTS.CONTROL_BAR_KEYS.CLOSED_CAPTION &&
+        (
+          !availableLanguages ||
+          !availableLanguages.languages ||
+          !availableLanguages.languages.length ||
```
aeng7: Do we check for available languages elsewhere? Should we have this be a utility function instead?

js/controller.js
```diff
@@ -2452,6 +2462,14 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
 
       // validate language is available before update and save
       if (language && availableLanguages && _.contains(availableLanguages.languages, language)) {
+        // The act of changing the CC language doesn't currently enable captions
+        // automatically. The core will set the forceEnabled parameter to true when
+        // it is necessary to also enable captions themselves
+        if (params.forceEnabled) {
+          this.state.closedCaptionOptions.enabled = true;
+          this.state.persistentSettings.closedCaptionOptions.enabled = true;
```
aeng7: Is it possible to have a single source of truth for this (and possibly store persistent settings as an aside)?

js/controller.js
jpfranco: Added

js/controller.js
jpfranco: Added

js/components/controlBar.js
jpfranco: This is the only check like this. I think it would be a bit overkill to have a utility function for this, but I simplified the statement a bit.

js/controller.js
jpfranco: Yes but I think this is outside the scope of this PR. The closed captions logic in the skin also needs a good refactor, but it's a significant amount of effort.

aeng7: Sounds good

js/components/controlBar.js
aeng7: Nice!

PR #1132, opened 2018-09-19
[PLAYER-3859] Muted state now overrides volume for VoiceOver (Narrato…

zelibobla:
Fixed volume control attribute to reflect muted state (if muted) over the volume value.


.gitignore
```diff
@@ -2,4 +2,5 @@ node_modules
 build
 coverage
 .idea
-.DS_Store
\ No newline at end of file
+.DS_Store
+.vscode/
```
jpfranco: Is updating the skin config submodule (below) deliberate? I don't see any changes on the diffs for that

js/components/volumeControls.js
```diff
@@ -92,6 +92,9 @@ var VolumeControls = createReactClass({
    * @returns {String} The current volume in a screen reader friendly format (i.e. 20% volume).
    */
   getAriaValueText: function() {
+    if (this.props.controller.state.volumeState.muted) {
```
jpfranco: Can we add a unit test that checks that the correct label is returned?

package.json
```diff
@@ -23,55 +23,55 @@
...
+    "classnames": "^2.2.6",
     "create-react-class": "^15.6.3",
     "deepmerge": "^1.4.0",
     "lodash.debounce": "4.0.3",
     "lodash.values": "4.1.0",
-    "prop-types": "^15.6.1",
-    "react": "^16.4.1",
-    "react-dom": "^16.4.1",
+    "prop-types": "^15.6.2",
+    "react": "^16.5.2",
```
jpfranco: Good idea to take the opportunity to update to latest React, thanks

package.json
```diff
@@ -23,55 +23,55 @@
...
     "react-scrollbar": "^0.5.4",
     "screenfull": "3.0.0"
   },
   "devDependencies": {
-    "@babel/core": "^7.0.0-beta.51",
-    "@babel/preset-env": "^7.0.0-beta.51",
-    "@babel/preset-react": "^7.0.0-beta.51",
+    "@babel/core": "^7.1.0",
+    "@babel/preset-env": "^7.1.0",
+    "@babel/preset-react": "^7.0.0",
```
aeng7: Should we update this to 7.1.0 as well? To keep in sync with the other babel presets.

.gitignore
zelibobla: It's just automatically updated due to latest master version is newer than the one is referred from .gitmodules. But particular files being pulled via .gitmodules settings are not changed, so this update actually makes no sense. I can roll it back if necessary.

js/components/volumeControls.js
zelibobla: done

package.json
zelibobla: done

.gitignore
jpfranco: Oh ok no problem. If it's the latest from master then it's fine

pilievOoyala:
LGTM

PR #1133, opened 2018-09-20
Player 4450

pilievOoyala:


aldigun:
LGTM. Though, should we give a space after IF and before "("?

pilievOoyala:
Yeah, I'll fix that.

PR #1135, opened 2018-09-21
bugfix/PLAYER-4436 Drag and drop side ways does not work properly

KseniiaSokolovaEpam:
At some moments a height of the player was returned as 0, and as a result, the  function "onTouchMove" used Infinity as the value for "pitch" when rotating a vr video.

Unit tests passed.


js/mixins/responsiveManagerMixin.js
```diff
@@ -1,8 +1,9 @@
-var ReactDOM = require('react-dom'),
-    ClassNames = require('classnames'),
-    debounce = require('lodash.debounce');
+const ReactDOM = require('react-dom');
+const ClassNames = require('classnames');
+const debounce = require('lodash.debounce');
+const Utils = require('../components/utils');
```
zelibobla: ES6 import would be nice here

js/mixins/responsiveManagerMixin.js
```diff
@@ -27,23 +28,47 @@ var ResponsiveManagerMixin = {
...
+    return componentHeight;
+  },
+
   generateResponsiveData: function() {
-    var componentWidth = 0;
-    var componentHeight = 1;
-    var dom = ReactDOM.findDOMNode(this);
+    let componentWidth = 0;
+    let componentHeight = 0;
+    const dom = ReactDOM.findDOMNode(this);
```
zelibobla: ES6 destructuring would be good here:
const { width, height } = dom.getBoundingClientRect();

tests/mixins/responsiveManagerMixin-test.js
```diff
@@ -0,0 +1,63 @@
+jest.dontMock('../../js/mixins/responsiveManagerMixin');
+jest.dontMock('../../js/components/utils');
+
+const React = require('react');
+const Enzyme = require('enzyme');
+const debounce = require('lodash.debounce');
+const createReactClass = require('create-react-class');
+const responsiveManagerMixin = require('../../js/mixins/responsiveManagerMixin');
```
zelibobla: Give a look on how it can be imported: https://github.com/ooyala/html5-skin/blob/master/tests/components/volumeControls-test.js

tests/mixins/responsiveManagerMixin-test.js
```diff
@@ -0,0 +1,63 @@
...
+
+let Foo = createReactClass({
```
zelibobla: I believe you won't reassign Foo. So it's a const.

tests/mixins/responsiveManagerMixin-test.js
```diff
@@ -0,0 +1,63 @@
...
+  mixins: [responsiveManagerMixin],
+  render: function() {
+    return (
+      <div>{this.props.children}</div>
+    );
+  }
+});
+
+describe('responsiveManagerMixin', function() {
+  let props =  {
```
zelibobla: const

KseniiaSokolovaEpam: No, it may be changeable in future, if someone wants to add tests for the file.

tests/mixins/responsiveManagerMixin-test.js
KseniiaSokolovaEpam: I thing it may be changeable too.

tests/mixins/responsiveManagerMixin-test.js
KseniiaSokolovaEpam: I know about "import", thank you :-)

js/mixins/responsiveManagerMixin.js
```diff
@@ -27,23 +28,47 @@ var ResponsiveManagerMixin = {
     this.generateResponsiveData();
   },
 
+  /**
+   * It gives the value of height, based on the width of mainVideoContainer (in a ratio 16:9);
+   * @param {number} componentWidth - width of the element
+   * @returns {number} - default height of the element
+   * If we do not have width of mainVideoContainer it returns 0
+   */
+  getDefaultElementHeight: function(componentWidth) {
```
pilievOoyala: It would be nice to be able to pass in any ratio but default to 16:9.  Also, what about vertical video (9:16)?

KseniiaSokolovaEpam: @pilievOoyala As I understood we are not able to check if the video is vertical in moments when we have to use function "getDefaultElementHeight" (for example before video starts). Could you confirm it or correct me?

aeng7: Can we use the mainVideoAspectRatio calculated here?

https://github.com/ooyala/html5-skin/blob/master/js/controller.js#L1351

KseniiaSokolovaEpam: @aeng7 as I can see we pass 2 params to the function (>>calculateAspectRatio: function(width, height) {... }), but we do not know height in a start moment . That was the problem.

aeng7:
LGTM

PR #1139, opened 2018-10-02
[PLAYER-4508] Replace custom cast button

marcosmx:
This branch contains the implementation of the Cast SDK button that handles the interaction of the sender and the receiver.  
Also removes the custom button used to manage the cast action from sender to receiver


aeng7:
LGTM

PR #1140, opened 2018-10-03
[PLAYER-4542] Controls look too big on mobile devices

marcosmx:
This branch contains the fix for restore the `responsiveClass` state at skin.js. Also mixes it with the `.oo-responsive` class.

### DIFF
--- a/js/skin.js
+++ b/js/skin.js
@@ -706,7 +706,7 @@ const Skin = createReactClass({
       }
     }
 
-    var className = ClassNames('oo-responsive', {
+    var className = ClassNames(this.state.responsiveClass, 'oo-responsive', {
       'oo-audio-only': this.props.controller.state.audioOnly
     });
 


aeng7:
LGTM

zelibobla:
LGTM

js/skin.js
```diff
@@ -706,7 +706,7 @@ const Skin = createReactClass({
       }
     }
 
-    var className = ClassNames('oo-responsive', {
+    var className = ClassNames(this.state.responsiveClass, 'oo-responsive', {
```
KseniiaSokolovaEpam: Where do you set value for responsiveClass state?

marcosmx: @KseniiaSokolovaEpam that value comes from `ResponsiveManagerMixin` used as a mixin at the Skin component at line 32

PR #1141, opened 2018-10-03
[PLAYER-4543] Enable/Disable chromecast button by page player param

marcosmx:
This branch contains the needed logic to enable/disable the chromecast button by using a player param with the following structure:
```
"chromecast": {
   "enable": true
}
```

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -414,10 +414,6 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
 
       this.state.hideMultiAudioIcon = !!params.hideMultiAudioIcon;
 
-      if (this.state.playerParam.chromecast && this.state.playerParam.chromecast.enable && this.state.playerParam.chromecast.enable == true) {
-        this.state.enableChromecast = true;
-      }
-
       // initial DOM manipulation
       this.state.mainVideoContainer.addClass('oo-player-container');
       this.state.mainVideoInnerWrapper.addClass('oo-player');
@@ -457,6 +453,16 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
 
       this.accessibilityControls = this.accessibilityControls || new AccessibilityControls(this); // keyboard support
       this.state.screenToShow = CONSTANTS.SCREEN.INITIAL_SCREEN;
+    },
+
+
+    isChromecastEnabled: function (params) {
+      var chromecastConfig = params.chromecast;
+      var appId = Utils.getPropertyValue(chromecastConfig, 'appId', '');
+      if (typeof appId === 'string' && appId !== '' && Utils.getPropertyValue(chromecastConfig, 'enable', false)) {
+          return true;
+        }
+      return false;
     },
 
     /**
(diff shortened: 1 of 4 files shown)


js/controller.js
```diff
@@ -412,6 +414,10 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
 
       this.state.hideMultiAudioIcon = !!params.hideMultiAudioIcon;
 
+      if (this.state.playerParam.chromecast && this.state.playerParam.chromecast.enable && this.state.playerParam.chromecast.enable == true) {
```
aldigun: Not sure about the assignment here. This makes the player param looks like:
var param = {
  chromecast.enable = true
};

Should we simplify it to be:
var params = {
  chromecast: true
}

aldigun: Please use strict equality ===

aldigun: I'm okay with either one.

aldigun: Thanks for confirming. Before we make any decision, I found that, even when you do not specify appID, the cast button still appear and you are able to cast. I think we need to disable chromecast IF:

chromecast.appID is not provided, and/OR if chromecast.enable is set to FALSE. By default, chromecast.enable is FALSE.
chromecast.appID should not have a default appID.

marcosmx: @aldigun Yeah, if the appId it's nor provided the core.js it's using our Prod AppID by default...Ok, I will work on that changes

marcosmx:
@aldigun Please review the new changes, this cover the scenario and changes on logic that you requested.

js/controller.js
```diff
@@ -453,6 +455,16 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       this.state.screenToShow = CONSTANTS.SCREEN.INITIAL_SCREEN;
     },
 
+
+    isChromecastEnabled: function (params) {
+      var chromecastConfig = params.chromecast;
```
aldigun: What if params.chromecast is undefined or null?

js/controller.js
```diff
@@ -453,6 +455,16 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
+      var appId = Utils.getPropertyValue(chromecastConfig, 'appId', '');
```
aldigun: What id appID is undefined or null, will getPropertyValue returns and not crashing? And is it returning the right value?

js/controller.js
```diff
@@ -453,6 +455,16 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
+      if (typeof appId === 'string' && appId !== '' && Utils.getPropertyValue(chromecastConfig, 'enable', false)) {
```
aldigun: What if "enable" is undefined or null?

aldigun:
LGTM if my comments are answered.

marcosmx:
@aldigun Regarding your questions, the util `getPropertyValue` takes care of those cases and I added more test cases to cover that special cases.

aldigun:
+1

PR #1142, opened 2018-10-17
[PLAYER-4385] Now reacting to playerType page-level param

jpfranco:
The player will now use the `playerType` page-level param in order enter audio-only mode instead of the `audioOnly` value from the skin.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -1638,7 +1638,17 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       var uiLanguage = Utils.getLanguageToUse(this.state.config);
       this.mb.publish(OO.EVENTS.SKIN_UI_LANGUAGE, uiLanguage);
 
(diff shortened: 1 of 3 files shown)


js/controller.js
```diff
@@ -1638,7 +1638,17 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
+      switch (params.playerType) {
+        case OO.CONSTANTS.PLAYER_TYPE.AUDIO:
+          this.state.audioOnly = true;
+          break;
+        case OO.CONSTANTS.PLAYER_TYPE.VIDEO:
+        default:
+          this.state.audioOnly = false;
+          break;
+      }
+
```
zelibobla: Could we replace it by one line?
`this.state.audioOnly = params.playerType === OO.CONSTANTS.PLAYER_TYPE.AUDIO;`

jpfranco: Updated

PR #1144, opened 2018-10-29
Feature/player 4600 - Migration from Gulp to Webpack

ALexanderSPB:


sample.html
```diff
@@ -5,8 +5,9 @@
   <script src="//player.ooyala.com/static/v4/production/latest/core.min.js"></script>
   <script src="//player.ooyala.com/static/v4/production/latest/video-plugin/main_html5.min.js"></script>
   <script src="//player.ooyala.com/static/v4/production/latest/video-plugin/bit_wrapper.min.js"></script>
-  <script src="//player.ooyala.com/static/v4/production/latest/skin-plugin/html5-skin.min.js"></script>
-  <link rel="stylesheet" href="//player.ooyala.com/static/v4/production/latest/skin-plugin/html5-skin.min.css"/>
+  <!--<script src="//player.ooyala.com/static/v4/production/latest/skin-plugin/html5-skin.min.js"></script>-->
```
aeng7: What was this change done for?

aeng7:
LGTM other than a minor comment

sample.html
ALexanderSPB: Html-webpack-plugin uses this file as a template to inject there the generated `html5-skin.min.js` and `html5-skin.min.css`. So with webpack-dev-server we will see the actual changes during development. 
I commented those `script` and `link` tags to avoid duplications in that case.
Is it better to revert changes in this file and create a new file - template for html-webpack-pluginwithout those lines ?

KseniiaSokolovaEpam:
@ALexanderSPB  Do we need to change README?

sample.html
aeng7: I think it would be better to do that. This sample.html is meant to show users of our skin a sample of how things are used.

aeng7:
If we update the README, we could include a section on the new template file if we go that route.

zelibobla:
I vote for @aeng7 original idea. Leave sample.html as it was and create one more `index.html` or something like it for debugging purposes only, if it's really necessary.

aeng7:
LGTM

PR #1145, opened 2018-10-30
Bugfix - PLAYER-4584

ALexanderSPB:
https://jira.corp.ooyala.com/browse/PLAYER-4584

Added oo-flex-end css class for justify-content: flex-end.

### DIFF
--- a/tests/components/controlBar-test.js
+++ b/tests/components/controlBar-test.js
@@ -926,6 +926,12 @@ describe('ControlBar', function() {
       var wrapperPause = Enzyme.mount(getControlBar(CONSTANTS.STATE.PAUSE));
       expect(typeof wrapperPause.ref('volumeIcon')).toBe('undefined');
     });
+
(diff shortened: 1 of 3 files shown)


aeng7:
@ALexanderSPB can we add a unit test for this? Otherwise LGTM

tests/components/controlBar-test.js
```diff
@@ -926,6 +926,12 @@ describe('ControlBar', function() {
       var wrapperPause = Enzyme.mount(getControlBar(CONSTANTS.STATE.PAUSE));
       expect(typeof wrapperPause.ref('volumeIcon')).toBe('undefined');
     });
+
+    it('uses oo-flex-end class if audio-only', function () {
```
KseniiaSokolovaEpam: Why does this test pass if you have not set an OS to iOS?

ALexanderSPB: It was set at the 902 line.

PR #1147, opened 2018-11-05
[PLAYER-4582] Disabling Skip Forward button when at video end

jpfranco:
The "Skip Forward" button will basically be greyed out and disabled when the video has ended. This is mostly for the audio-only player, but the behavior should be consistent for all player types.

Note that this doesn't fully fix the issue from PLAYER-4582, this is just a partial update for now.

### DIFF
--- a/js/components/skipControls.js
+++ b/js/components/skipControls.js
@@ -18,7 +18,7 @@ class SkipControls extends React.Component {
     this.onNextVideo = this.onNextVideo.bind(this);
     this.onSkipBackward = this.onSkipBackward.bind(this);
     this.onSkipForward = this.onSkipForward.bind(this);
-    this.isAtVideoEnd = this.isAtVideoEnd.bind(this);
+    this.isAtVideoEdge = this.isAtVideoEdge.bind(this);
     this.onMouseEnter = this.onMouseEnter.bind(this);
     this.onPlayPauseClick = this.onPlayPauseClick.bind(this);
   }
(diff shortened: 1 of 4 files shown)


js/components/skipControls.js
```diff
@@ -98,25 +98,27 @@ class SkipControls extends React.Component {
...
-   * playhead state and duration.
+   * Determines whether or not the current video is at the end (VOD) or at the
+   * live edge (DVR Live Streams) based on the playhead state and duration.
    * @private
-   * @return {Boolean} True if the video is at the live edge, false otherwise.
-   * Note: This function always returns false for VOD.
+   * @return {Boolean} True if the video is at the video end/live edge, false otherwise.
    */
-  isAtLiveEdge() {
+  isAtVideoEnd() {
```
zelibobla: I guess `End` is less correct than `Edge`. For live stream we're not in the end, so it's a kind of confusing for newcomer what this function does. My suggestion: `isAtVideoEdge`.

js/components/skipControls.js
```diff
@@ -98,25 +98,27 @@ class SkipControls extends React.Component {
...
-      const duration = Utils.getPropertyValue(this.props.controller, 'state.duration', 0);
-      const currentPlayhead = Utils.ensureNumber(this.props.currentPlayhead, 0);
-      const isLiveNow = Math.abs(currentPlayhead - duration) < 1;
-      return isLiveNow;
+      isVideoEnd = Math.abs(currentPlayhead - duration) < 1;
+    } else {
+      isVideoEnd = currentPlayhead >= duration;
     }
-    return false;
+    return isVideoEnd;
```
zelibobla: Let me suggest a bit shorted and more self explanatory code that doing the same:
```suggestion
      const duration = Utils.getPropertyValue(this.props.controller, 'state.duration', 0);
      const currentPlayhead = Utils.ensureNumber(this.props.currentPlayhead, 0);
      const isAtLiveEdge = isLiveStream && Math.abs(currentPlayhead - duration) < 1;
      const isAtVideoEnd = !isLiveStream && currentPlayhead >= duration;
      return isAtLiveEdge || isAtVideoEnd;
```

zelibobla: Also I'm not sure if I got right the root case, but I saw 4 seconds gap before video end, that should be considered (from your comment in the ticket), and I don't see you consider this gap in the code. Aren't we missing something like `currentPlayhead >= duration - 4;`?

js/components/skipControls.js
jpfranco: Will update the name

js/components/skipControls.js
jpfranco: @zelibobla I disagree with that being a more self explanatory code. The purpose of the chosen syntax is to improve readability, not to minimize the lines of code.

jpfranco: @zelibobla as mentioned on the PR description, this update is not addressing the root cause, it's just covering the UI side of the problem. The root cause will be addressed on a separate PR for video-plugins-internal

PR #1148, opened 2018-11-07
[PBW-7594] Fixing slider drag action on Firefox

jpfranco:
This fixes an issue in which the CC menu slider couldn't be dragged with the mouse on Firefox. It also maintains the behavior in which the blue focus outline is not shown while interacting with the mouse, but is activated when using the keyboard.

### DIFF
--- a/js/components/slider.js
+++ b/js/components/slider.js
@@ -130,6 +130,10 @@ var Slider = createReactClass({
 
   onMouseUp: function(event) {
     Utils.blurOnMouseUp(event);
+
+    this.setState({
+      isDragging: false
+    });
   },
 
   onMouseMove: function(event) {
--- a/js/components/slider.js
+++ b/js/components/slider.js
@@ -7,6 +7,13 @@ var createReactClass = require('create-react-class');
 var PropTypes = require('prop-types');
 
 var Slider = createReactClass({
+
+  getInitialState: function() {
+    return {
+      isDragging: false
+    };
+  },
+
   componentDidMount: function() {
     this.handleSliderColoring(this.props);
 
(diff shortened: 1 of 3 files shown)


tests/components/slider-test.js
```diff
@@ -153,13 +153,15 @@ describe('Slider', function() {
...
     renderComponent();
+    expect(element.classList.contains('oo-dragging')).toBe(false);
     component.simulate('mouseDown', { currentTarget: element });
+    expect(element.classList.contains('oo-dragging')).toBe(true);
     component.simulate('mouseMove', { currentTarget: element });
-    expect(spy.callCount).toBe(1);
-    spy.restore();
+    expect(element.classList.contains('oo-dragging')).toBe(true);
+    component.simulate('mouseUp', { currentTarget: element });
+    expect(element.classList.contains('oo-dragging')).toBe(true);
```
aeng7: Based on the unit test title, shouldn't this expectation be false?

js/components/slider.js
```diff
@@ -116,19 +123,16 @@ var Slider = createReactClass({
...
 
   onMouseDown: function() {
-    this.dragging = true;
+    this.setState({
+      isDragging: true
+    });
   },
 
   onMouseUp: function(event) {
-    this.dragging = false;
```
aeng7: Do we want to set the state 'isDragging' to false here?

tests/components/slider-test.js
jpfranco: Hmm it should, good catch. I'll take a look at this.

js/components/slider.js
jpfranco: Yes, I must've removed this by accident. Will fix.

jpfranco:
@aeng7 I updated the PR, please take a look when you get a chance.

PR #1149, opened 2018-11-09
[PLAYER-4369] Add in ui element to show that you are casting

marcosmx:
This branch contains the needed modifications to show a panel information about the device used to cast the content through Chromecast.
These are the modified screens and his behaviour:

- **PlayingScreen**: It will show a panel in the middle of the screen with the device name. Right now has a little consideration for when the `skipControls` are enabled and it will be placed below this controls.

- **PauseScreen**: As the above screen, a panel will be shown in the center of the screen and it could react to be placed below of the `skipControls`, if active.

- **EndScreen**: At the end of the video, the size of the replay button will be a little bit smaller just to make room to place the device information just below of it.

As a general rule for all the screens, whether or not the controlBar were active by skin configurations once the player enter in the cast mode the controlBar will be showed and remains in that state until the cast session ends and it will get it's normal behaviour.

Once the cast session is started, the poster image will be used as a background. This is because if the user reload the page the auto-join policy will act and the player will get in sync with the Chromecast, but since there is no video loaded a black screen it's showed, so in order handle this in a graceful way I take the decision to use the poster image always to keep a linear state.

### DIFF
--- a/js/views/endScreen.js
+++ b/js/views/endScreen.js
@@ -12,6 +12,8 @@ var React = require('react'),
     Utils = require('../components/utils');
 var createReactClass = require('create-react-class');
 
+const CastPanel = require('../components/castPanel');
+
 var EndScreen = createReactClass({
   mixins: [ResizeMixin],
 
@@ -51,10 +53,7 @@ var EndScreen = createReactClass({
     };
 
     if (this.props.controller.state.cast.connected) {
-      actionIconStyle = Utils.extend(
-        {fontSize: '125px'},
-        actionIconStyle
-      );
+      actionIconStyle['fontSize'] = '125px';
     }
 
     var titleStyle = {
@@ -94,8 +93,6 @@ var EndScreen = createReactClass({
 
     // Shows the information of the chromecast device just below the replay icon
     const castPanelClass = ClassNames({
-      'oo-info-panel-cast': true,
-      'oo-inactive': !this.props.controller.state.cast.connected,
       'oo-info-panel-cast-bottom': true
     })
 
@@ -134,9 +131,11 @@ var EndScreen = createReactClass({
           <Icon {...this.props} icon="replay" style={actionIconStyle} />
         </button>
 
-        <div className={castPanelClass}>
-          <p>Connected to <span>{this.props.controller.state.cast.device}</span></p>
-        </div>
+        <CastPanel
+          device={this.props.controller.state.cast.device}
+          connected={this.props.controller.state.cast.connected}
+          className={castPanelClass}
+        />
 
         <div className="oo-interactive-container">
           <ControlBar
--- a/js/views/endScreen.js
+++ b/js/views/endScreen.js
@@ -50,6 +50,13 @@ var EndScreen = createReactClass({
       opacity: this.props.skinConfig.endScreen.replayIconStyle.opacity
     };
 
+    if (this.props.controller.state.cast.connected) {
+      actionIconStyle = Utils.extend(
+        {fontSize: '125px'},
+        actionIconStyle
+      );
+    }
+
     var titleStyle = {
       color: this.props.skinConfig.startScreen.titleFont.color
     };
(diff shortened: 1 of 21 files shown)


js/views/higher-order/withAutoHide.js
```diff
@@ -55,6 +55,11 @@ const withAutoHide = function(ComposedComponent) {
           this.props.controller.showControlBar();
           this.startHideControlBarTimer();
         }
+
+        if (nextProps.controller.state.cast.connected) {
+          this.props.controller.showControlBar();
```
KseniiaSokolovaEpam: if conditions for prev "if" return true, do we need to showControlBar again?

marcosmx: Good catch!!! I will update the logic there to not repeat the same action if the prev and next props has the same value

Thanks,

js/views/endScreen.js
```diff
@@ -120,6 +134,10 @@ var EndScreen = createReactClass({
           <Icon {...this.props} icon="replay" style={actionIconStyle} />
         </button>
 
+        <div className={castPanelClass}>
+          <p>Connected to <span>{this.props.controller.state.cast.device}</span></p>
+        </div>
+
```
KseniiaSokolovaEpam: There are 3 identical divs
```
<div className={castPanelClass}>
        <p>Connected to <span>{this.props.controller.state.cast.device}</span></p>
</div> 
```

in 3 different files (pause, play, end screen). Maybe it would be better to put these blocks in one file and to render this block?

scss/components/_state-screen.scss
```diff
@@ -135,6 +135,36 @@
...
   }
 
+  .oo-info-panel-cast {
+    @extend .oo-center-vertical-horizontal;
+    @include vendor-prefixes (transition, $fade-up-transition);
+    text-align: center;
+    max-width: 80%;
+    border: 1px solid gray;
+    border-radius: 3px;
+    width: 100%;
```
KseniiaSokolovaEpam: Could you tell me, please, why you set width to 100%, but max-width to 80%? Is it a trick for anything?

js/views/endScreen.js
marcosmx: Yeah, I was thinking the same....I just follow the same that we had for other elements like the panel that shown the title and description on the start, pause and end screens...
Maybe I can move this to a separate component.

js/views/endScreen.js
```diff
@@ -50,6 +50,13 @@ var EndScreen = createReactClass({
       opacity: this.props.skinConfig.endScreen.replayIconStyle.opacity
     };
 
+    if (this.props.controller.state.cast.connected) {
+      actionIconStyle = Utils.extend(
+        {fontSize: '125px'},
+        actionIconStyle
+      );
```
KseniiaSokolovaEpam: Maybe in future it would be better to use "spread" operator instead of the function from Utils.js ? :-)

scss/responsive/_large.scss
```diff
@@ -24,6 +24,11 @@
         font-size: 1.8em;
       }
     }
+    .oo-info-panel-cast {
+      p {
+        font-size: 1.8em;
+      }
+    }
```
KseniiaSokolovaEpam: 1.8em is a common size for "oo-info-panel-cast" and "oo-state-screen-title" classes?
Could we move it into sass-constant? :-)

scss/components/_state-screen.scss
marcosmx: yeah, it's kindly a trick to tell the div to use all the space but not beyond the 80% of the parent width...
If you have another trick/way to get a similar result I'll happy to check it out :)

js/views/endScreen.js
marcosmx: Sure, I will update that to use the spread operator :)

scss/responsive/_large.scss
marcosmx: You're right, I'll move it to a constant

KseniiaSokolovaEpam:
By the way, do we need to have tests here? :-)

js/views/higher-order/withAutoHide.js
marcosmx: After review the logic behind the method `showControlBar` I found that it handles internally the decision to render or not when it found a difference between the states, so I decided to keep this as is.

marcosmx:
@KseniiaSokolovaEpam @zelibobla I covered all the PR comments, please take a look again and let me know if there's something else to improve.
Thanks,

PR #1150, opened 2018-11-14
[PBW-7601] Preventing unmute button click from triggering ad clickthrough

jpfranco:
This update fixes an issue in which clicking on the "unmute" button during an ad would also trigger a click on the `AdScreen`, which in turn would trigger the ad's clickthrough.

**Note:** I don't think this is an urgent issue so it's probably better not to include this on the candidate build yet.

### DIFF
--- a/js/components/unmuteIcon.js
+++ b/js/components/unmuteIcon.js
@@ -13,11 +13,14 @@ var UnmuteIcon = createReactClass({
   },
 
   unmuteClick: function(event) {
+    // Clicking on this button shouldn't trigger clicks on parent components
     event.stopPropagation();
     this.props.controller.handleMuteClick();
   },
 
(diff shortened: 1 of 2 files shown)


js/components/unmuteIcon.js
```diff
@@ -13,9 +13,14 @@ var UnmuteIcon = createReactClass({
   },
 
   unmuteClick: function(event) {
+    event.stopPropagation();
     this.props.controller.handleMuteClick();
   },
 
+  onMouseUp: function(event) {
```
jpfranco: This is needed because the ad screen uses the on `mouseup` event for the clickthrough.

zelibobla: May I ask you to put this note into comments?

PR #1156, opened 2018-11-19
bugfix - PLAYER-4675

ALexanderSPB:
Changed options of DeepMerge inside loadConfigData for mergedMetaData (need to prepend playbackSpeed button)
https://jira.corp.ooyala.com/browse/PLAYER-4675

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -1623,7 +1623,8 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       // override data in skin config with possible local storage settings, inline data input by user, and CMS settings in backlot/themebuilder
       var mergedMetaData = DeepMerge(SkinJSON, metaDataSettings, {
         arrayMerge: Utils.arrayDeepMerge.bind(Utils),
-        arrayUnionBy: 'name'
+        arrayUnionBy: 'name',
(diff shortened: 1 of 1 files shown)


js/controller.js
```diff
@@ -1623,7 +1623,8 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       // override data in skin config with possible local storage settings, inline data input by user, and CMS settings in backlot/themebuilder
       var mergedMetaData = DeepMerge(SkinJSON, metaDataSettings, {
         arrayMerge: Utils.arrayDeepMerge.bind(Utils),
-        arrayUnionBy: 'name'
+        arrayUnionBy: 'name',
+        buttonArrayFusion: 'prepend'
```
KseniiaSokolovaEpam: Do we need to save it in constants? Or it is only single using?

ALexanderSPB: It is an option for the inner arrayDeepMerge but I don't feel like we need to save it in constant. Actually I think the better way is to receive the `buttonArrayFusion` param from the metadata (looks like it was conceived initially), but it hasn't been implemented. So looks like it is a workaround.

PR #1164, opened 2018-11-26
[bugfix/PLAYER-4575] Total Duration of asset is not shown on start screen, it is shown as 0:00

KseniiaSokolovaEpam:
The problem:
Total Duration of asset is not shown on start screen.

The reason:
The value for "duration" was in State, and it was changed in "updatePlayhead", where new state value was count within function Utils.ensureNumber using itself.

The solution:
Return promise from updatePlayhead and resolve it once setState in there is done

Unit tests passed

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -1215,7 +1215,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
         OO.log('popping queued update');
         this.skin.updatePlayhead.apply(this.skin, this.state.queuedPlayheadUpdate).then(() => {
           this.state.queuedPlayheadUpdate = null;
-        }).catch(OO.log('onSeeked: Could not set new state for skin'));
+        }).catch(() => {OO.log('onSeeked: Could not set new state for skin')});
       }
       if (Utils.isIos() && this.state.screenToShow === CONSTANTS.SCREEN.END_SCREEN && this.state.fullscreen) {
         this.state.pauseAnimationDisabled = true;
--- a/js/controller.js
+++ b/js/controller.js
@@ -681,9 +681,11 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       this.state.playerState = CONSTANTS.STATE.START;
       var duration = Utils.ensureNumber(contentTree.duration, 0) / 1000;
       if (this.skin) {
-        this.skin.updatePlayhead(null, duration);
+        let promise = this.skin.updatePlayhead(null, duration);
+        promise.then(() => {
+          this.renderSkin({ contentTree: contentTree });
+        });
       }
-      this.renderSkin({ contentTree: contentTree });
     },
 
     onSkinMetaDataFetched: function(event, skinMetaData) {
@@ -728,8 +730,9 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       this.state.contentTree = contentTree;
       this.state.playerState = CONSTANTS.STATE.START;
       // Make sure playhead is reset when we switch to a new video
-      this.skin.updatePlayhead(0, contentTree.duration, 0, 0);
-      this.renderSkin({ contentTree: contentTree });
+      this.skin.updatePlayhead(0, contentTree.duration, 0, 0).then(() => {
+        this.renderSkin({ contentTree: contentTree });
+      });
     },
 
     onAssetUpdated: function(event, asset) {
@@ -1047,11 +1050,10 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
     onPlayed: function() {
       var duration = this.state.mainVideoDuration;
       this.state.duration = duration;
-      this.skin.updatePlayhead(duration, duration, duration);
       this.state.playerState = CONSTANTS.STATE.END;
 
       if (this.state.upNextInfo.delayedSetEmbedCodeEvent) {
-        var delayedContentData = this.state.upNextInfo.delayedContentData;
+        const delayedContentData = this.state.upNextInfo.delayedContentData;
         this.state.screenToShow = CONSTANTS.SCREEN.LOADING_SCREEN;
 
         if (delayedContentData.clickedVideo.embed_code) {
@@ -1090,7 +1092,9 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       }
       // In case a video plugin fires PLAYED event after stalling without firing BUFFERED or PLAYING first
       this.setBufferingState(false);
-      this.renderSkin();
+      this.skin.updatePlayhead(duration, duration, duration).then(() => {
+        this.renderSkin();
+      });
     },
 
     onVcPlayed: function(event, source) {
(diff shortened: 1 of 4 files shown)


js/skin.js
```diff
@@ -85,30 +86,36 @@ const Skin = createReactClass({
...
+        const currentAdPlayhead = Utils.ensureNumber(adPlayhead, prevState.currentAdPlayhead);
+        return {
+          currentPlayhead: currentPlayhead,
+          duration: duration,
+          buffered: buffered,
+          currentAdPlayhead: currentAdPlayhead,
+          totalTime: totalTime
+        }
+      });
+      resolve(true);
```
ALexanderSPB: Should we use `resolve` as a callback in setState function ? If the reject is possible in this `updatePlayhead` function, should we use try-catch ?

KseniiaSokolovaEpam: Oh. yes! Thank you, @ALexanderSPB ! Good catch!

js/skin.js
```diff
@@ -85,30 +86,36 @@ const Skin = createReactClass({
...
-      currentAdPlayhead: Utils.ensureNumber(adPlayhead, this.state.currentAdPlayhead)
+    let promise = new Promise((resolve, reject) => {
+      this.setState(function(prevState) {
+        const duration = Utils.ensureNumber(newDuration, prevState.duration);
+        const totalTime = this.getTotalTime(duration);
+        const currentPlayhead = Utils.ensureNumber(newPlayhead, prevState.currentPlayhead);
+        const buffered = Utils.ensureNumber(newBuffered, prevState.buffered);
+        const currentAdPlayhead = Utils.ensureNumber(adPlayhead, prevState.currentAdPlayhead);
+        return {
+          currentPlayhead: currentPlayhead,
```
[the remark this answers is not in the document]

Thank you for review, @nzaitcev ! :-)

js/controller.js
```diff
@@ -1214,9 +1217,10 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       this.state.seeking = false;
       if (this.state.queuedPlayheadUpdate) {
         OO.log('popping queued update');
-        this.skin.updatePlayhead.apply(this.skin, this.state.queuedPlayheadUpdate);
-        this.state.queuedPlayheadUpdate = null;
-        this.renderSkin();
+        this.skin.updatePlayhead.apply(this.skin, this.state.queuedPlayheadUpdate).then(() => {
```
zelibobla: The context of the this.skin.updatePlayhead could be binded shorter with arrow function or with bind in the class constructor. I vote for arrow function.

js/controller.js
```diff
@@ -1214,9 +1217,10 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
+          this.state.queuedPlayheadUpdate = null;
+          this.renderSkin();
```
zelibobla: Changing the state of the skin (updatePlayhead doing this) actually should trigger `skin.render` automatically. What would happen if we remove `this.renderSkin` following after `this.skin.updatePlayhead` call? Should continue working like a charm.

js/skin.js
```diff
@@ -85,30 +86,36 @@ const Skin = createReactClass({
   },
 
   updatePlayhead: function(newPlayhead, newDuration, newBuffered, adPlayhead) {
-    this.setState({
-      currentPlayhead: Utils.ensureNumber(newPlayhead, this.state.currentPlayhead),
-      duration: Utils.ensureNumber(newDuration, this.state.duration),
-      buffered: Utils.ensureNumber(newBuffered, this.state.buffered),
-      currentAdPlayhead: Utils.ensureNumber(adPlayhead, this.state.currentAdPlayhead)
+    let promise = new Promise((resolve) => {
```
zelibobla: It could be just `return new Promise((resolve)`

js/skin.js
```diff
@@ -85,30 +86,36 @@ const Skin = createReactClass({
...
+        const buffered = Utils.ensureNumber(newBuffered, prevState.buffered);
+        const currentAdPlayhead = Utils.ensureNumber(adPlayhead, prevState.currentAdPlayhead);
+        return {
+          currentPlayhead,
+          duration,
+          buffered,
+          currentAdPlayhead,
+          totalTime
+        }
+      }, function() {
```
zelibobla: 103 - 105 lines could be just `}, resolve});`. No anonymous function wrapper needed here

js/skin.js
```diff
@@ -85,30 +86,36 @@ const Skin = createReactClass({
   },
 
   updatePlayhead: function(newPlayhead, newDuration, newBuffered, adPlayhead) {
-    this.setState({
-      currentPlayhead: Utils.ensureNumber(newPlayhead, this.state.currentPlayhead),
-      duration: Utils.ensureNumber(newDuration, this.state.duration),
-      buffered: Utils.ensureNumber(newBuffered, this.state.buffered),
-      currentAdPlayhead: Utils.ensureNumber(adPlayhead, this.state.currentAdPlayhead)
+    let promise = new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
KseniiaSokolovaEpam: Do we continue to follow the old code style rules? 
if we do, then we need to save the variable, otherwise we may use "return"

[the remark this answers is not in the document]
zelibobla: We don't. There was a broadcasting letter regarding to this. New coding styles are: https://google.github.io/styleguide/jsguide.html 4 - 7 chapters

tests/html5skin-test.js
```diff
@@ -108,10 +108,14 @@ describe('Controller', function() {
...
-        this.state.duration = duration;
-        this.state.buffered = buffered;
-        this.state.currentAdPlayhead = currentAdPlayhead;
+        return new Promise((resolve, reject) => {
+          this.state.currentPlayhead = currentPlayhead;
+          this.state.duration = duration;
+          this.state.buffered = buffered;
+          this.state.currentAdPlayhead = currentAdPlayhead;
+          resolve(true);
+          reject(false);
```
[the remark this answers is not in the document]

js/controller.js

js/controller.js
KseniiaSokolovaEpam: Yes, it looks like you are right, so I removed all this.renderSkin from 'then' . But I did not tested all cases where we called this.renderSkin .

js/skin.js

js/controller.js
KseniiaSokolovaEpam: I did not get your point. 
As I understood method apply is used here because of next argument is array.

js/controller.js
```diff
@@ -681,9 +681,10 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       this.state.playerState = CONSTANTS.STATE.START;
       var duration = Utils.ensureNumber(contentTree.duration, 0) / 1000;
       if (this.skin) {
-        this.skin.updatePlayhead(null, duration);
+        this.skin.updatePlayhead(null, duration).then(() => {
+          this.renderSkin({ contentTree: contentTree });
```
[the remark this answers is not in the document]
KseniiaSokolovaEpam: I left only this part of this.skin.updatePlayhead (null, duration)

I added catch cases to controller.js

js/skin.js

js/controller.js
```diff
@@ -1210,13 +1209,13 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
+    onSeeked: function() {
       this.state.seeking = false;
       if (this.state.queuedPlayheadUpdate) {
         OO.log('popping queued update');
-        this.skin.updatePlayhead.apply(this.skin, this.state.queuedPlayheadUpdate);
-        this.state.queuedPlayheadUpdate = null;
-        this.renderSkin();
+        this.skin.updatePlayhead.apply(this.skin, this.state.queuedPlayheadUpdate).then(() => {
+          this.state.queuedPlayheadUpdate = null;
+        }).catch(OO.log('onSeeked: Could not set new state for skin'));
```
zelibobla: OO.log will execute immediately. You probably wanted `() => OO.log(...` here

PR #1165, opened 2018-11-27
[PLAYER-4063][ChromeCast][Receiver] ControlBar is not displayed on Pl…

nzaitcev:
https://jira.corp.ooyala.com/browse/PLAYER-4063

Clicking on the Pause/Play button and seeking forward/backward on the sender device does not show the current play head time and video duration on the receiver side.Also the control/Progress bar itself is not displayed at the receiver end.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -205,7 +205,8 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
       cast: {
         showButton: false,
         connected: false,
-        device: ""
+        device: "",
+        isReceiver: false
       },
 
       audioOnly: false
@@ -290,6 +291,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
           'customerUi',
           _.bind(this.onPlayheadTimeChanged, this)
         );
+        this.mb.subscribe(OO.EVENTS.SEEK, 'customerUi', this.onSeek.bind(this));
         this.mb.subscribe(OO.EVENTS.SEEKED, 'customerUi', _.bind(this.onSeeked, this));
         this.mb.subscribe(OO.EVENTS.BUFFERING, 'customerUi', _.bind(this.onBuffering, this));
         this.mb.subscribe(OO.EVENTS.BUFFERED, 'customerUi', _.bind(this.onBuffered, this));
@@ -928,6 +930,11 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
     },
 
     onVcPlay: function(event, source) {
+
+      if (this.state.cast.isReceiver) {
+        this.hideControlBar();
+      }
+
       this.state.currentVideoId = source;
       if (this.state.adWasPaused &&
         this.state.currentAdsInfo &&
@@ -971,6 +978,10 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
     onPause: function(event, source, pauseReason) {
       if (this.state.failoverInProgress) {
         return;
+      }
+
+      if (this.state.cast.isReceiver) {
+        this.showControlBar();
       }
 
       if (pauseReason === CONSTANTS.PAUSE_REASON.TRANSITION) {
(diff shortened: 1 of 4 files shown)


js/controller.js
```diff
@@ -953,7 +953,11 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
         this.setClosedCaptionsLanguage();
         this.removeBlur();
         this.state.isInitialPlay = false;
-        this.renderSkin();
+        this.renderSkin().then(() => {
+          if (this.state.cast.connected) {
```
marcosmx: You need to know that this `cast` property on the state is only used on the sender side (web), the receiver will never have this  enabled or used at all, it depends on an event triggered by the ChromecastSender.js module

KseniiaSokolovaEpam:
Do we need to cover the code by tests?

js/controller.js
```diff
@@ -1209,14 +1217,21 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
+    onSeeked: function(event, time) {
       this.state.seeking = false;
       if (this.state.queuedPlayheadUpdate) {
         OO.log('popping queued update');
         this.skin.updatePlayhead.apply(this.skin, this.state.queuedPlayheadUpdate).then(() => {
           this.state.queuedPlayheadUpdate = null;
         }).catch(() => {OO.log('onSeeked: Could not set new state for skin')});
       }
+      if (this.state.cast.isReceiver) {
+        this.skin.updatePlayhead(time);
```
marcosmx: @nzaitcev 
Just a heads up, this is already handled on the receiver in an non-elegant way(https://github.com/ooyala/chromecast-sample-receiver/blob/master/lib/src/app/player.js#L512)
It would be great to remove that part on the receiver and keep your change, which is the better solution.

js/controller.js
```diff
@@ -1209,14 +1217,21 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
+        if (this.state.playerState === CONSTANTS.STATE.PLAYING) {
```
marcosmx: How sure are we that the player state will be always equals to "playing" after a seek event to restrict the visibility of the control bar based on that state?

js/controller.js
```diff
@@ -1209,14 +1217,21 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
+          this.showControlBar();
+          this.startHideControlBarTimer();
```
marcosmx: You already handle this at the `onPlaying` method, I think this should be handled by the screen that is shown at the time (playingScreen, pauseScreen, endScreen)

js/controller.js
```diff
@@ -1022,7 +1026,11 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
           this.state.screenToShow = CONSTANTS.SCREEN.PAUSE_SCREEN;
         }
         this.state.playerState = CONSTANTS.STATE.PAUSE;
-        this.renderSkin();
+        this.renderSkin().then(() => {
+          if (this.state.cast.isReceiver) {
+            this.showControlBar();
```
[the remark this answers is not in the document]
nzaitcev: @marcosmx 
At first I added show/hide/scheduleHide logic in the `withAutoHide` HOC (with additional subscriptions to the mb play/pause events in the controller) but it caused unexpected behaviour since withAutoHide cancels timers in its `componentWillUnmount`. (e.g. _play_ event handler (in which I’m calling startHideControlBarTimer) would be fired before the withAutoHide’s willUnmount)
So I decided to move this logic here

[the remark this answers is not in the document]
marcosmx: I got your point, the same happens to me and I found a better way to implement this without the need to implement a Promise. 
First of all, We need to use the PlayingScreen and PauseScreen without wrapping them with the HOC (withAutoHide), this will remove that weird behaviour with the render cycle.

Why should we use this screens with no-HOC?
That HOC was created to provide or enhance the inner component with logic to handle the user interaction on actions like "mouse enter", "mouse leave", etc... being said that, on the receiver we don't have any user interaction trough the mouse, the player is handled via the API exposed by the core.

I can give you a show case on how to implement this when you get back from vacations, if you want.

@zelibobla ^^

js/controller.js
nzaitcev: I hope so) I checked it a couple of times. When video is playing and we move playhead along the scrubber bar the  _playerState_ remains the same

js/controller.js
```diff
@@ -290,6 +291,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
           'customerUi',
           _.bind(this.onPlayheadTimeChanged, this)
         );
+        this.mb.subscribe(OO.EVENTS.SEEK, 'customerUi', this.onSeek.bind(this));
```
marcosmx: You forgot to unsubscribe for this event :)

nzaitcev: yes, thank you)

PR #1171, opened 2018-12-06
Bugfix PBW-7648

ALexanderSPB:
Removed renderSkin call from onWillPlaySingleAd function

This fixes https://jira.corp.ooyala.com/browse/PBW-7648 for the case with 'vast ads' and bitmovin.


ALexanderSPB:
This `renderSkin` call is too early here it leads skin to the `adScreen` before the ad is actually loaded. That's why we can see the default image in WebView.

ALexanderSPB:
@zelibobla Next time `renderSkin` happens in the function `onMuteStateChanged` right before the start of advertisment. After this fix the start screen and a spinner will be shown longer instead of the black screen or the default image of pause in Webview.

PR #1177, opened 2018-12-14
[PLAYER-4678]- Issue with settings in backlot, none of the settings work for Audio only player

ipadilla4:
Changes to use backlot settings in audio player.

Jira ticket: https://jira.corp.ooyala.com/browse/PLAYER-4678

### DIFF
--- a/js/components/controlBar.js
+++ b/js/components/controlBar.js
@@ -317,6 +317,26 @@ var ControlBar = createReactClass({
     this.props.controller.toggleMoreOptionsScreen(this.moreOptionsItems);
   },
 
+  /**
+  * @description Retrieves configuration from server to be applied to audio skin
+  **/
+  getAudioControlsConfig: function() {
+    //We receive location param in desktopContent, instead of audioOnly.
+    //This is necessary to display or not the button, depending on Backlot settings.
+    var defaultConfig = JSON.parse(JSON.stringify(this.props.skinConfig.buttons.audioOnly));
+    this.props.skinConfig.buttons.desktopContent.forEach(item => {
+      if (item.location !== 'none') {
+        return;
+      }
+      defaultConfig.filter(
+        field => field.name !== item.name
+      ).forEach(
+        field => field.location = 'none'
+      );
+      });
+    return defaultConfig;
+  },
+
   populateControlBar: function() {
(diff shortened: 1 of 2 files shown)


js/components/controlBar.js
```diff
@@ -317,6 +317,26 @@ var ControlBar = createReactClass({
     this.props.controller.toggleMoreOptionsScreen(this.moreOptionsItems);
   },
 
+  filterAudioControls: function() {
+    //We receive location param in desktopContent, instead of audioOnly.
+    //This is necessary to display or not the button, depending on Backlot settings.
+    var defaultConfig = this.props.skinConfig.buttons.audioOnly;
```
zelibobla: To avoid unintentional change of the original config use `JSON.parse(JSON.stringify())` object clone please.

js/components/controlBar.js
```diff
@@ -317,6 +317,26 @@ var ControlBar = createReactClass({
     this.props.controller.toggleMoreOptionsScreen(this.moreOptionsItems);
   },
 
+  filterAudioControls: function() {
```
zelibobla: As far as I see this method doesn't filter. It produces new config block based on the retrieved from server. I'd call it `getAudioControlsConfig`. Since it's not a simple getter would be great to have a comment describing the transformations doing.

js/components/controlBar.js
```diff
@@ -317,6 +317,26 @@ var ControlBar = createReactClass({
...
+      var disableButton = configItems[k].location === "none";
+      if (disableButton) {
+        for (var j = 0; j < defaultConfig.length; j++) {
+          if (defaultConfig[j].name === configName) {
+            defaultConfig[j].location = "none";
+            break;
+          }
+        }
+      }
+    }
```
zelibobla:
```suggestion
this.props.skinConfig.buttons.desktopContent.forEach(item => {
  if (item.location !== 'none') {
    return;
  }
  defaultConfig.filter(
    field => field.name !== item.name
  ).forEach(
    field => field.location = 'none'
  );
});
```
ES6 made JS much more descriptive and less talkative. That was a great job of JS community. Let's utilize it.

ipadilla4:
I made some changes, please take a look :)

PR #1179, opened 2018-12-19
remove device check on video replay

ooyala-amohnatkin:


js/controller.js
```diff
@@ -2235,15 +2233,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
...
-          if (Utils.isAndroid() || Utils.isIos()) {
-            if (this.state.isSkipAdClicked) {
-              this.state.isSkipAdClicked = false;
-            } else {
-              this.mb.publish(OO.EVENTS.REPLAY);
-            }
-          } else {
-            this.mb.publish(OO.EVENTS.REPLAY);
-          }
+          this.mb.publish(OO.EVENTS.REPLAY);
```
zelibobla: Would be great to have a brief explanation why check for Android and IOS device is not actual anymore.

ooyala-amohnatkin: 1. me it's not the best approach to keep such flag
2. this behavior was introduced here: https://github.com/ooyala/html5-skin/commit/02ccaefaefcd2b2faeca6f3ace34ff67716dd981 with a comment `Hence in case of post roll ads the click on skip ad is causing the click on end screen which in turn replays the video`, but now clicking on `#Skip Ad` doesn't call `togglePlayPause` at all, so this code is obsolete

PR #1181, opened 2018-12-20
force `renderSkin` to show end screen after the video has been played

ooyala-amohnatkin:
There's a comment below about `END_SCREEN_SHOWN`. Does anybody have an idea?
Also we need a plan how to deal with event system. Now it's a mess.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -1091,6 +1091,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
         this.state.screenToShow = CONSTANTS.SCREEN.SHARE_SCREEN;
       } else {
         this.state.screenToShow = CONSTANTS.SCREEN.END_SCREEN;
+        this.renderSkin();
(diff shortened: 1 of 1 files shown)


js/controller.js
```diff
@@ -1091,6 +1091,7 @@ OO.plugin('Html5Skin', function(OO, _, $, W) {
         this.state.screenToShow = CONSTANTS.SCREEN.SHARE_SCREEN;
       } else {
         this.state.screenToShow = CONSTANTS.SCREEN.END_SCREEN;
+        this.renderSkin();
```
zelibobla: Have you got any idea why `this.renderSkin()` shouldn't be moved into 1106 line? Seem like the screens from other conditions of this method are also would be happy to get rendered.

zelibobla:
The comment about `END_SCREEN_SHOWN` is a part of major issue with all the code base: message bus events system must be reviewed. I've added this point in our plan: https://kb.epam.com/display/OOYLOVP/%5BWEB%5D+Tech+Debt+Elimination+Roadmap

PR #1186, opened 2018-12-25
fix pause on AD on iOs.

ooyala-amohnatkin:
Somebody already implemented this for Android, but didn't do the same for iOs.
Fix led to problems with shared state in unit tests, that's why there are so many changes. The main idea is to avoid global mutable object shared between tests.


ooyala-amohnatkin:
please don't merge

ooyala-amohnatkin:
@ooyala/player-web fixed couple of tests. Now should be ready

PR #1190, opened 2019-01-07
[PLAYER-4678] - Issue with settings in backlot, none of the settings work for Audio only player

ipadilla4:
Changes to use backlot settings in audio player and some test modifications.

### DIFF
--- a/js/components/playbackSpeedPanel.js
+++ b/js/components/playbackSpeedPanel.js
@@ -35,6 +35,10 @@ class PlaybackSpeedPanel extends React.Component {
   getPlaybackSpeedOptions() {
     // Only process speed options once per component mount
     if (!this.playbackSpeedOptions) {
+      if (this.props.skinConfig.playbackSpeed.length >= 1) {
+        // We take configured values from backlot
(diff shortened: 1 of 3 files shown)


js/components/playbackSpeedPanel.js
```diff
@@ -35,6 +35,11 @@ class PlaybackSpeedPanel extends React.Component {
   getPlaybackSpeedOptions() {
     // Only process speed options once per component mount
     if (!this.playbackSpeedOptions) {
+      if (this.props.skinConfig.playbackSpeed.length >= 1) {
+        // We take configured values from backlot
+        this.props.skinConfig.playbackSpeed.options =
+        JSON.parse(JSON.stringify(this.props.skinConfig.playbackSpeed));
```
zelibobla: Honestly saying `JSON.parse(JSON.stringify(...))` is a sort of ugly construction, right? So when you need to clone the array, that has the only primitives inside, you can use more simple and beautiful spread operator: `this.props.skinConfig.playbackSpeed.options = [... this.props.skinConfig.playbackSpeed]`

PR #1193, opened 2019-01-11
[bugfix/PLAYER-4606] Use the browser settings over skin.json

KseniiaSokolovaEpam:
Now a default language is the most prioritise language.

New language selection logic for UI:
- if we can get user language, and this language is in the list of available languages, then we use this language
- if we can get user language, but this language is not in the list of available languages, then we use the default language
- if it is impossible to determine the user's language, then use the default language

Unit tests passed.
\
\
\
Extra information:
At the beginning of the player starts, the "getLanguageToUse" function is called twice: when the function "onPlayerCreated"  is called (here we get the value from skin.json) and when function "onSkinMetaDataFetched" is called (here the value from skin.json is overridden by the value from page-level params).

### DIFF
--- a/js/components/utils.js
+++ b/js/components/utils.js
@@ -714,13 +714,7 @@ var Utils = {
     if ( !(availableLanguageList && Array.isArray(availableLanguageList) && languageCode) ) {
       return false;
     }
-    for (let index = availableLanguageList.length - 1; index >= 0; index--) {
-      let availableLanguageCode = availableLanguageList[index].language;
-      if (languageCode === availableLanguageCode) {
-        return true;
-      }
-    }
-    return false;
+    return !!availableLanguageList.find(languageObj => languageObj.language === languageCode);
   },
 
   /**
--- a/js/components/utils.js
+++ b/js/components/utils.js
@@ -635,27 +635,23 @@ var Utils = {
    * @returns {String} The ISO 639-1 code of the language to use or an empty string
    */
   getLanguageToUse: function(skinConfig) {
-    let language = '';
-    if (skinConfig) {
-      const localization = skinConfig.localization;
-
-      const userBrowserLanguage = this.getUserBrowserLanguage();
-
-      if (userBrowserLanguage && localization) {
-        let isLanguageCodeInAvailablelLanguageFile = this.isLanguageCodeInAvailablelLanguageFile(
-          localization.availableLanguageFile, userBrowserLanguage
-        );
-        if (isLanguageCodeInAvailablelLanguageFile) {
-          language = userBrowserLanguage;
-        } else {
-          language = this.getDefaultLanguage(localization);
-        }
-      } else {
-        language = this.getDefaultLanguage(localization);
-      }
+    if (!skinConfig) {
+      return '';
     }
 
-    return language;
+    const localization = skinConfig.localization;
+    const userBrowserLanguage = this.getUserBrowserLanguage();
+
+    const isLanguageCodeInAvailablelLanguageFile = !!localization &&
+      this.isLanguageCodeInAvailablelLanguageFile(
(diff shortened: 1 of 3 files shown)


js/components/utils.js
```diff
@@ -632,43 +632,95 @@ var Utils = {
...
+    if ( !(availableLanguageList && Array.isArray(availableLanguageList) && languageCode) ) {
+      return false;
+    }
+    for (let index = availableLanguageList.length - 1; index >= 0; index--) {
+      let availableLanguageCode = availableLanguageList[index].language;
+      if (languageCode === availableLanguageCode) {
+        return true;
+      }
+    }
+    return false;
```
zelibobla:
```suggestion for lines 717 - 723
    return !!availableLanguageList.find(languageObj => languageObj.language === languageCode);
```
This is not just simply shorter, but eliminates also return from an iteration of loop that is a sort of antipattern.

js/components/utils.js
```diff
@@ -632,43 +632,89 @@ var Utils = {
...
-        // window.navigator.language: the preferred language of the user, usually the language of the browser UI
-        language =
-          window.navigator.browserLanguage || window.navigator.userLanguage || window.navigator.language;
+        language = this.getDefaultLanguage(localization);
       }
+    }
 
-      // remove lang sub-code
-      var primaryLanguage = language.substr(0, 2);
+    return language;
```
ooyala-amohnatkin: empty string will be returned if no `skinConfig `

js/components/utils.js
```diff
@@ -632,43 +632,89 @@ var Utils = {
...
+   * @param languageCode - two-digit value of the language
+   * @returns {Boolean} "true" if the corresponding language code "languageCode" is
+   * in the array of available languages "availableLanguageList",
+   * otherwise "false"
+   */
+  isLanguageCodeInAvailablelLanguageFile: function(availableLanguageList, languageCode) {
+    if ( !(availableLanguageList && Array.isArray(availableLanguageList) && languageCode) ) {
+      return false;
+    }
+    return !!availableLanguageList.find(languageObj => languageObj.language === languageCode);
```
ooyala-amohnatkin: use `Array.some` instead of `!!Array.find`

js/components/utils.js
```diff
@@ -632,43 +632,89 @@ var Utils = {
...
-    // set lang to default lang in skin config
-    language = localization.defaultLanguage;
-
-    // if no default lang in skin config check browser lang settings
-    if (!language) {
-      if (window.navigator.languages) {
-        // A String, representing the language version of the browser.
-        // Examples of valid language codes are: "en", "en-US", "de", "fr", etc.
-        language = window.navigator.languages[0];
+    let language = '';
```
ooyala-amohnatkin: As for me it's better to return value immediately than use `let` and long `if` clause
like this:
```
getLanguageToUse: function(skinConfig) {
  if (!skinConfig) {
    return '';
  }

  const localization = skinConfig.localization;
  const userBrowserLanguage = this.getUserBrowserLanguage();

  const isLanguageCodeInAvailablelLanguageFile = this.isLanguageCodeInAvailablelLanguageFile(
    localization.availableLanguageFile, userBrowserLanguage
  );

  if (!userBrowserLanguage || !localization || !isLanguageCodeInAvailablelLanguageFile) {
    return this.getDefaultLanguage(localization);
  }

  return userBrowserLanguage;
}
```

js/components/utils.js
KseniiaSokolovaEpam: @ooyala-amohnatkin Thank you!
\+

js/components/utils.js
KseniiaSokolovaEpam: @ooyala-amohnatkin an interesting catch! Thank you))

PR #1195, opened 2019-01-14
[PLAYER-4881] pause on getting error

ooyala-amohnatkin:
When skin receives error from mjolnir it doesn't stop the audio.
I appreciate if someone would comment this solution.


js/controller.js
```diff
@@ -2084,6 +2084,7 @@ module.exports = function(OO, _, $, W) {
       this.state.screenToShow = CONSTANTS.SCREEN.ERROR_SCREEN;
       this.state.playerState = CONSTANTS.STATE.ERROR;
       this.state.errorCode = errorCode;
+      this.mb.publish(OO.EVENTS.PAUSE);
```
zelibobla: I'd suggest moving 2080 line into callback after renderSkin. Currently, it looks quite weird: you unsubscribePlaybackEvents first, after that publishing one more event.

PR #1196, opened 2019-01-15
[PLAYER-5008] render skin on AD start or end

ooyala-amohnatkin:


js/controller.js
```diff
@@ -1464,6 +1464,7 @@ module.exports = function(OO, _, $, W) {
         this.state.adEndTime = this.state.adStartTime + this.state.adVideoDuration;
         this.skin.state.currentPlayhead = 0;
         this.removeBlur();
+        this.renderSkin();
```
KseniiaSokolovaEpam: Is it not automatically rerendered when changing states?

ooyala-amohnatkin: nope
`state` here is just mutable class field

zelibobla: @KseniiaSokolovaEpam setState method doing what you expect. But it's not used here :)

KseniiaSokolovaEpam: @ooyala-amohnatkin @zelibobla ok, thank you, guys!

PR #1199, opened 2019-01-17
[PLAYER-5026][WEB][Chromecast] Reload the page when casting

nzaitcev:
[Jira Link](https://jira.corp.ooyala.com/browse/PLAYER-5026)

Corresponding PRs for [Receiver](https://github.com/ooyala/chromecast-sample-receiver/pull/68) and [Sender](https://git.corp.ooyala.com/projects/PBW/repos/mjolnir/pull-requests/618/overview)

This issue manifests itself when playback is done and I want to reload the page (when casting)

Steps:
* Open the debug link: https://tinyurl.com/yc2my2c8
* Connect to chromecast
* Video plays on receiver
* Refresh the page when playback's ended
* Player shows the start screen with play button but on clicking the play button it buffers indefinitely

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -1059,7 +1059,7 @@ module.exports = function(OO, _, $, W) {
     },
 
     onPlayed: function() {
-      var duration = this.state.mainVideoDuration;
+      var duration = this.state.mainVideoDuration || this.state.contentTree.duration / 1000;
       this.state.duration = duration;
       this.state.playerState = CONSTANTS.STATE.END;
 


js/controller.js
```diff
@@ -1059,7 +1059,7 @@ module.exports = function(OO, _, $, W) {
     },
 
     onPlayed: function() {
-      var duration = this.state.mainVideoDuration;
+      var duration = this.state.mainVideoDuration || this.state.contentTree.duration / 1000;
```
KseniiaSokolovaEpam: Is contentTree always known?
Could be this.state.contentTree.duration === undefined?

nzaitcev: Also I didn't come up with this line, I borrowed it from onAdsPlayed in which 
`this.state.duration = this.state.contentTree.duration / 1000;` duration gets updated the same way

KseniiaSokolovaEpam: @nzaitcev 
If the response from the backend does not come the value of the duration;
But I thought that in this case most of the logic of the player would be broken.

marcosmx: Your second approach also works, but I think this change it's just for cosmetic purposes, I mean, because of the lack of playhead activity after the page reload then the skin will show the zero value on duration field.

Good catch!!

PR #1201, opened 2019-01-21
Feature/player 4114 Add AirPlay button to Web Player UI

ALexanderSPB:
Add AirPlay button to Web Player UI
https://jira.corp.ooyala.com/browse/PLAYER-4114


js/controller.js
```diff
@@ -491,6 +491,20 @@ module.exports = function(OO, _, $, W) {
...
+      switch (event.availability) {
+        case 'available':
+          this.state.isAirPlayAvailable = true;
+          break;
+        case 'not-available':
+          this.state.isAirPlayAvailable = false;
+          break;
+        default:
+          break;
+      }
```
zelibobla:
```suggestion
      this.state.isAirPlayAvailable = event.availability === 'available';
```
Untill third option of `event.availability` flag appears, let's avoid undetermined state of 502-503 lines.

tests/html5skin-test.js
```diff
@@ -1721,6 +1721,23 @@ describe('Controller', function() {
 
   });
 
+  describe('AirPlay button', () => {
+    window.WebKitPlaybackTargetAvailabilityEvent = true;
```
ooyala-amohnatkin: will it be set to true in all other tests?

js/controller.js
ooyala-amohnatkin: as I remember we'll have to describe the default behavior after new eslint rules are applied

js/components/controlBar.js
```diff
@@ -216,6 +216,11 @@ var ControlBar = createReactClass({
     }
   },
 
+  handleAirPlayClick: function() {
+    const video = this.props.controller.state.mainVideoElement;
+    video.webkitShowPlaybackTargetPicker();
```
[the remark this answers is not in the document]
ALexanderSPB: Only in Safari 9+

tests/html5skin-test.js
ALexanderSPB: I think yes

ooyala-amohnatkin: so you should set back to `false`, to avoid unexpected tests failure

ALexanderSPB: ok

ooyala-amohnatkin: please do it in `beforeAll` and `afterAll` inside current `describe`

ALexanderSPB: Thanks, done.

scss/components/_control-bar.scss
```diff
@@ -160,8 +160,8 @@
       cursor: default;
     }
 
-    .oo-cast {
-        width: 45px;
+    .oo-cast, .oo-cast-air {
```
marcosmx: I think we can use the same class for Chromecast and Airplay and just make a difference on them using the Id attribute on each element.

tests/components/controlBar-test.js
```diff
@@ -1297,17 +1297,33 @@ describe('ControlBar', function() {
...
-    it('Should not show chromecast button when player param not exist', () => {
+    it('Should not show chromecast button when player param does not exist', () => {
       var wrapper = Enzyme.mount(getControlBar());
-      var chromecastBtn = wrapper.find(".oo-cast");
+      var chromecastBtn = wrapper.find('.oo-cast');
       expect(chromecastBtn.length).toBe(0);
     });
 
-    it('Should not show chromecast button when player param not exist', () => {
+    it('Should show chromecast button when player param exists', () => {
```
marcosmx: Thanks for the corrections =)

js/controller.js
```diff
@@ -567,6 +572,12 @@ module.exports = function(OO, _, $, W) {
       // add loadedmetadata event listener to main video element
       if (videoElement) {
         videoElement.addEventListener('loadedmetadata', this.metaDataLoaded.bind(this));
+
+        // add the AirPlay TargetAvailability event listener
+        if (window.WebKitPlaybackTargetAvailabilityEvent) {
+          videoElement.addEventListener('webkitplaybacktargetavailabilitychanged',
```
marcosmx: Maybe you can also add another listener for when the media it's already on the appleTV or being casted.
I found this article related with that:
https://developer.apple.com/documentation/webkitjs/adding_an_airplay_button_to_your_safari_media_controls/#2940027
It's not mandatory...up to you to implement this.

scss/components/_control-bar.scss
ALexanderSPB: We don't use ids for the ControlBar buttons, so it would not be consistent to add id to Airplay and Chromecast buttons. I 'd prefer not to use id also because of unit tests fixing after that =)

js/controller.js
zelibobla: Great catch @marcosmx!

PR #1204, opened 2019-01-24
[fix/PBW-7648] The default image appears on Mobile webview

ALexanderSPB:
Removed renderSkin() from onWillPlaySingleAD because of Webview default image appearing.
This was fixed previously by removing this renderSkin(), but it led to ssai issues.

https://jira.corp.ooyala.com/browse/PBW-7648

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -1463,7 +1463,7 @@ module.exports = function(OO, _, $, W) {
         this.state.adEndTime = this.state.adStartTime + this.state.adVideoDuration;
         this.skin.state.currentPlayhead = 0;
         this.removeBlur();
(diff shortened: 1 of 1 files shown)


js/controller.js
```diff
@@ -1463,7 +1463,7 @@ module.exports = function(OO, _, $, W) {
         this.state.adEndTime = this.state.adStartTime + this.state.adVideoDuration;
         this.skin.state.currentPlayhead = 0;
         this.removeBlur();
-        this.renderSkin();
+        if (adItem.ssai) this.renderSkin();
```
ooyala-amohnatkin: please use brackets

PR #1207, opened 2019-01-31
[feature/PLAYER-4606] default UI language: Use the browser settings over skin.json with page-level param

KseniiaSokolovaEpam:
New logic to choose default language

1. if "useBrowserLanguage" is set to true
try to check browser language,
if it is checkable, use it as a default language,
if it is not checkable, use defaultLanguage from inline props on page-level params,
if it is not checkable and defaultLanguage on page-level params is not set, use default language

2.  if "useBrowserLanguage" is not set to true
use defaultLanguage from inline props on page-level params,
defaultLanguage on page-level params is not set, use default language

Unit tests passed

### DIFF
--- a/js/components/utils.js
+++ b/js/components/utils.js
@@ -645,10 +645,10 @@ var Utils = {
     let userBrowserLanguage;
     let isLanguageCodeInAvailablelLanguageFile = false;
 
-    const isUseUserBrowserLanguage = playerParam && playerParam['useUserBrowserLanguage'];
+    const isUseBrowserLanguage = playerParam && playerParam['useBrowserLanguage'];
 
     // if useUserBrowserLanguage is set to true, use language from user browser settings
-    if (isUseUserBrowserLanguage) {
+    if (isUseBrowserLanguage) {
       userBrowserLanguage = this.getUserBrowserLanguage();
       isLanguageCodeInAvailablelLanguageFile = !!localization &&
         this.isLanguageCodeInAvailablelLanguageFile(
--- a/js/components/utils.js
+++ b/js/components/utils.js
@@ -632,43 +632,96 @@ var Utils = {
    *
    * @function getLanguageToUse
    * @param {Object} skinConfig - The skin configuration file to read languages from
-   * @returns {String} The ISO code of the language to use
+   * @param {Object} playerParam - Page-level params
+   * @returns {String} The ISO 639-1 code of the language to use or an empty string
    */
-  getLanguageToUse: function(skinConfig) {
-    var localization = skinConfig.localization;
-    var language, availableLanguages;
-
-    // set lang to default lang in skin config
-    language = localization.defaultLanguage;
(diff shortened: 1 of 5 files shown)


tests/components/utils-test.js
```diff
@@ -527,47 +527,164 @@ describe('Utils', function() {
...
+    );
+
+    it('should return "defaultLanguage" from "localization" object', function(){
+      const defaultLanguage = Utils.getDefaultLanguage({'defaultLanguage': 'en'});
+      expect(defaultLanguage).toBe('en');
+    });
+  });
+
+  describe('tests getLanguageToUse', function() {
+    let skinConfig = {
```
ooyala-amohnatkin: isn't it better to just declare `skinConfig` here and set value in `beforeEach`?

ooyala-amohnatkin: or maybe not declare it at all as it is used only in one test

js/components/utils.js
```diff
@@ -632,43 +632,96 @@ var Utils = {
...
+
+  /**
+   *
+   * @param {Object} localization - location configuration object
+   * @param {String} localization.defaultLanguage - value for default language
+   * @returns {String} defaultLanguage - language code from file "skin.json" (
+   * before executing function "onSkinMetaDataFetched") or language code that was set
+   * in the page-level parameters (after executing function "onSkinMetaDataFetched")
+   */
+  getDefaultLanguage: function(localization) {
```
ooyala-amohnatkin: not necessary, but think of writing in simpler way
```
const defaultLanguage = !!localization && localization.defaultLanguage;
return defaultLanguage || '';
```

js/components/utils.js
```diff
@@ -632,43 +632,96 @@ var Utils = {
...
+    let defaultLanguage = '';
```
ooyala-amohnatkin: not necessary, but think of writing in a simpler way
```
const defaultLanguage = !!localization && localization.defaultLanguage;
return defaultLanguage || '';
```

KseniiaSokolovaEpam: @ooyala-amohnatkin If you do not mind I would like to save the code, because I want to check if "localization" is an Object, not only that it is exists. :-)

ooyala-amohnatkin: This code also checks if it has `defaultLanguage` key and returns it. Otherwise, it will return empty string

tests/components/utils-test.js
KseniiaSokolovaEpam: \+

js/components/utils.js
KseniiaSokolovaEpam: yes, it is, it checks the key, but not type of "localization" :-)

KseniiaSokolovaEpam: @ooyala-amohnatkin oh, yes, my fault! Thank you!

I like your way!
\+

PR #1212, opened 2019-02-04
AirPlay added the icon for 'active' state

ALexanderSPB:
https://jira.corp.ooyala.com/browse/PLAYER-5094


js/controller.js
```diff
@@ -497,6 +497,17 @@ module.exports = function(OO, _, $, W) {
       this.renderSkin();
     },
 
+    toggleAirPlayIcon: function(event) {
+      if (!this.state.airPlayStatusIcon) {
+        this.state.airPlayStatusIcon = 'airPlay-disconnected';
```
zelibobla: Don't we want to render after this as well?

ALexanderSPB: Added renderSkin

PR #1217, opened 2019-02-06
Revert "Fixed iframe playerParams parsing (enhanced for inline skin)"

marcosmx:
Reverts ooyala/html5-skin#1184


marcosmx:
Hi, we need to revert this change only for the iframe.html file, there's an issue reported on 
https://jira.corp.ooyala.com/browse/PBW-7836

So for now we will revert this

ALexanderSPB:
Yes, the error occurs in JSON.parse(playerParams.inline); when playerParams.inline is undefined.

PR #1223, opened 2019-02-07
[fix/PLAYER-5165] Airplay icon is not displayed as expected on refresh/reload

ALexanderSPB:
https://jira.corp.ooyala.com/browse/PLAYER-5165

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -500,12 +500,13 @@ module.exports = function(OO, _, $, W) {
     toggleAirPlayIcon: function(event) {
       if (!this.state.airPlayStatusIcon) {
         const airPlayState = window.sessionStorage.getItem('airPlayState');
-        if (airPlayState) {
-          this.state.airPlayStatusIcon = airPlayState;
-          this.state.mainVideoElement.webkitShowPlaybackTargetPicker();
-        } else {
-          this.state.airPlayStatusIcon = 'airPlay-disconnected';
+        if (airPlayState === 'airPlay-connected') {
+          setTimeout(() => {
+            this.showControlBar();
+            this.state.mainVideoElement.webkitShowPlaybackTargetPicker();
+          }, 2000);
         }
+        this.state.airPlayStatusIcon = 'airPlay-disconnected';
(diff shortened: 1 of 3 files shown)


js/controller.js
```diff
@@ -500,12 +500,13 @@ module.exports = function(OO, _, $, W) {
     toggleAirPlayIcon: function(event) {
       if (!this.state.airPlayStatusIcon) {
         const airPlayState = window.sessionStorage.getItem('airPlayState');
-        if (airPlayState) {
-          this.state.airPlayStatusIcon = airPlayState;
-          this.state.mainVideoElement.webkitShowPlaybackTargetPicker();
-        } else {
-          this.state.airPlayStatusIcon = 'airPlay-disconnected';
+        if (airPlayState === 'airPlay-connected') {
```
ooyala-amohnatkin: I guess we should keep `airPlay-connected` as a constant

js/controller.js
```diff
@@ -500,12 +500,13 @@ module.exports = function(OO, _, $, W) {
...
+          setTimeout(() => {
```
ooyala-amohnatkin: please add a comment why do you use setTimeout and why it's so big

zelibobla:
Every PR should be followed by description what is done and why

js/controller.js
ALexanderSPB: Fixed

js/controller.js
ALexanderSPB: I thought it would be more userfriendly to show it with some delay after buffering. But it works without settimeout.

ALexanderSPB:
When airPlayState === CONSTANTS.AIRPLAY_STATE.CONNECTED, 
Safari calls toggleAirPlayIcon 2 times  but when it is DISCONNECTED toggleAirPlayIcon is called 1 time. Thats why Airplay icon was not displayed as expected on refresh/reload. So I set this.state.airPlayStatusIcon to CONSTANTS.AIRPLAY_STATE.DISCONNECTED in any case.

PR #1226, opened 2019-02-08
[PLAYER-5145] remove extra spinner on live stream without DVR

ooyala-amohnatkin:


ooyala-amohnatkin:
fix for live streams after https://github.com/ooyala/html5-skin/pull/1187 and https://github.com/ooyala/html5-skin/pull/1210

PR #1227, opened 2019-02-12
[fix/PLAYER-5095] Trigger targetPicker only after 'playing' event

ALexanderSPB:
https://jira.corp.ooyala.com/browse/PLAYER-5095

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -492,22 +492,21 @@ module.exports = function(OO, _, $, W) {
       this.renderSkin({cast: {connected:false, device: ""}});
     },
 
+    showAirplayPicker: function() {
+      this.showControlBar();
+      this.state.mainVideoElement.webkitShowPlaybackTargetPicker();
+      this.state.mainVideoElement.removeEventListener('playing', this.showAirplayPickerBound);
+    },
+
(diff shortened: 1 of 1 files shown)


js/controller.js
```diff
@@ -501,8 +501,12 @@ module.exports = function(OO, _, $, W) {
       if (!this.state.airPlayStatusIcon) {
         const airPlayState = window.sessionStorage.getItem('airPlayState');
         if (airPlayState === CONSTANTS.AIRPLAY_STATE.CONNECTED) {
-          this.showControlBar();
-          this.state.mainVideoElement.webkitShowPlaybackTargetPicker();
+          const onVideoPlaying = () => {
```
zelibobla: A function defined in another function is one more step to the code complexity. No necessity for this here. Let me suggest instead to define in on the same level as others and give a name to it like 'showAirplayPicker'

ooyala-amohnatkin: moreover, it's a bit unclear when `toggleAirPlayIcon` adds an event listener while it should just toggle icon

ALexanderSPB: Fixed

PR #1228, opened 2019-02-14
[feature/PLAYER-5204] Add page-level parameter to disable AirPlay

ALexanderSPB:
Add page-level parameter to disable AirPlay option

https://jira.corp.ooyala.com/browse/PLAYER-5204

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -601,7 +601,7 @@ module.exports = function(OO, _, $, W) {
         videoElement.addEventListener('loadedmetadata', this.metaDataLoaded.bind(this));
 
         // add the AirPlay TargetAvailability event listener
-        if (window.WebKitPlaybackTargetAvailabilityEvent) {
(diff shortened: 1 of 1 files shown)


js/controller.js
```diff
@@ -1745,6 +1745,8 @@ module.exports = function(OO, _, $, W) {
       this.state.cast.showButton = this.isChromecastEnabled(params);
       this.state.cast.isReceiver = params.chromecast && params.chromecast.isReceiver;
 
+      this.state.isAirplayAllowed = Utils.getPropertyValue(params, 'airplay', true);
```
KseniiaSokolovaEpam: Do we have to call the function or we have the value in this.state.playerParam.airplay?

ALexanderSPB: Yes we should have this value in the this.state.playerParam.airplay, and I used this function just for the default value.

PR #1229, opened 2019-02-15
feature/PLAYER-5058 Add Skin.JSON parameter to hide UI

KseniiaSokolovaEpam:
If a parameter "controlBar.enabled"  is false, then we do not show the default control bar.

If function "onTogglePlayPause" is defined in playerParams, then we use this function instead of our "togglePlayPause".

Unit tests passed.


js/components/controlBar.js
```diff
@@ -992,7 +992,9 @@ var ControlBar = createReactClass({
       'oo-flex-end': this.props.audioOnly && Utils.isIos()
     });
 
-    return (
+    return (this.props.skinConfig &&
```
ooyala-amohnatkin: isn't it better to return `null` at the beginning of render?

tests/html5skin-test.js
```diff
@@ -521,6 +521,69 @@ describe('Controller', function() {
       expect(controller.state.controlBarVisible).toBe(true);
     });
 
+    describe('test customPlayPause', () => {
+      let spy;
+      let testEvent;
+      let testPlayerState;
+      const onTogglePlayPause = (event, playerState) => {
```
ooyala-amohnatkin: I think it makes tests overcomplicated. You implement a method in the test and check it's implementation. Checking that it has been called using `toHasBeenCalledWith` would be enough

ooyala-amohnatkin: `const onTogglePlayPause = jest.fn();`

tests/html5skin-test.js
```diff
@@ -521,6 +521,69 @@ describe('Controller', function() {
...
+        testEvent = event;
+        testPlayerState = playerState;
+      };
+
+      beforeEach(function() {
+        spy = sinon.spy(controller.mb, 'publish');
```
ooyala-amohnatkin: as we use jest maybe replace it with `jest.fn()`?

tests/html5skin-test.js

tests/html5skin-test.js

js/components/controlBar.js
KseniiaSokolovaEpam: \+
thank you for the notice!

PR #1230, opened 2019-02-15
Chore/player 4980

zelibobla:
This PR with huge diff introduces ES6 + air-bnb linting. Starting from this the pipeline to reject any incompatibilities with linting rules will be applied.
Mixins removed, two or more classes files splitted, js => jsx renamed, a lot of minor stuff done as well.
In spite of tests work fine, some regression issues quite much possible, therefore all the commits in this branch to be squashed on merge.


js/components/accessibilityControls.jsx
```diff
@@ -166,34 +170,19 @@ AccessibilityControls.prototype = {
...
 
     if (isKeyDown === true) {
-      var newBtn = true;
-      for (var j = this.prevKeyPressedArr.length - 1; j >= 0; j--) {
-        if (this.prevKeyPressedArr[j] === charCode) {
-          newBtn = false; // there is extra pressDown for pressed btn in chrome (windows) and safari if to change to fullscreen mode
-          break; // we do not need to add charCode to prevKeyPressedArr in this case
-        }
-      }
+      const newBtn = this.prevKeyPressedArr.some(address => address === charCode);
```
ooyala-amohnatkin: no ugly `for`!🙌🙌🙌

js/components/accessibilityControls.jsx
```diff
@@ -166,34 +170,19 @@ AccessibilityControls.prototype = {
...
-      }
+      const inPrevKeyPressedArrIndex = this.prevKeyPressedArr.findIndex(address => address === charCode);
       if (inPrevKeyPressedArrIndex > -1) {
         this.prevKeyPressedArr.splice(inPrevKeyPressedArrIndex, 1);
       }
     }
     if (this.prevKeyPressedArr.length) {
-      isKeyDown = true;
-      charCode = this.prevKeyPressedArr[this.prevKeyPressedArr.length - 1];
+      isKeyDown = true; // eslint-disable-line
```
ooyala-amohnatkin: I think it would be better to add what rule exactly do you ignore here and below

js/components/adPanelTopBarItem.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * One item in ad panel component
+ */
+class AdPanelTopBarItem extends React.Component { // eslint-disable-line
```
ooyala-amohnatkin: why do you ignore line instead of rewriting to functional component?

zelibobla: I discovered that the HOC doesn't work as expected for pure components. Didn't spent enough of time to get a clear answer why.

js/components/accessibilityControls.jsx
zelibobla: Moreover almost any `eslint-disable...` should get removed from repo, but for speed purposes I had to use this short way as a temporary solution. The task for the rest is here:
https://jira.corp.ooyala.com/projects/PLAYER/issues/PLAYER-5215

PR #1232, opened 2019-02-15
fix/BL-15587 live stream can be defined not only in auth data but in content tree

zelibobla:
Backend team won't change the output for auth request for live assets located not remotely. So we need to patch frontend part to make it work the same way as remote assets do.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -745,7 +745,8 @@ function controller(OO, _, $) {
     },
 
     onAuthorizationFetched(event, authorization) {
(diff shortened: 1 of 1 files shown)


ooyala-amohnatkin:
`authorization.streams[0].is_live_stream`
It seems that this check is useless now. Is it possible to remove it?

zelibobla:
> `authorization.streams[0].is_live_stream`
> It seems that this check is useless now. Is it possible to remove it?

@alexeyMohnatkin nope, it still used for remote assets :) I've explained in the Jira ticket this is not the way how this job must be done by backend team

PR #1238, opened 2019-02-19
[fix/PLAYER-5209] Only enable Discovery button if Discovery is enabled

ALexanderSPB:
https://jira.corp.ooyala.com/browse/PLAYER-5209

### DIFF
--- a/js/components/controlBar.jsx
+++ b/js/components/controlBar.jsx
@@ -927,7 +927,8 @@ class ControlBar extends React.Component {
         // do not show discovery button if no related videos available
         && (
           item.name !== 'discovery'
-          || (!controller.state.isOoyalaAds && controller.state.discoveryData)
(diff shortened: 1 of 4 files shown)


js/components/controlBar.jsx
```diff
@@ -927,7 +927,11 @@ class ControlBar extends React.Component {
         // do not show discovery button if no related videos available
         && (
           item.name !== 'discovery'
-          || (!controller.state.isOoyalaAds && controller.state.discoveryData)
+          || (!controller.state.isOoyalaAds
+            && controller.state.metadata
```
ooyala-amohnatkin: don't you need to add the same check in other places? For example, here: https://github.com/ooyala/html5-skin/blob/13f86ee34408e69684e69067170c13bf48a9fede/js/controller.js#L1110

ALexanderSPB: Thank you, I think you are right. I have added other checks.

PR #1239, opened 2019-02-20
[fix/PLAYER-5204] Add page-level parameter to disable AirPlay option

ALexanderSPB:
https://jira.corp.ooyala.com/browse/PLAYER-5204

case 1: 
load an asset with airplay enabled
try loading another asset in the same window with airplay set as false
airplay continues with the existing session
but in this case, the user would not be able to opt out of airplay (since the icon would no longer be visible).

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -630,17 +630,28 @@ function controller(OO, _, $) {
       if (videoElement) {
         videoElement.addEventListener('loadedmetadata', this.metaDataLoaded.bind(this));
 
-        // add the AirPlay TargetAvailability event listener
-        if (window.WebKitPlaybackTargetAvailabilityEvent && this.state.isAirplayAllowed) {
-          videoElement.addEventListener('webkitplaybacktargetavailabilitychanged',
-            _.bind(this.airPlayListener, this));
+        // add the AirPlay event listeners
+        if (window.WebKitPlaybackTargetAvailabilityEvent) {
+          const airPlayState = window.sessionStorage.getItem('airPlayState');
+          if (airPlayState === CONSTANTS.AIRPLAY_STATE.CONNECTED || this.state.isAirplayAllowed) {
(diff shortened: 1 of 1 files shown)


js/controller.js
```diff
@@ -630,17 +630,28 @@ function controller(OO, _, $) {
       if (videoElement) {
         videoElement.addEventListener('loadedmetadata', this.metaDataLoaded.bind(this));
 
-        // add the AirPlay TargetAvailability event listener
-        if (window.WebKitPlaybackTargetAvailabilityEvent && this.state.isAirplayAllowed) {
-          videoElement.addEventListener('webkitplaybacktargetavailabilitychanged',
-            _.bind(this.airPlayListener, this));
+        // add the AirPlay event listeners
+        if (window.WebKitPlaybackTargetAvailabilityEvent) {
```
ooyala-amohnatkin: handling airplay functionality grows on so I suggest move it from `onVcVideoElementCreated` to a separate method

js/controller.js
```diff
@@ -631,28 +642,7 @@ function controller(OO, _, $) {
...
-
-            // This event fires when a media element starts or stops AirPlay playback
-            videoElement.addEventListener(
-              'webkitcurrentplaybacktargetiswirelesschanged',
-              _.bind(this.toggleAirPlayIcon, this)
-            );
-
-            // Showing the AirPlay target picker if it was connected before reloading
-            this.showAirplayPickerBound = this.showAirplayPicker.bind(this);
-            videoElement.addEventListener(
```
ooyala-amohnatkin: are you sure to remove this listener?

ALexanderSPB: I have moved this functionality to the `toggleAirPlayIcon`, because we should call the target picker after the second event of the `webkitcurrentplaybacktargetiswirelesschanged`

PR #1247, opened 2019-03-01
[fix/PLAYER-5241] Issue with buffered part and played part in full sc…

ALexanderSPB:
Issue with buffered part and played part in full screen

This issue has appeared after mixins/ResizeMixin removal, so I added its functionality to the scrubberBar component.

https://jira.corp.ooyala.com/browse/PLAYER-5241

### DIFF
--- a/js/components/scrubberBar.jsx
+++ b/js/components/scrubberBar.jsx
@@ -54,7 +54,7 @@ class ScrubberBar extends React.Component {
       this.setState({ transitionedDuringSeek: false });
     }
     if (nextProps.componentWidth !== componentWidth) {
-      this.handleResize(nextProps);
+      this.handleResize();
     }
   }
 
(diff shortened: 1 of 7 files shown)


js/components/scrubberBar.jsx
```diff
@@ -49,9 +49,13 @@ class ScrubberBar extends React.Component {
   componentWillReceiveProps(nextProps) {
     const { transitionedDuringSeek } = this.state;
     const { seeking } = nextProps;
+    const { componentWidth } = this.props;
     if (transitionedDuringSeek && !seeking) {
       this.setState({ transitionedDuringSeek: false });
     }
+    if (nextProps.componentWidth !== componentWidth) {
+      this.handleResize(nextProps);
```
KseniiaSokolovaEpam: Do we need to add a test?

ALexanderSPB: Added

ooyala-amohnatkin: `handleResize` doesn't take arguments

tests/components/scrubberBar-test.js
```diff
@@ -557,4 +557,23 @@ describe('ScrubberBar', function() {
...
+    const wrapper = Enzyme.mount(
+      <ScrubberBar
+        controller={mockController}
+        componentWidth={600}
+      />
+    );
+    const handleResize = jest.spyOn(wrapper.instance(), 'handleResize');
+    wrapper.setProps({ componentWidth: 1900 });
+
+    expect(handleResize).toHaveBeenCalled();
```
ooyala-amohnatkin: please use blackbox testing and check actual result, not that some inner method has been called

ooyala-amohnatkin: I suppose you should check `playheadPaddingStyle`

PR #1248, opened 2019-03-01
bugfix/PLAYER-5242 Player controls are not displayed on fullscreen

KseniiaSokolovaEpam:
There was an checking if(this.isMounted ()) {...}, but function "isMounted" never existed, so when the window was resized (switching to fullscreen mode), the function was called and there was an error.

Unit tests passed.

### DIFF
--- a/js/views/adScreen.jsx
+++ b/js/views/adScreen.jsx
@@ -15,8 +15,6 @@ import Utils from '../components/utils';
  * The screen to be displayed on advertising
  */
 class AdScreen extends React.Component {
-  isMounted = false;
-
   constructor(props) {
     super(props);
     this.isMobile = this.props.controller.state.isMobile;
@@ -26,15 +24,14 @@ class AdScreen extends React.Component {
   }
 
   componentDidMount() {
-    this.isMounted = true;
     // for mobile or desktop fullscreen, hide control bar after 3 seconds
     if (this.isMobile || this.props.fullscreen) {
       this.props.controller.startHideControlBarTimer();
     }
   }
 
-  componentWillReceiveProps(nextProps) {
-    if (nextProps.componentWidth !== this.props.componentWidth) {
+  componentDidUpdate(prevProps) {
+    if (prevProps.componentWidth !== this.props.componentWidth) {
       this.handleResize();
     }
   }
@@ -68,7 +65,6 @@ class AdScreen extends React.Component {
   }
 
   componentWillUnmount() {
-    this.isMounted = false;
     this.props.controller.cancelTimer();
   }
 
@@ -194,9 +190,7 @@ class AdScreen extends React.Component {
    * @param {Object} event object
    */
   handleResize = () => {
-    if (this.isMounted) {
-      this.props.controller.startHideControlBarTimer();
-    }
+    this.props.controller.startHideControlBarTimer();
   };
 
   /**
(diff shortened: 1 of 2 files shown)


js/views/adScreen.jsx
```diff
@@ -15,6 +15,8 @@ import Utils from '../components/utils';
  * The screen to be displayed on advertising
  */
 class AdScreen extends React.Component {
+  isMounted = false;
```
ooyala-amohnatkin: This is obsolete. Component will not receive new prop if it's unmounted

ooyala-amohnatkin: also replace `componentWillReceiveProps` with `componentDidUpdate`

KseniiaSokolovaEpam: Thank you Alexey! @ooyala-amohnatkin

tests/views/adScreen-test.js
```diff
@@ -270,13 +271,44 @@ describe('AdScreen', function() {
...
+    expect(controlBarVisible).toBe(true);
+
+    const node = document.createElement('div');
+    Enzyme.mount(
+      <AdScreen
+        controller={mockController}
+        skinConfig={mockSkinConfig}
+        componentWidth={400}
+        fullscreen={false}
+      />, node
```
ooyala-amohnatkin: `mount` doesn't take second argument

tests/views/adScreen-test.js
```diff
@@ -270,13 +271,44 @@ describe('AdScreen', function() {
...
+    );
+
+    Enzyme.mount(
```
ooyala-amohnatkin: As per documentation, it unmounts component and mounts it again, so you don't simulate resize here.
Please use `setProps` to update width

tests/views/adScreen-test.js
```diff
@@ -270,13 +271,44 @@ describe('AdScreen', function() {
...
       />);
-    var unmuteIcons = wrapper.find(UnmuteIcon);
+    const unmuteIcons = wrapper.find(UnmuteIcon);
     expect(unmuteIcons.length).toBe(0);
   });
 
+  it('should hide controlBar after resize', () => {
+    let controlBarVisible = true;
+
+    mockController.startHideControlBarTimer = () => {
```
ooyala-amohnatkin: would be great to use `jest.fn()` instead and change the expectation to `expect(mockController.startHideControlBarTimer).toBeCalled()`;

tests/views/adScreen-test.js
```diff
@@ -270,13 +271,44 @@ describe('AdScreen', function() {
...
+        fullscreen={false}
+      />, node
+    );
+
+    Enzyme.mount(
+      <AdScreen
+        controller={mockController}
+        skinConfig={mockSkinConfig}
+        componentWidth={800}
+        fullscreen={true}
```
ooyala-amohnatkin: `startHideControlBarTimer` is called here just because `fullscreen` is set to `true`

KseniiaSokolovaEpam: @ooyala-amohnatkin :-) yes. you are right

tests/views/adScreen-test.js
KseniiaSokolovaEpam: \+

tests/views/adScreen-test.js
KseniiaSokolovaEpam: \+

tests/views/adScreen-test.js
KseniiaSokolovaEpam: \+

PR #1250, opened 2019-03-04
[PLAYER-5311] Move airplay init logic to the core

nzaitcev:
[Core PR](https://git.corp.ooyala.com/projects/PBW/repos/mjolnir/pull-requests/648/overview)

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -220,6 +220,11 @@ function controller(OO, _, $) {
         isReceiver: false,
       },
 
+      airplay: {
+        available: false,
+        statusIcon: CONSTANTS.AIRPLAY_STATE.DISCONNECTED,
+      },
+
       audioOnly: false,
     };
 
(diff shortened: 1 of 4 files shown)


js/controller.js
```diff
@@ -529,50 +549,29 @@ function controller(OO, _, $) {
...
+      this.state.airplay.available = isAvailable;
+      this.renderSkin();
+    },
 
-        // This event fires when a media element starts or stops AirPlay playback
-        videoElement.addEventListener(
-          'webkitcurrentplaybacktargetiswirelesschanged',
-          _.bind(this.toggleAirPlayIcon, this)
-        );
+    onAirplayConnectedChanged(eventName, isConnected) {
```
ooyala-amohnatkin: maybe better onAirplayConnect~ed~ionChanged?

js/controller.js
```diff
@@ -529,50 +549,29 @@ function controller(OO, _, $) {
...
+        this.state.airplay.statusIcon = CONSTANTS.AIRPLAY_STATE.CONNECTED;
+      } else {
+        this.state.airplay.statusIcon = CONSTANTS.AIRPLAY_STATE.DISCONNECTED;
       }
+      this.renderSkin();
     },
 
-    airPlayListener(event) {
-      this.state.isAirPlayAvailable = event.availability === 'available';
+    onAirplaySessionResumedWhenPageWasReloaded() {
```
ooyala-amohnatkin: I think it's better to name handler close to the event name. Nothing points that it happens on page reload and looks confusing

nzaitcev: > I think it's better to name handler close to the event name.

okay, I thought it would be better to elaborate (for the person who would read it) since sessionResume event is fired only when page is reloaded while casting. will remove it)

> Nothing points that it happens on page reload and looks confusing

it fires only on page reload. (Named it by analogy with chromecast) Can rename it

PR #1251, opened 2019-03-05
[bugfix/PLAYER-5272] loading spinner is displayed on IMA ads

KseniiaSokolovaEpam:
We show a spinner if prop "buffering" is true or prop "buffered" is 0, but we can not check changes of the value of "buffered" for "ima" video (the value is 0 all time). So we check if the video is ima, and if it is, do not use this rule.

[Connected PR](https://git.corp.ooyala.com/projects/PBW/repos/mjolnir/pull-requests/650)

Unit tests passed.

### DIFF
--- a/js/controller.js
+++ b/js/controller.js
@@ -32,7 +32,7 @@ function controller(OO, _, $) {
   const UNSUPPORTED_PLAYER_VERSION = 3;
 
   if (!OO.playerParams.core_version || OO.playerParams.core_version <= UNSUPPORTED_PLAYER_VERSION) {
-    console.error('Html5Skin requires at least player version 4.'); // eslint-disable-line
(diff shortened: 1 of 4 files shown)


js/controller.js
```diff
@@ -32,7 +32,7 @@ function controller(OO, _, $) {
   const UNSUPPORTED_PLAYER_VERSION = 3;
 
   if (!OO.playerParams.core_version || OO.playerParams.core_version <= UNSUPPORTED_PLAYER_VERSION) {
-    console.error('Html5Skin requires at least player version 4.');
+    console.error('Html5Skin requires at least player version 4.'); // eslint-disable-line
```
ooyala-amohnatkin: disable particular rule

PR #1252, opened 2019-03-07
[PLAYER-1934] Chapter Markers 

marcosmx:
This branch contains the following new components:

- marker: This represents a region on the scrubber bar to show whenever a marker starts and ends.
- markerIcon: This will render an icon with text or an image that could be expanded in order to show more information. This component will allow the user to perform a seek action when a click event it's detected on it.


tests/components/markers/marker-test.js
```diff
@@ -0,0 +1,112 @@
...
+
+import React from 'react';
+import Enzyme from 'enzyme';
+import Marker from '../../../js/components/markers/marker';
+
+describe('Marker component', function () {
+
+  let baseProps;
+
+  function renderComponent() {
```
ooyala-amohnatkin: Please don't use a mutable object from closure, pass props directly to this function. Otherwise, it's hard to follow what's going on

marcosmx: Thanks, good catch!

marcosmx: Done. I made a few changes to pass a new set of props to that function.

tests/components/markers/markerIcon-test.js
```diff
@@ -0,0 +1,301 @@
...
+    baseProps.data.text = 'test';
+    baseProps.accentColor = '#0000';
+    baseProps.duration = 300;
+    baseProps.scrubberBarWidth = 800;
+    const markerIcon = renderComponent();
+    expect(markerIcon.exists('div.oo-marker-bubble')).toBeTruthy();
+    expect(markerIcon.find('div.oo-marker-bubble').prop('style')).toHaveProperty('backgroundColor', '#0000');
+  });
+
+  it('Should render a markerIcon wtih a global background color (skin.json)', function () {
```
ipadilla4: Little typo here, should be "with"

js/components/scrubberBar.jsx
```diff
@@ -291,9 +294,23 @@ class ScrubberBar extends React.Component {
...
+    if (
+      event.target.className.match('oo-marker-bubble')
+      || event.target.parentElement.className.match('oo-marker-bubble')
+    ) {
+      return;
+    }
 
-    this.setState({ hoveringX: event.nativeEvent.offsetX });
+    controller.setScrubberBarHoverState(true);
+    // eslint-disable-next-line prefer-destructuring
```
ooyala-amohnatkin: please don't disable this rule here, use destructuring

js/components/scrubberBar.jsx
```diff
@@ -283,6 +285,7 @@ class ScrubberBar extends React.Component {
    */
   handleScrubberBarMouseOver = (event) => {
     const { controller, skinConfig } = this.props;
+    let offsetX;
```
ooyala-amohnatkin: it seems that we don't need to declare it here

js/components/markers/helpers.js
```diff
@@ -0,0 +1,41 @@
+const MAXSIZE = 2;
+
+/**
+ * getPosition return the position in pixels for the markers
+ *
+ * @param {*} duration duration in seconds of the asset
```
ooyala-amohnatkin: should it be a `number`?

js/components/scrubberBar.jsx
```diff
@@ -507,6 +572,11 @@ class ScrubberBar extends React.Component {
         onMouseLeave={this.handleScrubberBarMouseLeave}
         onMouseMove={scrubberBarMouseMove}
       >
+        {controller.state.markers && controller.state.markers.list.length > 0 && (
```
ooyala-amohnatkin: `controller.state.markers` will be rendered instead of div if `controller.state.markers.list.length <= 0`. Convert it into boolean: `{!!controller.state.markers && controller.state.markers.list.length > 0 && (`

tests/components/markers/markerIcon-test.js
```diff
@@ -0,0 +1,421 @@
...
+      }
+    },
+    level: 0
+  };
+
+  function renderComponent(props) {
+    return Enzyme.mount(<MarkerIcon {...props}/>);
+  };
+
+  it('Shouldn\'t render a markerIcon', function () {
```
ooyala-amohnatkin: please add a more verbose description of why it shouldn't  be rendered

js/components/scrubberBar.jsx
```diff
@@ -507,6 +572,11 @@ class ScrubberBar extends React.Component {
...
+          <div id="oo-marker-container">
```
ooyala-amohnatkin: Is it safe to use id here? What if there are 2 players on the page?

ooyala-amohnatkin:
@marcosmx great job!

js/components/scrubberBar.jsx
marcosmx: nice catch!

js/components/markers/helpers.js
marcosmx: Hahaha I just forget to add the type here.

scss/responsive/_xsmall.scss
```diff
@@ -427,6 +427,9 @@
         }
       }
     }
+    #oo-marker-container {
```
ooyala-amohnatkin: this should be a class too

marcosmx: done, thanks for notice that!!!

js/components/markers/markerIcon.jsx
```diff
@@ -95,21 +95,22 @@ class MarkerIcon extends Component {
...
       case CONSTANTS.MARKERS.TYPE.TEXT: {
-        if (!data.text || data.text.length === 0) {
+        if (!marker.text || marker.text.length === 0) {
           return null;
         }
-        const truncatedText = data.text.length <= CONSTANTS.MARKERS.MAXCHAR
-          ? data.text : data.text.slice(0, CONSTANTS.MARKERS.MAXCHAR).concat(' ...');
+        const truncatedText = marker.text.length <= CONSTANTS.MARKERS.MAXCHAR
+          ? marker.text : marker.text.slice(0, CONSTANTS.MARKERS.MAXCHAR).concat(' ...');
         content = <p>{truncatedText}</p>;
```
ooyala-amohnatkin: Isn't it better to return immediately as you do on lines 105 and 114 and avoid using an additional variable?

marcosmx: I can do it, but I will need to disable the following rule on the ESLINT config:
https://eslint.org/docs/rules/no-unreachable

marcosmx: Nervermind, you are right....since I'm not doing anything else I can avoid the usage of the `break` statement and simply return the desired content...thank you for the observation!!!

PR #1256, opened 2019-03-19
[PLAYER-5374] Fix to get position of marker at seek

marcosmx:
This branch allow the seek action on the scrubber bar along with markers.

### DIFF
--- a/js/components/scrubberBar.jsx
+++ b/js/components/scrubberBar.jsx
@@ -263,6 +263,10 @@ class ScrubberBar extends React.Component {
         : event.nativeEvent.offsetX;
     }
 
+    if (event.target.className.match('oo-marker')) {
+      offsetX += event.target.offsetLeft;
+    }
+
     this.setState({
       scrubbingPlayheadX: offsetX,
     });


js/components/scrubberBar.jsx
```diff
@@ -263,6 +263,10 @@ class ScrubberBar extends React.Component {
         : event.nativeEvent.offsetX;
     }
 
+    if (event.target.className.match('oo-marker')) {
```
KseniiaSokolovaEpam: Do we need to test it?

marcosmx: I tried, but looking at the test scenarios of the scrubber bar I didn't find any elaborate test over the scrubber bar actions ( mouseDown, mouseUp)...the ones that are written only check the existence of the event and nothing more.

ooyala-amohnatkin: what if `event.target` is a child of `.oo-marker`?

marcosmx: `.oo-marker` it's not intended to have child elements, so in this case we will never get the event propagation from a different source.

PR #1257, opened 2019-03-19
[PBW-7817] Added roles for screen reader NVDA

ALexanderSPB:
I have discovered that NVDA doesn't loose focus when clicking on elements with the `role="slider"` attribute. So I've added the `role="radio"` attribute for the 'Play/Pause' and 'Mute/Unmute' buttons. I had to set `role="radio"` instead of `role="button"` because this doesn't help to handle focus on the button properly.
https://jira.corp.ooyala.com/browse/PBW-7817


js/components/controlBar.jsx
```diff
@@ -545,6 +546,7 @@ class ControlBar extends React.Component {
             icon={volumeIcon}
             tooltip={mutedInUi ? CONSTANTS.SKIN_TEXT.UNMUTE : CONSTANTS.SKIN_TEXT.MUTE}
             onClick={this.handleVolumeIconClick}
+            role="radio"
```
ooyala-amohnatkin: Does it work if you try to mute and then unmute using the keyboard only?

KseniiaSokolovaEpam: I am interested if it works with other readers, but I guess it does.

ALexanderSPB: Yes it will work. Also it was working before but the mute button was losing only visual focus overlay. The Play/Pause button was loosing visual and actual focus.

PR #1258, opened 2019-03-19
[PLAYER-5380] Issue with end screen

marcosmx:
Fix error on OSP trying to access metadata information
This also fixes PLAYER-5381 (Issue when video is paused, for both live and VOD assets)


ipadilla4:
SVBC

PR #1259, opened 2019-03-21
[PLAYER-5375] Update playhead on marker clicked at pause screen

marcosmx:
Added the needed code to update the playhead once the marker gets clicked.

### DIFF
--- a/tests/components/markers/markerIcon-test.js
+++ b/tests/components/markers/markerIcon-test.js
@@ -244,10 +244,14 @@ describe('MarkerIcon component', function () {
   it('Should render a markerIcon (text), hover it and click to seek to the marker start', function () {
     let seeked = false;
     let seekTime = 0;
+    let playheadUpdated = false;
     const controller = {
       seek : (time) => {
         seeked = true;
         seekTime = time;
+      },
+      updateSeekingPlayhead: () => {
+        playheadUpdated = true;
       },
       state: {
         isMobile: false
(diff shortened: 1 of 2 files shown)


marcosmx:
@ooyala-amohnatkin test cases updated and passing now.

tests/components/markers/markerIcon-test.js
```diff
@@ -273,7 +277,8 @@ describe('MarkerIcon component', function () {
     
     markerIcon.find('div.oo-marker-bubble.oo-marker-text').simulate('click');
     expect(seeked).toBe(true);
-    expect(seekTime).toBe(markerIcon.props().data.start);    
+    expect(seekTime).toBe(markerIcon.props().data.start);
+    expect(playheadUpdated).toBe(true);
```
KseniiaSokolovaEpam: There is a one more way to check if function was called. You could mock the function
```
updateSeekingPlayhead: jest.fn()
```
and after that use [toHaveBeenCalled](https://jestjs.io/docs/en/expect#tohavebeencalled) 
or [toHaveBeenCalledWith](https://jestjs.io/docs/en/expect#tohavebeencalledwith)
or .not.toHaveBeenCalled , etc

marcosmx: Thanks, I'll use it next time =)

ooyala-amohnatkin: @KseniiaSokolovaEpam let's leave it for now and refactor tests later

PR #1263, opened 2019-03-27
[PLAYER-5104] [v4] [airplay] video doesnt play on turning off airplay using native (IOS) airplay option

ALexanderSPB:
I have removed the airplay button for Iphones because the IOS native AirPlay control button in the full-screen mode can not fully synchronize with our AirPlay button.


js/components/controlBar.jsx
```diff
@@ -957,7 +957,7 @@ class ControlBar extends React.Component {
         )
         && (
           item.name !== 'airPlay'
-          || (controller.state.airplay && controller.state.airplay.available)
+          || (controller.state.airplay && controller.state.airplay.available && !Utils.isIPhone())
```
ooyala-amohnatkin: please cover your changes in tests

PR #1273, opened 2019-05-22
Revert "[PBW-7817] Added roles for screen reader NVDA"

marcosmx:
Reverts ooyala/html5-skin#1257

We need to revert this in order to fix PBW-8050, this is because the screen readers are identifying play and mute icons as radio button.
For now Windows7+IE11 + NVDA are working fine and the issue remains only for Windows7+Firefox + NVDA


marcosmx:
Closed because the customer don't want a revert and will wait till we can find a fix.

PR #1275, opened 2019-06-17
[PBW-8120] styles and skin adopted for RTL mode

zelibobla:
As discovered in the PBW-8120 when <HTML dir="rtl"> attribute appeared the skin shown closed captions with shift. After my investigations several issues more found:
- Icon "play" was invalid
- Scrubber bar playhead marker moved from left to right
- ClosedCaptions popup markup got broken
- ClosedCaptions tuning screen got broken as well among with `close` button
Fixed it all.

### DIFF
--- a/js/components/closeButton.jsx
+++ b/js/components/closeButton.jsx
@@ -6,22 +6,20 @@ import CONSTANTS from '../constants/constants';
 
 class CloseButton extends React.Component { // eslint-disable-line
   render() {
-    let { cssClass } = this.props;
     const {
       className,
       controller,
+      cssClass,
       role,
       closeAction,
     } = this.props;
-    if (controller.state.readDirection === 'rtl') {
-      cssClass += ' oo-close-button-left';
-    }
     return (
       <AccessibleButton
         className={cssClass}
         focusId={`${CONSTANTS.FOCUS_IDS.CLOSE}-${Date.now()}`}
         ariaLabel={CONSTANTS.ARIA_LABELS.CLOSE}
         role={role}
+        dir={controller.state.readDirection}
         onClick={closeAction}
       >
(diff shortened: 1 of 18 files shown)


js/components/closeButton.jsx
```diff
@@ -6,12 +6,16 @@ import CONSTANTS from '../constants/constants';
...
+    let { cssClass } = this.props;
     const {
       className,
-      cssClass,
+      controller,
       role,
       closeAction,
     } = this.props;
+    if (controller.state.readDirection === 'rtl') {
+      cssClass += ' oo-close-button-left';
```
ooyala-amohnatkin: I suggest using `[dir=rtl]` in CSS instead of class names. This would be more readable and declarative

js/views/popover.jsx
```diff
@@ -29,27 +29,30 @@ class Popover extends React.Component {
...
         restoreToggleButtonFocus: targetIsChildElement,
       });
     }
   }
 
   render() {
+    const { children, dir, popoverClassName } = this.props;
     return (
       <div
+        dir={dir}
```
ooyala-amohnatkin: do you still need `dir` attribute for divs?

scss/components/_content-screen.scss
```diff
@@ -35,6 +40,10 @@
   right: $panel-padding;
 }
 
+@at-root[dir=rtl] .oo-content-screen .oo-close-button{
+  left: $panel-padding;
```
ooyala-amohnatkin: should there be `right: auto;`?

js/views/popover.jsx
zelibobla: +

scss/components/_content-screen.scss
zelibobla: +

PR #1278, opened 2019-07-08
[PLAYER-5215] eslint fixes

ooyala-amohnatkin:


js/components/utils.js
```diff
@@ -461,44 +464,6 @@ Utils.truncateTextToWidth = (element, text) => {
...
 
-/**
- * Returns a shallow clone of the object given
- *
- * @function clone
- * @param {Object} object - Object to be cloned
- * @returns {Object} Clone of the given object
- */
-
-Utils.clone = (object) => {
```
marcosmx: is this removed because there are no reference of the method in the entire project or was replaced with another function?

marcosmx: never mind, I notice that you replaced it by other function/approach

